PR #25, opened 2018-06-18
Use WORKSPACE relative file paths when using direct sources and WKTs

### LINKED ISSUE #23: Imported well known types fails to compile
When importing well know types from google protobuf the compilation fails.
It looks like the well know types files are not filtered out as they should be by:

https://github.com/bazelbuild/rules_swift/blob/a02d1b58d1128d8c038cffebfe14dfbf2abd3e77/swift/internal/swift_protoc_gen_aspect.bzl#L40

because the path is prefixed by `../com_google_protobuf/` (see debug output):

```
bazel build -s --verbose_failures //:someproto_swift_proto
DEBUG: /private/var/tmp/_bazel_aamand/c0c24d6962ccc7935d8d8f222862d21c/external/build_bazel_rules_swift/swift/internal/swift_protoc_gen_aspect.bzl:51:3: WNT ["../com_google_protobuf/google/protobuf/duration.proto"]
DEBUG: /private/var/tmp/_bazel_aamand/c0c24d6962ccc7935d8d8f222862d21c/external/build_bazel_rules_swift/swift/internal/swift_protoc_gen_aspect.bzl:51:3: WNT ["some.proto"]
INFO: Analysed target //:someproto_swift_proto (2 packages loaded).
INFO: Found 1 target...
SUBCOMMAND: # @com_google_protobuf//:duration_proto [action 'Generating Swift sources for @com_google_protobuf//:duration_proto']
(cd /private/var/tmp/_bazel_aamand/c0c24d6962ccc7935d8d8f222862d21c/execroot/bazelswiftproto && \
  exec env - \
  bazel-out/host/bin/external/build_bazel_rules_swift/tools/mkdir_and_run/mkdir_and_run bazel-out/darwin-fastbuild/bin bazel-out/host/bin/external/com_google_protobuf/protoc '--plugin=protoc-gen-swift=bazel-out/host/bin/external/com_github_apple_swift_swift_protobuf/ProtoCompilerPlugin' '--swift_out=bazel-out/darwin-fastbuild/bin' '--swift_opt=FileNaming=FullPath' '--swift_opt=Visibility=Public' '--swift_opt=ProtoPathModuleMappings=bazel-out/darwin-fastbuild/bin/external/com_google_protobuf/duration_proto.protoc_gen_swift_modules.asciipb' --descriptor_set_in bazel-out/darwin-fastbuild/genfiles/external/com_google_protobuf/duration_proto-descriptor-set.proto.bin bazel-out/darwin-fastbuild/genfiles/external/com_google_protobuf/google/protobuf/duration.proto)
ERROR: /private/var/tmp/_bazel_aamand/c0c24d6962ccc7935d8d8f222862d21c/external/com_google_protobuf/BUILD:245:2: Generating Swift sources for @com_google_protobuf//:duration_proto failed (Exit 1): mkdir_and_run failed: error executing command
  (cd /private/var/tmp/_bazel_aamand/c0c24d6962ccc7935d8d8f222862d21c/execroot/bazelswiftproto && \
  exec env - \
  bazel-out/host/bin/external/build_bazel_rules_swift/tools/mkdir_and_run/mkdir_and_run bazel-out/darwin-fastbuild/bin bazel-out/host/bin/external/com_google_protobuf/protoc '--plugin=protoc-gen-swift=bazel-out/host/bin/external/com_github_apple_swift_swift_protobuf/ProtoCompilerPlugin' '--swift_out=bazel-out/darwin-fastbuild/bin' '--swift_opt=FileNaming=FullPath' '--swift_opt=Visibility=Public' '--swift_opt=ProtoPathModuleMappings=bazel-out/darwin-fastbuild/bin/external/com_google_protobuf/duration_proto.protoc_gen_swift_modules.asciipb' --descriptor_set_in bazel-out/darwin-fastbuild/genfiles/external/com_google_protobuf/duration_proto-descriptor-set.proto.bin bazel-out/darwin-fastbuild/genfiles/external/com_google_protobuf/google/protobuf/duration.proto)
```

steeve:
When using descriptor_set_in, protoc expects files as named when
creating it. While File.short_path is supposed to do the trick, it is
not working for extenal `WORKSPACE`s. In turn, Well Known Types and
direct source build fail to compile because they are not recognized
by protoc.

Fixes #23

### DIFF
--- a/swift/internal/swift_protoc_gen_aspect.bzl
+++ b/swift/internal/swift_protoc_gen_aspect.bzl
@@ -37,6 +37,18 @@ _WELL_KNOWN_TYPE_PATHS = [
     "google/protobuf/wrappers.proto",
 ]
 
+def _workspace_relative_path(f):
+  """Returns the path of a file relative to its workspace.
+
+  Args:
+    f: The File object.
+
+  Returns:
+    The path of the file relative to its workspace.
+  """
+  wpath = paths.join(f.root.path, f.owner.workspace_root)
+  return paths.relativize(f.path, wpath)
+
 def _filter_out_well_known_types(srcs):
   """Returns the given list of files, excluding any well-known type protos.
 
@@ -47,7 +59,8 @@ def _filter_out_well_known_types(srcs):
     The given list of files with any well-known type protos (those living under
     the `google.protobuf` package) removed.
   """
-  return [f for f in srcs if f.short_path not in _WELL_KNOWN_TYPE_PATHS]
+  return [f for f in srcs
+          if _workspace_relative_path(f) not in _WELL_KNOWN_TYPE_PATHS]
 
 def _pbswift_file_path(target, proto_file=None):
   """Returns the `.pb.swift` short path corresponding to a `.proto` file.
@@ -173,7 +186,7 @@ def _register_pbswift_generate_action(
                     format="--swift_opt=ProtoPathModuleMappings=%s")
   protoc_args.add("--descriptor_set_in")
   protoc_args.add(transitive_descriptor_sets, join_with=":")
-  protoc_args.add(direct_srcs)
+  protoc_args.add([_workspace_relative_path(f) for f in direct_srcs])
 
   additional_command_inputs = [
       mkdir_and_run, protoc_executable, protoc_gen_swift_executable,


swift/internal/swift_protoc_gen_aspect.bzl
```diff
@@ -24,18 +24,35 @@ load("@bazel_skylib//:lib.bzl", "dicts", "paths")
...
+def _workspace_relative_path(f):
+  """Returns the path of a file relative to its workspace.
+
+  Args:
+    f: The File object.
+
+  Returns:
+    The path of the file relative to its workspace.
+  """
+  path = f.path
```
allevato: You can take advantage of `paths.relativize` from Skylib to simplify this function. Something like this should work:

```python
workspace_root_path = paths.join(f.root.path, f.owner.workspace_root)
return paths.relativize(f.path, workspace_root_path)
```

In the case of a file in the main workspace, `workspace_root_path` will be the empty string, and you should still get the correct path.

swift/internal/swift_protoc_gen_aspect.bzl
```diff
@@ -173,7 +191,7 @@ def _register_pbswift_generate_action(
                     format="--swift_opt=ProtoPathModuleMappings=%s")
   protoc_args.add("--descriptor_set_in")
   protoc_args.add(transitive_descriptor_sets, join_with=":")
-  protoc_args.add(direct_srcs)
+  protoc_args.add([_workspace_relative_path(f) for f in direct_srcs])
```
allevato: Are you sure this is needed? Passing the sequence of `File` objects to `Args.add` should be fine, I think. The reason the workspace-relative path is an issue above is because we were using it to filter out items from the list, not to pass them to the command line.

swift/internal/swift_protoc_gen_aspect.bzl
```diff
@@ -24,18 +24,35 @@ load("@bazel_skylib//:lib.bzl", "dicts", "paths")
...
-    "google/protobuf/duration.proto",
-    "google/protobuf/empty.proto",
-    "google/protobuf/field_mask.proto",
-    "google/protobuf/source_context.proto",
-    "google/protobuf/struct.proto",
-    "google/protobuf/timestamp.proto",
-    "google/protobuf/type.proto",
-    "google/protobuf/wrappers.proto",
-]
+_WELL_KNOWN_TYPE_PATHS = {
```
allevato: I'm not too concerned about the performance of a linear search through a list of 11 items here, but if we really want to optimize it, I recommend using the `new_sets` module in Skylib, which provides a cleaner API vs. the direct dictionary. (That being said, I'd prefer if this change was moved out of this PR, since it's not relevant to solving the particular issue at hand.)

swift/internal/swift_protoc_gen_aspect.bzl
steeve: Looking at the source code for `relativize`, I have the feeling it may be faster to use my version, what do you think?

https://github.com/bazelbuild/bazel-skylib/blob/master/lib/paths.bzl#L174-L190

swift/internal/swift_protoc_gen_aspect.bzl
steeve: `protoc` excepts the filepath to the be one in the filedescriptorset, not the one on the disk, so we need to use it here.

swift/internal/swift_protoc_gen_aspect.bzl
steeve: Sure think, rebasing and removing.

swift/internal/swift_protoc_gen_aspect.bzl
steeve: This should be the case for any proto file that's stored in an external repo.

allevato: Right, ok!

swift/internal/swift_protoc_gen_aspect.bzl
allevato: That's a premature optimization, IMO. The cost of these operations isn't significant, and I'm more concerned about making sure that we use common code to solve the same problems; the reason we have Skylib is to avoid duplication. It should be clear to someone reading `paths.relativize` what the code is doing, but less so if we implement a slightly different version by hand that produces the same result.

steeve: Gotcha, will update

swift/internal/swift_protoc_gen_aspect.bzl
```diff
@@ -28,6 +28,7 @@ _WELL_KNOWN_TYPE_PATHS = [
     "google/protobuf/any.proto",
     "google/protobuf/api.proto",
     "google/protobuf/duration.proto",
+    "google/protobuf/descriptor.proto",
```
allevato: The definitions in `descriptor.proto` aren't "well-known types" as defined by the protobuf spec, so `SwiftProtobuf` doesn't bundle it as part of the runtime. If we exclude it, users wouldn't be able to refer to it directly by its label.

However, I just tried to do so with a quick example, and the build is failing saying that `descriptor.pb.swift` wasn't generated. Since this is a cross-repository reference, I'm wondering if there's a separate issue here regarding cross-repo outputs. (Was the file you tried to reference in #26 in a different repository?)

steeve: It is part of a list of protos that are "sort of" considered WKTs, since they are used in the same way.
- `compiler_plugin.proto`
- `descriptor.proto`
- `source_context.proto`

`descriptor.proto` for instance is used when writing extension protos, so it will be required when building should your import an extension (such as what we do with `gogo.proto`).

`rules_go` considers it a WKT (https://github.com/bazelbuild/rules_go/blob/master/proto/wkt/well_known_types.bzl#L6), and we do to, since like other WKTs it is shipping with with the proto libraries, since it belongs in the `google/protobuf` package.

It is already shipping with SwiftProtobuf (https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobufPluginLibrary/descriptor.pb.swift), although in the plugin library.

I am puzzled as to why it is not generating though, since it contains actual message definitions.

steeve: And the file I tried in #26 was in a different repository (https://github.com/gogo/protobuf).

allevato: I'm one of the lead maintainers of SwiftProtobuf, so I'm familiar with how descriptors are distributed there. 🙂  It's in the plug-in library because the Swift generator uses it, but we don't want users to have to depend on that whole target just if they want descriptors for some other purpose.

Whether it's "sort of" a WKT or considered to be one by other rules repositories isn't relevant here—by design, this list should _only_ contain protos that are _baked into the SwiftProtobuf runtime library._ Our advice to SwiftProtobuf users in general has been that if they need descriptors, they should just generate descriptor.proto into their project and use it that way. But for rules_swift users, that's not necessary—they should just be able to depend on `@com_google_protobuf//:descriptor_proto`. Listing it here would prevent users from doing so, because it would be filtered out.

If you revert this change, I think this is ready to merge, and I can look into the separate issue of files not being generated correctly when they're cross-repo references.

steeve: > I'm one of the lead maintainers of SwiftProtobuf
Haha awesome! We've been using it for almost two years now, so thank you for everything :D

Sure thing, I'll revert it then.

steeve:
On a side note, it's great to see this repo having so much activity

allevato:
Thanks for contributing the fix! To borrow what you said, it's nice to see this repo having activity that isn't just mine 😄  Please keep bug reports and such coming!

Issue #37, opened 2018-07-11
swift_import not linked with ios_unit_test


keith:
Currently if you define a `swift_import` rule such as:

```
swift_import(
  name = "Foo",
  archives = [
    ":libFoo.a",
  ],
  swiftmodules = [
    ":Foo.swiftmodule",
  ],
)
```

And then you include that from a `swift_library` with another `swift_library` for tests:

```
swift_library(
  name = "Bar",
  srcs = glob(["*.swift"]),
  deps = [":Foo"],
)

swift_library(
  name = "BarTestsBinary",
  srcs = glob(["Tests/*.swift"]),
  deps = [":Bar"],
)
```

And finally you try to create a `ios_unit_test` bundle for this:

```
ios_unit_test(
  name = "BarTests",
  deps = [":BarTestsBinary"],
)
```

When you run `bazel test //:BarTests` you end up getting a linker error because the symbols from `Foo` are not defined. If you re-run this with `--linkopt -v`, you can see the `ld` invocation doesn't contain `-Lpath/to/Foo` or `-lFoo`. I noticed [here](https://github.com/bazelbuild/rules_swift/blob/e7b811857c7f9cf8eb874f3eefaf04acc0558b8c/swift/internal/swift_import.bzl#L43) that it seems like we might want to be propagating those flags here, but after making that change locally, it still doesn't appear that `ios_unit_test` is reading these.

Is this the correct configuration or should I be using different rules for this?

allevato:
I think I see what's happening here.

`ios_unit_test` uses `apple_binary` underneath to link the binary. Unlike `swift_binary`, `apple_binary` is a native rule so it doesn't know anything about `SwiftInfo` provider; it can only get the libraries that need to be linked from the `Objc` provider. We make this work for `swift_library` by *also* creating and propagating an `Objc` provider with the library in it: https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/api.bzl#L626

So fixing this should just be a matter of having `swift_import` also propagate an `Objc` provider.

keith:
[...]

PR #38, opened 2018-07-12
Support swift_import in ObjC rules

keith:
Previously it was not possible to use a `swift_import` in upstream rules
such as `ios_unit_test` since it was not propagating a compatible
provider. Now `swift_import` propagates Objective-C providers so it can
correctly be linked.

This fixes https://github.com/bazelbuild/rules_swift/issues/37

This implementation doesn't feel ideal, but I wanted to use it as a starting off point for solving this problem.

### DIFF
--- a/swift/internal/swift_import.bzl
+++ b/swift/internal/swift_import.bzl
@@ -17,11 +17,11 @@
 load(":api.bzl", "swift_common")
 load(":attrs.bzl", "swift_common_rule_attrs")
 load(":compiling.bzl", "new_objc_provider")
-load(":providers.bzl", "SwiftClangModuleInfo", "merge_swift_clang_module_infos", "SwiftToolchainInfo")
+load(":providers.bzl", "SwiftClangModuleInfo", "SwiftToolchainInfo", "merge_swift_clang_module_infos")
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 
 def _link_name(library):
-  return library[3:-2]
+    return library[3:-2]
 
 def _swift_import_impl(ctx):
     archives = ctx.files.archives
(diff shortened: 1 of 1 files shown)


swift/internal/swift_import.bzl
```diff
@@ -45,6 +49,25 @@ def _swift_import_impl(ctx):
         ),
     ]
 
+    toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
+    objc_fragment = (ctx.fragments.objc if toolchain.supports_objc_interop else None)
+    if toolchain.supports_objc_interop and objc_fragment:
+      for index, archive in enumerate(archives):
```
keith: This has the issue mentioned [here](https://github.com/bazelbuild/rules_swift/issues/37#issuecomment-404353046) we're assuming 2 things with this implementation:

1. You have 1 swiftmodule per archive (IE you're not passing individual architecture's swiftmodules
2. The list of swiftmodules is in the corresponding order to the list of archives

swift/internal/swift_import.bzl
```diff
@@ -45,6 +49,25 @@ def _swift_import_impl(ctx):
...
+        library_path = "-L{}".format(archive.dirname)
+        library_link_command = "-l{}".format(_link_name(archive.basename))
+        linkopts = [library_path, library_link_command] + swift_common.swift_runtime_linkopts(False, toolchain)
+
+        providers.append(new_objc_provider(
+            deps = deps + toolchain.implicit_deps,
+            include_path = archive.dirname,
+            link_inputs = [archive],
+            linkopts = linkopts,
```
keith: I'm not sure if this is the level where we should be creating these arguments? Originally I assumed that `link_inputs` or `include_path` would translate to `-L/path/to/it`

swift/internal/swift_import.bzl
```diff
@@ -45,6 +49,25 @@ def _swift_import_impl(ctx):
         ),
     ]
 
+    toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
+    objc_fragment = (ctx.fragments.objc if toolchain.supports_objc_interop else None)
+    if toolchain.supports_objc_interop and objc_fragment:
+      for index, archive in enumerate(archives):
+        library_path = "-L{}".format(archive.dirname)
```
keith: This path isn't absolute right now, I'm not sure if that's a problem

allevato:
Thanks for the PR! I need a bit of time before I can give a proper review of this; I've been making some fairly broad changes to the rules to fix some linking issues on Apple platforms that I want to push out first, just to make sure you're working with a current version of the code base.

keith:
Sounds good. Yep this was from master. Any idea on the timeline of your updates?


--
Keith Smiley

keith:
@allevato I've rebased this on the recent source dump if you have a chance to look at this now

keith:
Rebased on the new source dump

keith:
Comments got folded because I ran buildifier, but most still apply

PR #40, opened 2018-07-16
Add support for force loading libraries

keith:
ld doesn't have an -ObjC type flag for Swift. Because of this there are
common cases you can get in to that require the library to be force
loaded by any dependents.

This patch relies on the existing force_load_library behavior from the
ObjCProvider https://docs.bazel.build/versions/master/skylark/lib/ObjcProvider.html#force_load_library

More details: https://bugs.swift.org/browse/SR-6004

Once some fix for https://github.com/bazelbuild/rules_swift/pull/38 is merged, we can also add this there.

### DIFF
--- a/swift/internal/derived_files.bzl
+++ b/swift/internal/derived_files.bzl
@@ -213,18 +213,21 @@ def _reexport_modules_src(actions, target_name):
     """
     return actions.declare_file("{}_exports.swift".format(target_name))
 
-def _static_archive(actions, link_name):
+def _static_archive(actions, alwayslink, link_name):
     """Declares a file for the static archive created by a compilation rule.
 
     Args:
       actions: The context's actions object.
+      alwayslink: Whether to always link all object files from the archive,
+          even if they aren't referenced.
       link_name: The name of the library being built, without a `lib` prefix or
           file extension.
 
     Returns:
       The declared `File`.
     """
-    return actions.declare_file("lib{}.a".format(link_name))
+    extension = "lo" if alwayslink else "a"
+    return actions.declare_file("lib{}.{}".format(link_name, extension))
 
 def _swiftc_output_file_map(actions, target_name):
     """Declares a file for the JSON-formatted output map for a compilation action.
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -610,6 +610,7 @@ def _compile_as_library(
         toolchain,
         additional_inputs = [],
         allow_testing = True,
+        alwayslink = False,
         cc_libs = [],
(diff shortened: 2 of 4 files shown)


allevato:
This is a good start, but let's align its interface and behavior with Bazel's C++ rules; they already implement this feature and one of the design principles of these new Swift rules is that they mirror the `cc_*` counterparts when possible.

`cc_library` has an [`alwayslink`](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.alwayslink) attribute that accomplishes the same thing, and I've been meaning to implement it but I haven't gotten around to it yet. So here's the direction I'd like to see this go:

1. Change the name of the attribute to `alwayslink`.
2. If that attribute is `True`, then give the static archive produced by the rule the extension `.lo` instead of `.a`.
3. When you construct the `Objc` provider, only add the archive to the `force_load_library` field if its extension is `.lo`.

This unifies things nicely, because our rules already do the right thing (on Linux) if you have a `swift_library` that depends on a `cc_library` with `alwayslink=True` (see [linking.bzl](https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/linking.bzl#L146), which is what `swift_{binary,test}` use to link).

Those steps above get us 95% of the way there, and would solve the problem you're running into. What would still be missing is supporting `alwayslink` with `swift_{binary,test}` on Darwin, because the map function linked above needs to use `-force_load` instead of `--whole-archive`. My suggestion here is to not worry about it. It can be done as a follow-up—it's a deeper change that will require passing additional information about the platform's linker through the toolchain providers.

keith:
@allevato I've rebased and switched the implementation to `alwayslink` here!

keith:
Hmm looks like the library name change causes new issues, investigating

keith:
Ok fixed that part

swift/internal/derived_files.bzl
```diff
@@ -213,18 +213,21 @@ def _reexport_modules_src(actions, target_name):
...
+      alwayslink: Whether to always link all object files from the archive,
+          even if they aren't referenced.
       link_name: The name of the library being built, without a `lib` prefix or
           file extension.
 
     Returns:
       The declared `File`.
     """
-    return actions.declare_file("lib{}.a".format(link_name))
+    extension = "lo" if alwayslink else "a"
```
keith: I can move this duplicate logic somewhere if there's a place that makes sense

keith:
Rebased on new source dump

keith:
A follow up https://github.com/bazelbuild/rules_swift/pull/100

PR #61, opened 2018-09-25
Change Args file format from multiline to shell

keith:
With the shell file format arguments on each line are quoted, which
makes these use cases compatible with files with spaces in their paths.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -471,7 +471,7 @@ def _compile_as_objects(
     compile_args = actions.args()
     if is_feature_enabled(SWIFT_FEATURE_USE_RESPONSE_FILES, feature_configuration):
         compile_args.use_param_file("@%s", use_always = True)
-        compile_args.set_param_file_format("multiline")
+        compile_args.set_param_file_format("shell")
 
     compile_args.add("-emit-object")
     compile_args.add_all(compile_reqs.args)
--- a/swift/internal/archiving.bzl
+++ b/swift/internal/archiving.bzl
@@ -155,7 +155,7 @@ def _register_libtool_action(
     args.add("-filelist")
 
     filelist = actions.args()
-    filelist.set_param_file_format("shell")
+    filelist.set_param_file_format("multiline")
     filelist.use_param_file("%s", use_always = True)
     filelist.add_all(objects)
     filelist.add_all(libraries)
--- a/swift/internal/archiving.bzl
+++ b/swift/internal/archiving.bzl
@@ -155,7 +155,7 @@ def _register_libtool_action(
     args.add("-filelist")
 
     filelist = actions.args()
-    filelist.set_param_file_format("multiline")
+    filelist.set_param_file_format("shell")
     filelist.use_param_file("%s", use_always = True)
     filelist.add_all(objects)
     filelist.add_all(libraries)
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -75,7 +75,7 @@ def register_link_action(
     ]
 
     link_input_args = actions.args()
-    link_input_args.set_param_file_format("multiline")
+    link_input_args.set_param_file_format("shell")
     link_input_args.use_param_file("@%s", use_always = True)
 
     if toolchain.root_dir:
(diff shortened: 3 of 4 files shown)


keith:
So I actually found a problem with this a little further down the chain. Setting this results in this flag being wrapped in quotes `'-DDEBUG=1'` which ends up making clang assume this is a file path, resulting in an error because a swift frontend invocation attempts to exec clang only once, but it ends up assuming 2 files need to be handled. I'm not exactly sure what the workaround is here.

keith:
The specific clang command (found using `-Xfrontend -dump-clang-diagnostics` is:

```
'clang' '-fsyntax-only' '-fblocks' '-D__swift__=40200' '-fretain-comments-from-system-headers' '-isystem' '/Applications/Xcode-10.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift' '-fPIC' '-fmodules' '-Xclang' '-fmodule-feature' '-Xclang' 'swift' '-Werror=non-modular-include-in-framework-module' '-x' 'objective-c' '-std=gnu11' '-fobjc-arc' '-DSWIFT_CLASS_EXTRA=__attribute__((annotate("__swift native")))' '-DSWIFT_PROTOCOL_EXTRA=__attribute__((annotate("__swift native")))' '-DSWIFT_EXTENSION_EXTRA=__attribute__((annotate("__swift native")))' '-DSWIFT_ENUM_EXTRA=__attribute__((annotate("__swift native")))' '-D_ISO646_H_' '-D__ISO646_H' '-DSWIFT_SDK_OVERLAY_APPKIT_EPOCH=2' '-DSWIFT_SDK_OVERLAY_FOUNDATION_EPOCH=8' '-DSWIFT_SDK_OVERLAY2_SCENEKIT_EPOCH=3' '-DSWIFT_SDK_OVERLAY_GAMEPLAYKIT_EPOCH=1' '-DSWIFT_SDK_OVERLAY_SPRITEKIT_EPOCH=1' '-DSWIFT_SDK_OVERLAY_COREIMAGE_EPOCH=2' '-DSWIFT_SDK_OVERLAY_DISPATCH_EPOCH=2' '-DSWIFT_SDK_OVERLAY_PTHREAD_EPOCH=1' '-DSWIFT_SDK_OVERLAY_COREGRAPHICS_EPOCH=0' '-DSWIFT_SDK_OVERLAY_UIKIT_EPOCH=2' '-D__SWIFT_COMPILER_VERSION=1000037001000' '-isysroot' '/Applications/Xcode-10.0.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator12.0.sdk' '-fmodules-validate-system-headers' '-Xclang' '-fmodule-format=obj' '-fapinotes-modules' '-iapinotes-modules' '/Applications/Xcode-10.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/iphonesimulator/x86_64' '-fapinotes-swift-version=4.2' '-target' 'x86_64-apple-ios10.0' '<swift-imported-modules>' '-resource-dir' '/Applications/Xcode-10.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/clang' '-working-directory' '/private/var/tmp/_bazel_ksmiley/0b4aec06d0a277d8a5aa9cd2752bc453/execroot/__main__' '-v' '-iquotebazel-out/darwin-fastbuild/genfiles' '-iquote.' '-O0' ''-DDEBUG=1''
```

Which results in this error:

```
<unknown>:0: error: unable to handle compilation, expected exactly one compiler job in
...
<unknown>:0: error: clang importer creation failed
```

Because both `''-DDEBUG=1''` and `<swift-imported-modules>` create invocations to clang from the frontend

keith:
I am able to workaround this by setting `experimental_objc_fastbuild_options` to nothing, since that's where the debug flag is coming from https://github.com/bazelbuild/bazel/blob/d65ed23beb4b97c6b33eb7d625f98da7e0893b58/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcCommandLineOptions.java#L140-L141

keith:
Thinking more about this, this might be because of how we're forwarding the arguments to python's check_output. Because this only happens when going through our WIP persistent worker. I'll investigate further

keith:
It actually looks like libtool handles spaces in file paths in these files, so quoting it is redundant and invalid. To be safe I think I'll change this to only quote the Swift place as long as this builds otherwise

Issue #46, opened 2018-08-07
Enable WMO by default


keith:
Considering rules_swift doesn't support incremental compilation (related issue: https://github.com/bazelbuild/rules_apple/issues/21) it seems like we might be best off enabling WMO by default. It looks like [this solution](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-swift.md#whole-module-optimization) still works, and it's simple to add this to your `.bazelrc` file, but as far as I know there shouldn't be any downsides if we're already not getting the benefits of compiling incrementally.

Thoughts?

keith:
> speed-ups similar to WMO

Just FYI on this, in our testing WMO is still 25% faster than batch mode for clean builds like bazel will do.

PR #62, opened 2018-09-26
Enable whole module optimization for opt builds

kastiglione:
Follow up to https://github.com/bazelbuild/rules_swift/issues/46#issuecomment-424074965. This enables `-whole-module-optimization` for `opt` builds, which matches Xcode/Apple behavior.


allevato:
This is more complicated than it might seem because there is some fairly tricky interaction between the presence of WMO flags and the declaration of output files from the compilation action due to the way the Swift driver invokes frontend subprocesses.

kastiglione:
Thanks, @Keith brought this up too.

keith:
In the meantime I've submitted https://github.com/bazelbuild/bazel/pull/6921 to support that flag

PR #65, opened 2018-09-27
Change library arguments to list

keith:
Since Args objects cannot be combined and this use case shouldn't be a
performance hotspot, we can instead pass these as a list to better
support persistent workers which require a single Args object.

Related issue upstream https://github.com/bazelbuild/bazel/issues/6230

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -392,7 +392,7 @@ def _compile_as_objects(
 
     Args:
       actions: The context's `actions` object.
-      arguments: A list of `Args` objects that provide additional arguments to the
+      arguments: A list of arguments that provide additional to pass to the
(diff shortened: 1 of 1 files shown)


keith:
Before the addition of the wrapper this was the only piece required to support workers for this. Changing that one to a list if we wanted to would be more work here because we're passing compiler args without the wrapper args in most places.

If we're ok with going down this path until this is fixed in bazel core, I can look into that one separately

keith:
It looks like my original understanding of this was incorrect. I'm waiting to hear back about that here: https://github.com/bazelbuild/bazel/issues/6230#issuecomment-426103277

But this now seems to work with persistent workers and only this change https://github.com/bazelbuild/rules_swift/pull/68

PR #67, opened 2018-09-27
Support batch compilation for Xcode 10 and higher.

allevato:
Support batch compilation for Xcode 10 and higher.


allevato:
Of course, none of the tests that are currently running will exercise the feature, because our Travis config is still using Xcode 9.

allevato:
I'll go ahead and merge this to avoid GH from diverging from our internal version. Our internal testing shows that it should be safe.

PR #68, opened 2018-10-02
Use param file for library copts

keith:
This Args object gets passed to the Swift compiler. In order to support
persistent workers, it must be a param file.


keith:
Supersedes https://github.com/bazelbuild/rules_swift/pull/65

allevato:
This is a much more preferable solution than #65, thanks! Testing it now internally, and if everything passes I'll close everything out.

allevato:
Closed by #71.

PR #70, opened 2018-10-02
Clean up unused loads

allevato:
Clean up unused loads

This contains no functional changes, only removals of unused `load`ed symbols.


allevato:
Hmm, the workflow I used should have retained @kastiglione's authorship of the commit, but it got lost. Investigating.

allevato:
There we go!

PR #73, opened 2018-10-04
Use basename for macOS test bundle binaries

keith:
When using swift_test on macOS, the resulting bundle must have this
layout: `foo.xctest/Contents/MacOS/foo`. Previously using `dirname` to
create the `.xctest` directory resulted in the parent directory of the
module, which doesn't necessarily match the binary's name.

### DIFF
--- a/tools/xctest_runner/xctest_runner.sh.template
+++ b/tools/xctest_runner/xctest_runner.sh.template
@@ -26,7 +26,7 @@ set -eu
 BINARY="$TEST_SRCDIR/$TEST_WORKSPACE/%binary%"
 BUNDLE_DIR="$(dirname "$BINARY")/$(basename "$BINARY").xctest"
 mkdir -p "$BUNDLE_DIR/Contents/MacOS"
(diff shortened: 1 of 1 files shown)


allevato:
I suppose this always worked for the `xctest` example because the package and target were both named `xctest`.

But won't your change as written also modify the location where the bundle is created? As currently written, we create it in the same directory where the test's other runfiles are, so we don't clobber other files that might be elsewhere in the runfiles view of the repository (chances are it's rare, but I imagine it could happen).

I think what we want here is `$(dirname "$BINARY")/$(basename "$BINARY").xctest`.

keith:
I wondered that as well, sounds good, updated

keith:
@allevato I found another bug while testing this, if you used `--spawn_strategy=standalone` then symlink failed because it already existed. I've added `-f` to the invocation here to fix that. If you'd like me to split these into separate PRs let me know

allevato:
Including the change in this PR is fine (and preferred); I'll try to pull it in in the next day or two to run in our internal tests.

keith:
Thanks!

PR #74, opened 2018-10-17
Use replacement git/http external repo rules

kastiglione:
This update moves away from the deprecated external repo rules to use their recommended replacements. `rules_apple` uses the new rules too. See: https://docs.bazel.build/versions/master/be/workspace.html#git_repository


kastiglione:
Looks like `BAZEL=HEAD` is broken.

allevato:
I'll have to look into the HEAD failure; but at least it's unrelated to this change.

PR #75, opened 2018-10-17
Support linking objc_framework dependencies

keith:
Previously if you depended on a objc_framework rule and attempted to
link a binary, it would fail because the paths to the binaries you
depended on where not passed to the linker.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -22,7 +22,7 @@ load(
     "SwiftInfo",
     "SwiftToolchainInfo",
 )
-load(":utils.bzl", "collect_transitive")
+load(":utils.bzl", "collect_transitive", "objc_provider_framework_name")
 load("@bazel_skylib//lib:collections.bzl", "collections")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 
@@ -347,7 +347,7 @@ def objc_compile_requirements(args, deps, objc_fragment):
             "-Xfrontend",
             [
                 "-disable-autolink-framework",
-                _objc_provider_framework_name(framework),
+                objc_provider_framework_name(framework),
             ],
         ))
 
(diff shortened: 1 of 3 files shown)


keith:
Note there are some more things on the Objc provider we might want to add here, like linkopt, I can add those if that makes sense to do.

sergiocampama:
This one's going to be tricky, as we're removing `objc_framework` in lieu of a new `apple_framework_import` rule which will live in `rules_apple`, but since `rules_apple` depends on `rules_swift`, adding that import will result in a cyclical dependency.

One idea to move forward would be for `rules_swift` to define a provider interface to accept external frameworks that then `apple_framework_import` provides for compatibility. Another idea would be to reuse the CC support, if something like this already exists. I'm not entirely sure which one's the best path forward though, since I'm not very familiar with CC support for precompiled binaries.

allevato:
Well, it's still tricky 🙂 

@sergiocampama, will `apple_framework_import` still propagate an `Objc` provider? Then this solution would still work once that rule has been introduced as well.

Using the C++ support would be tricky, because you can pass linkopts or libraries to link, but nothing like frameworks yet—we can't do the combination of "I need these _files_ to be available to the linker action but they shouldn't be _passed as_ files, but as framework flags instead". (We may need to have the C++ folks add that.)

allevato:
In an ideal world, anything about compiling/linking that needs to be propagated for Objective-C code would live in the C++ rule APIs, which rules_apple and rules_swift would both depend on. If designed correctly, there should be no reason for any other rules compiling/linking code to depend on rules_apple—for example, even if a specific _rule_ like `apple_framework_import` lives in rules_apple, the API it exposes to dependent targets would be described in terms of C++ providers and rules_swift or rules_whatever would never need to know "this thing is an Apple framework".

PR #76, opened 2018-10-17
Remove exported but unused helper functions

kastiglione:
These two functions are not used.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -1116,31 +1116,6 @@ def _merge_swift_info_providers(targets):
         transitive_swiftmodules = depset(transitive = transitive_swiftmodules),
     )
 
-def _swift_runtime_linkopts(is_static, toolchain, is_test = False):
-    """Returns the flags that should be passed to `clang` when linking a binary.
-
-    This function provides the appropriate linker arguments to callers who need to
-    link a binary using something other than `swift_binary` (for example, an
-    application bundle containing a universal `apple_binary`).
-
-    Args:
-      is_static: A `Boolean` value indicating whether the binary should be linked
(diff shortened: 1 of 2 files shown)


swift/internal/api.bzl
```diff
@@ -1116,31 +1116,6 @@ def _merge_swift_info_providers(targets):
         transitive_swiftmodules = depset(transitive = transitive_swiftmodules),
     )
 
-def _swift_runtime_linkopts(is_static, toolchain, is_test = False):
```
allevato: This one is used by `rules_apple`: https://github.com/bazelbuild/rules_apple/blob/master/apple/bundling/swift_support.bzl#L83

`swift_common` is _the_ external API for the Swift build rules; something being exported here doesn't necessarily have to be used within this project.

swift/internal/derived_files.bzl
```diff
@@ -282,18 +282,6 @@ def _whole_module_object_file(actions, target_name):
     """
     return actions.declare_file(_ar_safe_object_path("{}.o".format(target_name)))
 
-def _xctest_bundle(actions, target_name):
```
allevato: Likewise, this isn't currently used, but it will be:

https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/swift_binary_test.bzl#L246-L247

The cited bug may actually be fixed now (it was purely an internal bug), so I can look at resolving that TODO.

swift/internal/api.bzl
kastiglione: Good to know to look at `rules_apple`. I understood that exported could mean used somewhere, I just didn't know where. There were other unused exported functions (`compilation_mode_copts`, `swiftc_command_line_and_inputs`), but they are documented in docs/api.md, so I left those alone.

PR #77, opened 2018-10-17
Don't add duplicate -emit-object flag

kastiglione:
The `-emit-object` flag was being added twice to `compile_args`. Once directly, and another time via `_swiftc_command_line_and_inputs()`. This removes the one being added directly.


allevato:
I believe we want to remove the one inside `_swiftc_command_line_and_inputs` instead, because it's conceivable to run a Swift compilation that doesn't emit objects. I've already made the change in the version I pulled internally to test, so you don't need to change it here—once I verify that there are no unintended consequences, I'll merge it.

kastiglione:
that makes more sense

PR #82, opened 2018-10-19
Use no generated header feature constant

keith:


allevato:
Thanks! I've made a small tweak to this in the version that I'm merging—the feature constant should be loaded from `features.bzl`, not from `api.bzl` (where its availability there is only a side effect of the fact that it loads it, and Starlark unfortunately doesn't prevent transitive imports yet to my knowledge).

Issue #52, opened 2018-08-16
swift_c_module nested cc_library not propagated to other rules


keith:
If you try to build a repo like [this one](https://github.com/jpsim/Yams) that has this source layout:

```
Sources
├── CYaml
│   ├── include
│   │   ├── CYaml.h
│   │   └── yaml.h
│   └── src
│       ├── api.c
│       ├── emitter.c
│       ├── parser.c
│       ├── reader.c
│       ├── scanner.c
│       ├── writer.c
│       └── yaml_private.h
└── Yams
    ├── Constructor.swift
    ├── Decoder.swift
    ├── Emitter.swift
    ├── Encoder.swift
    ├── Mark.swift
    ├── Node.Mapping.swift
    ├── Node.Scalar.swift
    ├── Node.Sequence.swift
    ├── Node.swift
    ├── Parser.swift
    ├── Representer.swift
    ├── Resolver.swift
    ├── String+Yams.swift
    ├── Tag.swift
    ├── YamlError.swift
    ├── Yams.h
    └── shim.swift
```

And a config like this:

```python
load(
    "@build_bazel_rules_apple//apple:macos.bzl",
    "macos_unit_test",
)

load(
    "@build_bazel_rules_swift//swift:swift.bzl",
    "swift_c_module",
    "swift_library",
    "swift_test",
)

cc_library(
  name = "CYamlLib",
  srcs = glob([
    "Sources/CYaml/src/*.c",
    "Sources/CYaml/src/*.h",
  ]),
  hdrs = glob([
    "Sources/CYaml/include/*.h",
  ]),
  includes = ["Sources/CYaml/include"],
  linkstatic = True,
)

swift_c_module(
  name = "CYaml",
  deps = [":CYamlLib"],
  module_map = "Sources/CYaml/include/foo.modulemap",
)

swift_library(
  name = "Yams",
  deps = [":CYaml"],
  srcs = glob(["Sources/Yams/*.swift"]),
  defines = ["SWIFT_PACKAGE"],
  module_name = "Yams",
)

swift_library(
  name = "YamsTestsBinary",
  srcs = glob(["Tests/YamsTests/*.swift"]),
  deps = [":Yams", ":CYaml"],
)

macos_unit_test(
  name = "YamsTests",
  deps = [":YamsTestsBinary"],
)
```

The `foo.modulemap` with this contents:

```
module CYaml {
    umbrella header "CYaml.h"
    export *
}
```

When you try to build the tests with:

```
[...]

PR #87, opened 2018-10-23
Generate module maps for direct `cc_library` deps.

allevato:
Generate module maps for direct `cc_library` deps.

RELNOTES: The Swift build rules now automatically generate Swift-compatible
module maps for any `cc_library` targets that are direct dependencies of a
Swift target so that C definitions in those libraries can be imported.

By default, the module name will be derived from the `cc_library` target
label using the same logic used for Swift targets. This can be overridden
by specifying `tags = ["swift_module=Foo"]` on the `cc_library`,
replacing `Foo` with the desired module name.

### DIFF
--- a/examples/xplatform/c_from_swift/BUILD
+++ b/examples/xplatform/c_from_swift/BUILD
@@ -1,9 +1,4 @@
-load(
-    "@build_bazel_rules_swift//swift:swift.bzl",
-    "swift_binary",
-    "swift_c_module",
-    "swift_library",
-)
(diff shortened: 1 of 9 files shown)


keith:
Possibly useful test case, it should help with the rule in the description here https://github.com/bazelbuild/rules_swift/issues/52

allevato:
Thanks, I'll try that out. I'm surprised it didn't work before, but now that this change lets you remove the intermediate `swift_c_module` entirely, I'd be surprised if it doesn't work _now._

PR #88, opened 2018-10-24
Propagate simple defines between Swift and Objective-C targets.

allevato:
Propagate simple defines between Swift and Objective-C targets.


keith:
Nice, this should fix https://github.com/bazelbuild/rules_swift/pull/61#issuecomment-424229318

PR #100, opened 2018-11-16
Add alwayslink support for swift_test & binary

keith:
Previously I added support to alwayslink libraries when they were
eventually linked into an Objective-C rule. This doesn't apply for
swift_test and swift_binary rules. This is useful for sharing rules
between macOS and Linux where you alwayslink a library.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -98,8 +98,9 @@ def register_link_action(
         link_input_args.add(runtime_object_path)
 
     link_input_args.add_all(objects)
-    library_args = [_link_library_map_fn(toolchain, x) for x in libraries]
-    link_input_args.add_all(library_args)
+    is_darwin = toolchain.system_name == "darwin"
+    map_fn = _darwin_link_library_map_fn if is_darwin else _linux_link_library_map_fn
+    link_input_args.add_all(libraries, map_each = map_fn)
 
     # Add various requirements from propagated Objective-C frameworks:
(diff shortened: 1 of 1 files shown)


swift/internal/linking.bzl
```diff
@@ -126,19 +127,22 @@ def register_link_action(
         outputs = outputs,
     )
 
-def _link_library_map_fn(lib):
+def _link_library_map_fn(toolchain, lib):
```
keith: It doesn't seem like map_each supports any way for us to send a second argument to this function, which is why I pulled out the map above

swift/internal/linking.bzl
```diff
@@ -126,19 +127,22 @@ def register_link_action(
...
     Returns:
         A list of command-line arguments (strings) that link the library correctly.
     """
-    if lib.basename.endswith(".lo"):
-        return "-Wl,--whole-archive,{lib},--no-whole-archive".format(lib = lib.path)
-    else:
+    if not lib.basename.endswith(".lo"):
         return lib.path
+
+    if toolchain.system_name == "darwin":
```
keith: I see this check in another place, there might be a nicer way for us to extract these?

swift/internal/linking.bzl
allevato: What about having separate Darwin and Linux map functions, and then you don't have to iterate over the list manually or worry about passing the toolchain around? We want to avoid iterating over `libraries` because it's a `depset`.

PR #106, opened 2018-11-26
Minor cleanup to how static frameworks have autolink disabled.

allevato:
Minor cleanup to how static frameworks have autolink disabled.


keith:
Thanks!

PR #107, opened 2018-12-02
Link opt builds with -dead_strip on macOS

kastiglione:
This change makes opt builds on macOS link with the `-dead_strip` flag.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -1151,7 +1151,7 @@ def _merge_swift_info_providers(targets):
         transitive_swiftmodules = depset(transitive = transitive_swiftmodules),
     )
 
-def _swift_runtime_linkopts(is_static, toolchain, is_test = False):
+def _swift_runtime_linkopts(is_static, toolchain, is_test = False, is_opt = False):
     """Returns the flags that should be passed to `clang` when linking a binary.
 
     This function provides the appropriate linker arguments to callers who need to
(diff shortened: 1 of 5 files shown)


allevato:
The Swift linking rules by design ask C++ Starlark APIs for the linking command line that should be used for anything that isn't specifically needed for Swift (like the `-pie` change in #99).

If you want general flags like `-dead_strip` that aren't specific to Swift, they should be added to CROSSTOOL, since it's not strictly the Swift rules' responsibility to handle that.

kastiglione:
I did this because the flag for this, `--objc_enable_binary_stripping` does not result in `-dead_strip` being added. This does come from CROSSTOOL. Is this a bug in rules_swift that it's not observed? Is there a different more general flag for this?

allevato:
What kind of target are you linking? A `swift_test` or one of the rules from rules_apple that has Swift dependencies?

kastiglione:
I noticed on a `swift_binary`. I should check `ios_application`.

allevato:
`--objc_enable_binary_stripping` is a legacy flag from before Bazel's Objective-C rules were implemented on top of CROSSTOOL. It conflates two unrelated things (dead stripping and symbol stripping) and we should probably get rid of it.

Does it work if you use [`--feature=dead_strip`](https://source.bazel.build/bazel/+/master:tools/osx/crosstool/CROSSTOOL.tpl;l=1031?ss=bazel) on your command line? If so, then I think it's a CROSSTOOL/C++ question whether that feature should be getting enabled by default for `opt` builds or not.

kastiglione:
--features=dead_strip worked thanks. How did you look it up? Is there a way to discover the full set of applicable features available for any given build or platform?

allevato:
For certain definitions of "discover". I basically just trawl the [`CROSSTOOL.tpl`](https://source.bazel.build/bazel/+/master:tools/osx/crosstool/CROSSTOOL.tpl?ss=bazel) file for the flag I want and then find which `feature`s it gets added by 😕

PR #111, opened 2018-12-04
Add descriptive progress messages to most actions.

allevato:
Add descriptive progress messages to most actions.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -559,6 +559,7 @@ def _compile_as_objects(
         inputs = all_inputs,
         mnemonic = "SwiftCompile",
         outputs = compile_outputs,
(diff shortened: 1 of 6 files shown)


swift/internal/api.bzl
```diff
@@ -559,6 +559,7 @@ def _compile_as_objects(
         inputs = all_inputs,
         mnemonic = "SwiftCompile",
         outputs = compile_outputs,
+        progress_message = "Compiling Swift module {}".format(module_name),
```
keith: Does this entirely replace the current logging? Otherwise you could probably exclude the module name since the file path will still be shown?

allevato: It completely replaces the current logging. If a progress message isn't specified, the default output is `"{mnemonic} {some output file from the action}"`.

PR #125, opened 2019-01-08
Add "--" to the targets to fix the label parsing.

philwo:
Unfortunately the bazelci.py script doesn't insert the argument separator yet, so we have to add it manually in the test_targets for now.


allevato:
Thanks! I've pulled this (and #124) in internally and they'll be closed by my next push to GitHub.

PR #129, opened 2019-01-14
Migrate to CcInfo

keith:


keith:
Ok it sounds like those helper functions aren't actually to be used https://github.com/bazelbuild/bazel/issues/7036#issuecomment-454318679 so I'll need to figure out what specifically we want for the callsites here, and then this should be green now

keith:
I've updated it, but turns out `quote_includes` was added after 0.21, so it's still red. These build locally so I think once that's released and we update CI's version we should be good so this should be ready for review

keith:
It looks like buildkite hasn't been updated to 0.21 yet, so these won't pass until it is. Also it looks like there's an actual failure on Linux related to these changes.

keith:
Ok the Linux issue should be fixed now, it was an ordering of the libraries to link issue. We'll see if buildkite is updated too.

keith:
Awesome it looks like at least some of buildkite has been updated

keith:
@allevato any update?

allevato:
Things have been a little slower-going because of Washington Snowpocalypse the past two weeks. 🥶

This is still moving forward though—today I just submitted a change internally (I'll push it out soon) that removes the `swift_library.cc_libs` attribute, which I wanted to do before migrating to `CcInfo` because I didn't want to bother migrating those usages as well. I'm going to follow that up by breaking up your change into a couple different ones, to migrate to the new provider and then do some cleanup based on that (e.g., I can completely get rid of `SwiftClangModuleInfo`).

PR #130, opened 2019-01-18
Update travis to Xcode 10

keith:


keith:
This is now suffering from this issue https://github.com/bazelbuild/bazel/pull/6891 but I guess it's ok

PR #135, opened 2019-01-29
Make swift_module_alias expose the ObjC provider

steeve:
It allows `swift_proto_library` to be used by `ios_static_framework`, among others.


steeve:
You are absolutely right, I was tired last night :)

PR #136, opened 2019-01-29
Set ZERO_AR_DATE=1 for ar and invoke bazel's libtool wrapper on macOS

steeve:
This PR does 2 things:
- Add `ZERO_AR_DATE=1` to the `ar` command env
- Use bazel's libtool wrapper instead of directly invoking it

Continuing what was started at #120.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -818,12 +818,15 @@ def _compile_as_library(
 
     if toolchain.system_name == "darwin":
         ar_executable = None
+        libtool_executable = toolchain.libtool_executable
     else:
         ar_executable = toolchain.cc_toolchain_info.ar_executable
+        libtool_executable = None
 
     register_static_archive_action(
         actions = actions,
         ar_executable = ar_executable,
+        libtool_executable = libtool_executable,
         libraries = cc_lib_files,
         mnemonic = "SwiftArchive",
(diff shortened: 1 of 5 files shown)


allevato:
Repeating my comment from the other PR:

I've given this some thought, and here's where I'm at: the Swift rules shouldn't be involved with the nitty-gritty details of static linking _at all._ The fact that `swift_library` has to make decisions about whether it should use `ar` or `libtool` is already going in the wrong direction.

Fortunately the work that's been going on with the C++/CROSSTOOL APIs (`cc_common`) will help us do the right thing. There are a couple gaps still in those APIs, but once they're ready, the Swift rules can just pass the `.o` files to that API and it'll do the right thing based on the CROSSTOOL being used, without having to worry about which tool is invoked or what platform it's on.

For that reason I'd prefer to wait for those APIs rather than cause more churn in the current code, especially when that churn would include pulling in platform-specific external dependencies (which we would also have to handle differently internally within Google). Implementing this in terms of the new API will be a nice clean-up opportunity.

allevato:
Now that #195 is pushed, this is no longer necessary—static archiving actions use the underlying C++ CROSSTOOL. So, if there are still any hermeticity issues with archiving, they should be fixed there.

steeve:
thanks @allevato !

PR #142, opened 2019-02-15
Update Skylib, remove some attributes, migrate to CcInfo.

allevato:
Move skylib to 0.7.0.


keith:
@allevato any update on this? Just trying to stay ahead of incompatible changes from bazel

PR #147, opened 2019-02-26
Remove unnecessary apple_platform_type flags

keith:
With bazel 0.23.0 this is no longer needed because macOS is now the default


keith:
If you'd like I can leave the ability to set the platform specific bazelrc files, they just aren't needed right now

allevato:
For the sake of simplicity, I'm fine with removing the platform-specific rc files. We can always add them back if they become necessary again.

PR #151, opened 2019-03-04
Remove ar mri logic

keith:
Now that cc_libs has been removed, we don't need this script and can
instead call ar directly. This has the benefit of removing the name
translation that converted + -> _P, which also affected the index store
data produced by swiftc.

### DIFF
--- a/swift/internal/archiving.bzl
+++ b/swift/internal/archiving.bzl
@@ -22,7 +22,6 @@ def register_static_archive_action(
         ar_executable,
         output,
         toolchain,
-        libraries = [],
         mnemonic = "Archive",
         objects = [],
         progress_message = None):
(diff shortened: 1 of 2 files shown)


swift/internal/archiving.bzl
```diff
@@ -92,41 +92,19 @@ def _register_ar_action(
...
-    ])
 
     args = actions.args()
-    args.add(mri_script)
+    args.add("cr")
     args.add(output)
 
-    run_toolchain_shell_action(
+    args.add_all(objects)
+    args.add_all(libraries)
```
allevato: This line should be removed (and from the `inputs` argument of the action below, too). IIRC, passing a `.a` file to `ar` outside of an MRI script just archives the `.a` file as an entry itself (rather than the desired effect of merging its entries with the archive being created).

Technically that makes the contract of this function a lie, but nobody should be passing `libraries` to this function now anyway AFAIK, so it's safer to remove this and make sure the wrong behavior doesn't accidentally happen.

keith: I was able to remove this all together instead

PR #153, opened 2019-03-06
Add fallthrough fail if compilation_mode isn't handled

keith:


keith:
Based on what we're seeing, I currently expect this to fail CI since `fastbuild` doesn't seem to be handled

keith:
Hopefully this can be fixed by https://github.com/bazelbuild/bazel/pull/7651

keith:
@allevato can we merge this now to make sure this doesn't break in the future?

PR #154, opened 2019-03-07
Specify simulator in target triple

kastiglione:
I notice that simulator builds don't have the `-simulator` [environment](http://llvm.org/doxygen/classllvm_1_1Triple.html#a1778f5c464f88710033f7e11e84a9324) in the target triple. With this change, `-simulator` is added for platforms where `is_device` is false. This matches Xcode.


kastiglione:
I noticed that Bazel's osx crosstool also doesn't include `-simulator`. I don't know if it matters for linking, so there's probably no rush, but maybe they should be updated for consistency.

PR #159, opened 2019-03-11
Update doc references to "cc" provider

kastiglione:
Follow up to #129


allevato:
These .md files are autogenerated from the .bzl files, so I just need to run that script again instead of updating them manually.

PR #165, opened 2019-04-01
Force -fmodule-map-file= flags to be absolute paths to fix debugging in Swift 5.0.

allevato:
Force -fmodule-map-file= flags to be absolute paths to fix debugging in Swift 5.0.

Forcing the paths to be absolute seems like the opposite of what we want, and long-term it is. But lldb in Swift 5.0 does not remap relative -fmodule-map-file= paths correctly, so in order to salvage the debugging experience we have to do this temporarily.


allevato:
The buildifier issue is "harmless"; we're tracking a way to get better feedback about what actually needs to be reformatted in https://github.com/bazelbuild/continuous-integration/issues/593.

Going ahead and merging this.

PR #174, opened 2019-04-08
Add a persistent worker to support `swiftc` incremental compilation.

### LINKED ISSUE #48: Support -incremental mode
We may be able to do this with a Bazel persistent worker, which would let us non-hermetically collect the outputs of a library's previous compilations and feed them back in as inputs to future compilations, so that the driver can detect whether they have changed.

However, this is more complicated than simply flipping on a flag—we'll need to design and build a local server that is used as the worker process.

swiple-rules-gardener:
Add a persistent worker to support `swiftc` incremental compilation.

To perform incremental builds, pass the `--strategy=SwiftCompile=worker` flag to Bazel (or add it to your .bazelrc file).

Fixes #48.

### DIFF
--- a/tools/worker/main.cc
+++ b/tools/worker/main.cc
@@ -1,0 +1,115 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <unistd.h>
+
+#include <iostream>
+
+#include <google/protobuf/io/zero_copy_stream_impl.h>
+#include <google/protobuf/util/delimited_message_util.h>
+#include "third_party/bazel_protos/worker_protocol.pb.h"
+#include "tools/worker/work_processor.h"
+
+// How Swift Incremental Compilation Works
+// =======================================
+// When a Swift module is compiled, the output file map (a JSON file mapping
+// source files to outputs) tells the compiler where to write the object (.o)
+// files and partial .swiftmodule files. For incremental mode to work, the
+// output file map must also contain "swift-dependencies" entries; these files
+// contain compiler-internal data that describes how the sources in the module
+// are interrelated. Once all of these outputs exist on the file system, future
+// invocations of the compiler will use them to detect which source files
+// actually need to be recompiled if any of them change.
+//
+// This compilation model doesn't interact well with Bazel, which expects builds
+// to be hermetic (not affected by each other). In other words, outputs of build
+// N are traditionally not available as inputs to build N+1; the action
+// declaration model does not allow this.
+//
+// One could disable the sandbox to hack around this, but this should not be a
+// requirement of a well-designed build rule implementation.
+//
+// Bazel provides "persistent workers" to address this. A persistent worker is a
+// long-running "server" that waits for requests, which it can then handle
+// in-process or by spawning other commands (we do the latter). The important
+// feature here is that this worker can manage a separate file store that allows
+// state to persist across multiple builds.
+//
+// However, there are still some caveats that we have to address:
+//
+// - The "SwiftCompile" actions registered by the build rules must declare the
+//   object files and partial .swiftmodules as outputs, because later actions
+//   need those files as inputs (e.g., archiving a static library or linking a
+//   dynamic library or executable).
+//
+// - Because those files are declared action outputs, Bazel will delete them or
+//   otherwise make them unavailable before the action executes, which destroys
+//   our persistent state.
+//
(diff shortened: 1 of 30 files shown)


tools/worker/main.cc
```diff
@@ -0,0 +1,115 @@
...
+// invocations of the compiler will use them to detect which source files
+// actually need to be recompiled if any of them change.
+//
+// This compilation model doesn't interact well with Bazel, which expects builds
+// to be hermetic (not affected by each other). In other words, outputs of build
+// N are traditionally not available as inputs to build N+1; the action
+// declaration model does not allow this.
+//
+// One could disable the sandbox to hack around this, but this should not be a
+// requirement of a well-designed build rule implementation.
```
kastiglione: We've discussed this before, but it has never gone beyond labeling ("hack", "well designed", etc). I really want to understand this. Are there documents or discussions somewhere that we can link to that discuss real pros and cons? Perhaps there is info that googlers are exposed to that the community is not, because my outsider perspectives (takes) are:

1. persistent workers are themselves a hack, but since they're blessed there's an implicit appeal to authority, or something
2. adding a "no-sandbox" requirement to just those actions that need incremental appears to achieve the exact same level of escape, but is simpler, doesn't bring in all the accidental complexity of persistent workers
3. persistent workers don't model persistent files (adding persistent processes where they aren't required for anything)

I could go on, but I mainly just want to highlight that I feel a misunderstanding. I've been trying to see it from the other angle, but I can't, and when I've asked questions, I still come up empty. I feel no-sandbox is objectively better in every way, why is that you don't? 😂

tools/worker/work_processor.cc
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  std::ostringstream stderr_stream;
+  int exit_code = RunSubProcess(processed_args, &stderr_stream,
+                                /*stdout_to_stderr=*/true);
+
+  if (!is_wmo) {
+    // Copy the output files from the incremental storage area back to the
+    // locations where Bazel declared the files.
```
kastiglione: Nice!

tools/worker/work_processor.cc
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      // If WMO is forcing us out of incremental mode, just put the original
+      // output file map back so the outputs end up where they should.
+      params_file_stream << "-output-file-map\n";
+      params_file_stream << output_file_map_path << '\n';
+    }
+
+    // Pass the incremental flags regardless of whether WMO is enabled or not.
+    // The driver will prefer WMO over incremental, but by passing this flag, we
+    // don't silently swallow a warning from the compiler that incremental mode
```
keith: I think for us we would rather silence this worker. My thinking with our prototype worker was that we would always pass `--strategy=SwiftCompile=worker` and then in some cases optionally pass wmo. Reason being I'd rather always go through the persistent worker codepaths, or never go through it, since I think that would reduce the surface area of places we have to debug when something goes wrong.

tools/worker/work_processor.cc
allevato: Also planning to look at CoW (on macOS) or hard linking as possible optimizations here, as long as that doesn't interfere with Bazel in weird ways.

tools/worker/work_processor.cc
allevato: Yeah, that's a good point. In some cases (for example, `-wmo` on individual targets), I think it makes sense to keep the warning because one day I hope people won't be setting that manually, but that's overshadowed by the fact that you'd have to switch strategies just to do an optimized build or you'd get a ton of noise in your build.

keith: Yea, another case that @kastiglione brought up is that depending on the target wmo builds may be faster than incremental builds, or you may want to do wmo builds for some targets just because they user doesn't care about having incremental builds for them, in those cases I think you'd also like to not have the warning

kastiglione: I just added `-whole-module-optimization` to an Xcode invocation using `-incremental`, and there was no warning. Also just `scrum swiftc -incremental -whole-module-optimization ... *.swift` didn't produce a warning.

tools/worker/main.cc
allevato: Disabling the sandbox is using a big hammer to hit a very specific nail, and it's not something that should be done lightly because it can lead to correctness issues. In the past, we've seen incorrect Objective-C builds due to sandboxing differences in local and remote builds—a local build (with sandboxing disabled) would import headers that were in build targets that it forgot to depend on, and then the corresponding remote build (with sandboxing enabled) would break post-submit because the forgotten dependency meant the headers wouldn't be pulled into the remote sandbox.

That particular situation is less of an issue for Swift because of its dependency model, but as a general rule of thumb disabling the sandbox should be considered a "last resort" (for example, rules_apple does this for ibtool/actool actions that act up with sandboxing enabled) but not as a way to accomplish something that has other means that have been designed for that purpose. The implementation of the persistent worker provides better control over what's going on than just "let this action access anywhere it wants."

I also have plans to get rid of the bazel_xcode and wrapped_swift wrappers and merge everything into the worker, using it as a sort of "überdriver" for all Swift actions—this will greatly simplify the action registration logic in the rules. (Not that that will affect end users that much, but it's a tech debt reduction I'm looking forward to.)

tools/worker/work_processor.cc
allevato: Ah-ha, I see why. It only shows up when you have `-debug-show-incremental` enabled; I thought it was present unconditionally.

PR #185, opened 2019-04-16
Make the Swift module cache path explicitly different from Clang's.

swiple-rules-gardener:
Make the Swift module cache path explicitly different from Clang's.

The cached modules aren't actually shared anyway—the Clang invocation used to generate them from Objective-C differs from the Clang invocation used by Swift's ClangImporter, so they hash to different paths (this is not just a Bazel characteristic, it happens with Xcode as well). At best, this makes sharing the location worthless; at worst, it opens the possibility for corruption or collision.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -384,21 +384,30 @@ def _sanitizer_copts(feature_configuration):
             copts.extend(flags)
     return copts
 
-def _global_module_cache_path(genfiles_dir):
-    """Returns the path of the global Clang module cache.
+def _global_module_cache_path(bin_dir):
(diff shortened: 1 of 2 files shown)


keith:
With this change should we remove `swift.use_global_module_cache`? IIRC the only case for it was issues with Swift conflicting with the ObjC module cache?

allevato:
Eventually. I need to have some of our teams who were affected try switching to this renamed module cache and exercising it for a bit before I remove the ephemeral module cache, in order to determine if it was the *shared* module cache that was the problem or if there are other module cache corruption issues at play.

PR #192, opened 2019-04-19
Enable warnings as errors on CI

keith:


allevato:
Unfortunately, we depend on far too many third-party packages that we don't have any direct control over for this to work.

keith:
Yea I wondered how many failures there would be here. FYI I enabled -Werror on protobuf upstream so at least that one won't be an issue next time we update

keith:
@allevato I'm happy to try and figure out how to fix / suppress these warnings, if it were green would you want this?

allevato:
Even if someone goes through and cleans up all the warnings on the current versions of the dependencies, I don't want to be in a situation where one of those dependencies will upgrade and introduce a new warning, and then I need to upgrade to that new version, but before I can do so I have to chase down the warnings in those projects and fix them (and then wait for yet another release of that dependency).

People should care more about keeping their projects free of unnecessary warnings, but it's the responsibility of those project owners to do so; globally expecting all of our deps to be warning-free all the time is pretty untenable.

keith:
Sounds good, I'll try to enable more of these upstream.

FWIW we have 100% overlap of dependencies here and we make sure we have a warning free project, but I understand your concern. For future reference I validated the only changes necessary to fix all the current warnings are to upgrade dependencies, and pass `-Wno-unused-functions` for BoringSSL

PR #194, opened 2019-04-19
Use cpp command-line linkopts during final link

chronal-xnor:
Currently, if you have some `--linkopt=...` options specified on the Bazel command line, `swift_binary`/swift_test targets will NOT get these in their final link. This means that you cannot reliably use e.g. `-fsanitize=address` with your cpp transitive dependencies, because the final link might not pass that option to clang and it won't include the library support necessary to define some of the symbols that were inserted earlier in the compile by ASan. This PR attempts to fix this by adding `ctx.fragments.cpp.linkopts` to the list of linker options in these targets, so that the requested linkopts are respected in the final link. First time contributor, so let me know if this is the right way to do this, or if there's a better way.


allevato:
Looks good! Thanks for fixing this oversight.

PR #196, opened 2019-04-23
Rollback of commit 522cd9d4e8f88fb5caac2945b3f6d38bd0cddba3

swiple-rules-gardener:
Rollback of commit 522cd9d4e8f88fb5caac2945b3f6d38bd0cddba3

*** Reason for rollback ***

Broke some internal targets.

*** Original change description ***

Use the C++ toolchain to register static archive actions.

This lets us leverage any platform-specific flags and wrappers that might be present in CROSSTOOL, instead of replicating that functionality here.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -800,15 +800,12 @@ def _compile_as_library(
 
     register_static_archive_action(
         actions = actions,
-        cc_feature_configuration = _cc_feature_configuration(
-            feature_configuration = feature_configuration,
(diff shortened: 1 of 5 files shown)


keith:
:sob:

allevato:
I know ☹️   Ironically I was more worried about the open-source CROSSTOOL failing...

keith:
Is the fix do-able short term?

allevato:
God I hope so.

PR #214, opened 2019-05-02
Add custom macOS Swift toolchain support

keith:
This allows you to pass `--define=SWIFT_CUSTOM_TOOLCHAIN=toolchain.id`
to specify a custom swift toolchain like the ones downloaded from
https://swift.org/download/#snapshots to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,22 @@ environment variable `CC=clang` when invoking Bazel.
 This step is not necessary for macOS users because the Xcode toolchain always
 uses `clang`.
 
+## Building with Custom Toolchains
+
+**macOS hosts:** You can build with a custom toolchain installed in
+`/Library/Developer/Toolchains` instead of Xcode's default. To do so, pass the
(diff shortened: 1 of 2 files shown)


keith:
I'm not exactly sure where in the current docs structure to add this, thoughts?

allevato:
Maybe add something like this in README.md above the Future Work section?

````markdown
## Building with Custom Toolchains

**macOS hosts:** You can build with a custom toolchain installed in
`/Library/Developer/Toolchains` instead of Xcode's default. To do so, pass the
following flag to Bazel:

```
--define=SWIFT_CUSTOM_TOOLCHAIN=toolchain.id
```

where `toolchain.id` is the value of the `CFBundleIdentifier` key in the
toolchain's Info.plist file.

**Linux hosts:** At this time, Bazel uses whichever `swift` executable is
encountered first on your `PATH`.
````

keith:
Updated with that blurb

PR #224, opened 2019-05-09
Enable exclusivity enforcement for dbg builds

kastiglione:
Add `-enforce-exclusivity=checked` for `dbg` builds. This matches Xcode and without reasoning to the contrary, seems like a good default for rules_swift too.

Note that with Swift 5 this was also enabled for release builds too. https://swift.org/blog/swift-5-exclusivity/


keith:
Note we don't need to explicitly pass this since it's the default in the compiler https://gist.github.com/keith/3f01e1c9b763e9aceb70411927a0c42c

kastiglione:
Thanks, interesting that Xcode still passes it explicitly.

keith:
Yea in theory we could too, I don't have a strong preference on that

PR #227, opened 2019-05-13
Add option to disable proto module generation

steeve:
This PR does the following:
- change `FileNaming=FullPath` to `FileNaming=PathToUnderscores` so that file names don't collide
- add a `pbswift_files` output group to access the generated `.pbswift` files (from a macro, say)
- Allow to disable modules generation via a `modules` attribute

In our case, all the generated proto files are grouped inside the same module, which is embedded in a `.framework` file to be included in Xcode by our iOS team. So we can't use aliases unfortunately.

What we do is use the `pbswift_files` output group to grab all the generated `.pbswift` files and bundle them together in the same `swift_library` target:
```python
filegroup(
    name = "pbswift_files",
    srcs = [<swift_proto_library targets with modules="off">],
    output_group = "pbswift_files",
)

swift_library(
    name = "XXX",
    srcs = [
        ":pbswift_files",
    ],
    module_name = "MyModule",
    deps = ["@com_github_apple_swift_swift_protobuf//:SwiftProtobuf"],
)
```

While this introduces a new API in the form of the `modules` attribute, the default behaviour is kept.

In any case, I think the first two commits can be useful in their own rights, and we'd be fine keeping the last one as a rebased patch.

### DIFF
--- a/swift/internal/proto_gen_utils.bzl
+++ b/swift/internal/proto_gen_utils.bzl
@@ -115,7 +115,7 @@ def _generated_file_path(name, extension_fragment, proto_file = None):
     )
     if proto_file:
         generated_file_path = paths.replace_extension(
-            workspace_relative_path(proto_file),
+            workspace_relative_path(proto_file).replace("/", "_"),
             ".{}.swift".format(extension_fragment),
         )
         return paths.join(dir_path, generated_file_path)
--- a/swift/internal/swift_proto_library.bzl
+++ b/swift/internal/swift_proto_library.bzl
@@ -59,6 +59,15 @@ def _swift_proto_library_impl(ctx):
 
 swift_proto_library = rule(
     attrs = {
+        "modules": attr.string(
+            default = "on",
+            values = ["on", "off"],
+            doc = """
+Controls wether generate individual modules per proto file. Either `on` (the
+current and default behaviour) or `off`. This is mostly useful when using the
+`pbswift_files` output group.
+            """,
+        ),
         "deps": attr.label_list(
             aspects = [swift_protoc_gen_aspect],
             doc = """
--- a/swift/internal/swift_proto_library.bzl
+++ b/swift/internal/swift_proto_library.bzl
@@ -42,8 +42,12 @@ def _swift_proto_library_impl(ctx):
         cc_info,
         swift_info,
         # Repropagate the `SwiftProtoInfo` provider so that downstream rules/aspects can access the
-        # sources if necessary. (This should typically only be used for IDE support.)
+        # sources if necessary. (This should typically only be used for IDE support.) Also, add an
+        # output group to for the generated files.
         swift_proto_info,
+        OutputGroupInfo(
+            pbswift_files = swift_proto_info.pbswift_files,
+        ),
     ]
 
     # Repropagate the apple_common.Objc provider if present so that apple_binary targets link
--- a/swift/internal/swift_protoc_gen_aspect.bzl
+++ b/swift/internal/swift_protoc_gen_aspect.bzl
@@ -258,7 +258,7 @@ def _swift_protoc_gen_aspect_impl(target, aspect_ctx):
     proto_deps = [dep for dep in aspect_ctx.rule.attr.deps if SwiftProtoInfo in dep]
 
     minimal_module_mappings = []
-    if direct_srcs:
+    if direct_srcs and aspect_ctx.attr.modules == "on":
         minimal_module_mappings.append(
             _build_module_mapping_from_srcs(target, direct_srcs),
         )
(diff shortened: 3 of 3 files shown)


allevato:
This change is essentially counter to Bazel's build model, which is that a library is a module and there's an implied assumption that everything is built with Bazel. I don't think it makes sense to increase the maintenance burden of the rules to add support for a niche use case where you want to build a framework with Bazel that gets fed into Xcode, which would involve new attributes and changes to the flags passed into `protoc-gen-swift`.

Since you need something custom, I think you should write a custom rule to generate the protos with the flags that you need, and then feed those generated sources into a `swift_library`. Then you don't need to make any changes here.

However, keep in mind that those generated protos will be incompatible with the same generated protos from the actual `swift_proto_library` targets, if you ever wanted to mix them, since they'd be in separate modules.

steeve:
I'm not sure building a framework with Bazel that gets fed into Xcode is a niche use case. There's already `ios_static_framework` and, in fact there is ongoing work to expand that use case at https://github.com/bazelbuild/rules_apple/issues/355.

Frameworks artefacts being a good fit for distribution between teams, and I don't think it's wise (or even possible) to expect everybody to adopt Bazel at the same time inside an organisation.

Regarding `PathToUnderscores`, I feel it's pretty harmless, but also fixes a bug where the linker complained about the file names being the same, so I guess it's nice to have anyway. The output_group also has value in its own right, should people want to get the generated files from command line (in a CI context, for instance).

Finally, I do agree that the `modules` attr is a though sell. But at the same time I feel it solves a valid use case in a real context, with a very low footprint here. As I said though, I'm fine vendoring and rebasing a patch in our codebase (as we're doing today).

keith:
To tack on a bit here, this is the second time this week that not using the rules_swift proto library rule has come up. The other was in a conversation between @allevato and I about how we want to use our own plugin instead of the default one. If the course of action here is to write our own rules for customizing things like this, it would be nice if the implementation of the rule was more extensible so it didn't rely on consumers re-writing every piece of this. Especially since any external rules are still subject to rules_swift changes because of the integration with swift_library.

steeve:
@keith could consuming the `output_group` and wrapping with `swift_library` (as we did here) would have satisfied your use case ?

keith:
This wouldn't satisfy our use case since we want to entirely swap out the plugin for our custom one

tinder-maxwellelliott:
Sorry to revive an old thread here, but was is the guidance here for developers using `proto_library` and `swift_proto_library`? I see the following in this https://blog.bazel.build/2017/02/27/protocol-buffers.html

> 1. One proto_library rule per .proto file.
> 2. A file named foo.proto will be in a rule named foo_proto, which is located in the same package.
> 3. A <lang>_proto_library that wraps a proto_library named foo_proto should be called foo_<lang>_proto, and be located in the same package.

So for a team that currently does this in some xcodebuild build phase (`protoc` sources directly during build) would need to:
1. Go to their proto repo and add BUILD files for all models via `proto_library`, they would also need to figure out `deps` and such between all protos.
2. Create `proto_swift_libraries` for the protos they wish to consume
3. Add external dependencies to the WORKSPACE pull these build definitions into the build (we still need to solve how to do this in parallel with Cocoapods, since most teams are coming from systems like that and will want to share artifacts to prevent divergence in the systems (if doing parallel builds, most likely))

This is a **significant** departure from the current proto workflow, it requires a much more active role of the mobile developers understanding the protos. One should assume that this part of the app is "Dark Debt" and nearly nobody knows how it works or has an active interest into learning more beyond "The code appears in the editor".

What would be nice is if we could potentially figure out how to make the `PathToUnderscores` change an option on the aspect (This solves an issue for many teams around name collisions on proto filenames). I believe that making a proto_library like so would be much more approachable for the developers today, even if it is not "perfect":

```python
load(
    "@rules_proto//proto:defs.bzl",
    "proto_library"
)
load(
    "@build_bazel_rules_swift//swift:swift.bzl",
    "swift_proto_library"
)

proto_library(
    name = "mymodule_protos",
    srcs = glob(["Protos/**/*.proto"]), # Against the rules, I know
)

swift_proto_library(
    name = "mymodule_swift_protos",
    deps = [
        ": mymodule_protos",
    ]
)
```

In closing I believe more and more people are trying to onboard onto Bazel and they are burdened by these existing and previously ignored issues in their builds. One route is to just fix everything, rename all the protos to prevent collisions, commit all the build files you need to, force everyone to write BUILD files. The other is to get something working and prove value early. Moves like this would allow more teams to realize the improvements to their builds via Bazel.

keith:
The current proto rule is definitely optimizing for how google does things, and expects you to convert to be the same way. I'm not sure what the right path forward is with it because I haven't spent much time looking at those rules, but I wonder if it makes sense to split the rules instead of trying to shoe-horn more standard proto workflows into it

tinder-maxwellelliott:
Thanks for the response @keith 

I think long term we will do things to google way. The only issue I see is now is that I do not believe these modules can be exported to other build systems, so we are blocked if that is the case until we depreciate cocoapods.

PR #232, opened 2019-05-16
Request cloning using COPYFILE_CLONE

kastiglione:
This should be a safe simple addition. From man `copyfile(3)`:

> Try to clone the file instead.  This is a best try flag i.e. if cloning fails, fallback to copying the file.

### DIFF
--- a/tools/common/file_system.cc
+++ b/tools/common/file_system.cc
@@ -33,7 +33,8 @@ bool CopyFile(const std::string &src, const std::string &dest) {
 #ifdef __APPLE__
   // The `copyfile` function with `COPYFILE_ALL` mode preserves permissions and
   // modification time.
-  return copyfile(src.c_str(), dest.c_str(), nullptr, COPYFILE_ALL) == 0;
+  return copyfile(src.c_str(), dest.c_str(), nullptr,
+                  COPYFILE_ALL | COPYFILE_CLONE) == 0;
 #elif __unix__
   // On Linux, we can use `sendfile` to copy it more easily than calling
   // `read`/`write` in a loop.
--- a/tools/worker/work_processor.cc
+++ b/tools/worker/work_processor.cc
@@ -160,8 +160,6 @@ void WorkProcessor::ProcessWorkRequest(
   if (!is_wmo) {
     // Copy the output files from the incremental storage area back to the
     // locations where Bazel declared the files.
-    // TODO(allevato): Investigate copy-on-write on macOS, or hard-linking in
-    // general, as a possible optimization.
     for (auto expected_object_pair : output_file_map.incremental_outputs()) {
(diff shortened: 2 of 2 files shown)


tools/common/file_system.cc
```diff
@@ -33,7 +33,8 @@ bool CopyFile(const std::string &src, const std::string &dest) {
 #ifdef __APPLE__
   // The `copyfile` function with `COPYFILE_ALL` mode preserves permissions and
   // modification time.
-  return copyfile(src.c_str(), dest.c_str(), nullptr, COPYFILE_ALL) == 0;
+  return copyfile(src.c_str(), dest.c_str(), nullptr,
+                  COPYFILE_ALL | COPYFILE_CLONE) == 0;
```
keith: Does this need to be gated on an OS version?

kastiglione:
Tested with our build, checked with `--swiftcopt=-driver-show-incremental`

tools/common/file_system.cc
kastiglione: ¯\\\_(ツ)_/¯ Nothing in `copyfile.h` mentions versions, there are no uses of `availability` attributes.

kastiglione: We could add a `ifdef COPYFILE_CLONE` to be extra safe.

keith: I don't have a machine on any older macOS version than Mojave but we don't expect it to be on any older version since it's related to APFS do we?

allevato: For what it's worth, LLVM supports this and checks that it's on macOS 10.12 or greater: https://reviews.llvm.org/rL360174

kastiglione: Does the worker support 10.11 or earlier?

allevato: And to finish my thought, I don't know if we care about anyone running such an old version of macOS. I'd say leave it unconditional, and if someone hits a bug because of it, we can patch it then.

allevato: How did we both manage to wait an hour to reply at exactly the same time?

allevato:
I initially had some concerns about the clone behavior of "fails if the file already exists", but as long as we can assume that Bazel does the appropriate cleanup of the copy destination (whether sandboxed or not), this _should_ be safe.

tools/common/file_system.cc
kastiglione: maybe it was lunch 😋

kastiglione:
Should we add a fallback second call to `copyfile()` without `_CLONE` to handle that case?

allevato:
> Should we add a fallback second call to `copyfile()` without `_CLONE` to handle that case?

I'm not that worried about it right now. I've also already started the internal submit that will merge this 😛

kastiglione:
Thanks

PR #233, opened 2019-05-16
Set worker exit_code to failure if copying files fails

kastiglione:
Follow up to #232. If copying output files fails, mark the worker response as having failed.

### DIFF
--- a/tools/worker/work_processor.cc
+++ b/tools/worker/work_processor.cc
@@ -164,6 +164,7 @@ void WorkProcessor::ProcessWorkRequest(
       if (!CopyFile(expected_object_pair.second, expected_object_pair.first)) {
         std::cerr << "Could not copy " << expected_object_pair.second << " to "
                   << expected_object_pair.first << " (errno " << errno << ")\n";
+        exit_code = EXIT_FAILURE;
       }
     }
   }


kastiglione:
I noticed this just as you said the other PR was mid merge.

tools/worker/work_processor.cc
```diff
@@ -164,6 +164,7 @@ void WorkProcessor::ProcessWorkRequest(
       if (!CopyFile(expected_object_pair.second, expected_object_pair.first)) {
```
kastiglione: Should copying happen unconditionally? Or only when `RunSubProccess` returns 0?

allevato: I don't think it matters; the build will fail either way and the outputs don't matter.

kastiglione: To put it another way, in the case of compilation error, should outputs be copied, or no? Currently this change has the potential to clobber the `exit_code` from `RunSubProccess()`, if it tries to copy outputs and fails.

kastiglione: Should I add an `if` to prevent clobbering the `exit_code` of the failed `RunSubProccess()`?

kastiglione: I don't even know if `swiftc` has meaningful exit codes.

allevato: Hmm... I suppose that's a rare possibility. Follow-up PR? Because I've already started merging this one too. 

(If you're gonna ask questions about a PR after uploading it, use the Draft PR feature because otherwise I assume they're ready to merge and I like to burn through these small ones quickly 😄)

kastiglione: I will keep that in mind. Please know I didn't intend to be thrashy.

allevato: It's all good. You probably didn't expect me to be this on top of the PRs given my history 😉 

The automated workflow has been a _huge_ help.

kastiglione: Haha, I did not expect such instant turn arounds.

From a look at the swiftc driver, it seems it is using returning 0/1 for the most part.

allevato: Yeah, I think it's safe to rely on `swiftc` returning non-zero if it fails. Otherwise, we wouldn't be able to rely on Bazel at all to stop the build after a bad action.

kastiglione: What I meant was does swiftc ever return non-zero values that have meaning like -2 meaning one thing an -5 meaning another.

allevato: I don't believe so, but in any case I think we'd want to treat any non-zero code as a failure and not try to do something more complex because if we were invoking `swiftc` directly from Bazel, that's what it would do with the error code.

Issue #94, opened 2018-11-04
How to use bridging headers?


codesuki:
I have a project that uses `[project]-Bridging-Header.h` in XCode to `import` some headers. Those headers seem to become implicitly available to all Swift source files. Therefore most of them miss imports like `import Foundation`, `import UIKit`.

The good thing to do would probably be to add all imports, but I want to try and compile the project as is, without changes, is that possible?

How can one use bridging headers?

keith:
You can try using these by manually adding the swiftc argument to the library's copts

codesuki:
That's a great idea! I went ahead and fixed all the imports so I got it working, but I'll try the `copts` method just for the sake of trying. Thank you!

PR #235, opened 2019-05-20
Fix copts and linkopts relying on input files

FuegoFro:
In https://github.com/bazelbuild/rules_swift/commit/32275ca0536eca5610a915f7b26cb841ca0031a3 when the additional_input local variable was introduced, this went from passing `ctx.attr.swiftc_inputs` to passing `ctx.files.swiftc_inputs` into `expand_locations`. This breaks, since the underlying `expand_location` only allows `Target` types, resulting in the following error when passing any file-like target to `swiftc_inputs`:

```
expected type 'Target' for 'targets' element but got type 'File' instead
```

This switches the two `expand_locations` calls back to using `ctx.attr.swiftc_inputs`.

### DIFF
--- a/swift/internal/swift_library.bzl
+++ b/swift/internal/swift_library.bzl
@@ -52,8 +52,10 @@ def _maybe_parse_as_library_copts(srcs):
 
 def _swift_library_impl(ctx):
     additional_inputs = ctx.files.swiftc_inputs
-    copts = expand_locations(ctx, ctx.attr.copts, additional_inputs)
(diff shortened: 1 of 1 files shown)


FuegoFro:
Thank you for the quick response! I'm using a custom bridging header to rely on Objective-C code 🙂

I was following the advice from https://github.com/bazelbuild/rules_swift/issues/94 and using
```bzl
swift_library(
    ...
    swiftc_inputs = [":<my-header-location>"],
    copts = ["-import-objc-header", "$(location :<my-header-location>)"],
)
```

PR #236, opened 2019-05-20
Remove the `-Wl,-z,notext` workaround.

swiple-rules-gardener:
Remove the `-Wl,-z,notext` workaround.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -281,12 +281,6 @@ def register_link_executable_action(
         before_each = "-framework",
     )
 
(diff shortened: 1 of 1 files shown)


keith:
What bazel change do we have to have to be compatible with this?

allevato:
This isn't tied to a Bazel version change. I'd be surprised if it ever affected you in the first place; it was a workaround for a very specific case of linking with `lld` and an internal library propagating that linker flag, but that library is no longer propagating the flag.

(Of course, I'm running my internal tests now to verify that, but the act of running the presubmit caused this to get pushed out.)

PR #243, opened 2019-05-29
Clean up clang usage for linking executables.

swiple-rules-gardener:
Clean up clang usage for linking executables.

With all the new C++ APIs in Bazel, we should no longer be grabbing the clang executable separately from a legacy provider field. Instead, use cc_common to get the tool path and command line as we now do for the archiver.


allevato:
Currently blocked on Bazel pending https://github.com/bazelbuild/bazel/pull/8495.

allevato:
This has been superseded by #313.

PR #246, opened 2019-06-01
Add -module-name flag outside of params file

kastiglione:
Adding the `-module-name` flag outside of the params file makes it more convenient to introspect on commands/processes using system tools and APIs.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -606,7 +606,7 @@ def _compile(
 
     run_swift_action(
         actions = actions,
-        arguments = [tool_args, args],
+        arguments = [tool_args, "-module-name", module_name, args],
         execution_requirements = execution_requirements,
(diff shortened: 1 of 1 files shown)


thomasvl:
Isn't there just a flag to have bazel dump what goes into the args files?  If you are debugging something that seems better then trying to craft what goes in/out of the args file.

allevato:
Can you describe better what you're trying to? "Introspect on commands using system tools and APIs" is pretty vague; it would help to know what you're trying to process.

allevato:
Ping—what are you trying to do here?

kastiglione:
Apologies. So basically with any process aware tool, from simple `ps` to the more sophisticated `dtrace` backed tools, it's impossible to distinguish one swiftc compile from another. Putting the module name outside of the args file helps this.

PR #248, opened 2019-06-01
Also pass header search paths to Clang when compiling Swift.

chenxiao0228:
If Swift imports an ObjC module, when preprocessing C headers Clang needs
the header search paths too.


allevato:
Do you have an example of a bug you're hitting without this? The implementation of ClangImporter in `swiftc` should already be configuring the `CompilerInvocation` with the search paths passed via `-I` without the need to pass them separately with `-Xcc`.

chenxiao0228:
Thanks for the quick response. It’s probably the header maps we are passing
in via -I..

I’ll work on a minimum project.

allevato:
We chatted in person at WWDC; IIRC the resolution here was to simply pass the header maps to `swiftc` using `-Xcc -I` instead of `-I` so that they are passed directly to Clang.

PR #260, opened 2019-06-19
Add cc_library deps to the force_load list if they have a .lo extension

### LINKED ISSUE #259: alwayslink=True on a cc_library dep is not fully respected
The library itself gets the `.lo` extension, but is put into the `.objlist` file and not linked with `-force_load`.

jfaust:
Fixes #259


allevato:
We had to revert this in https://github.com/bazelbuild/rules_swift/pull/261 because some of our internal  projects broke. I'll have to dig into the cause and then resubmit this with a fix.

jfaust:
OK - can you re-open #259?

I was wondering if having the library in both the force list and the `library` depset would cause problems - it didn't in our (simple) use case, but depending on what problem you're seeing that might be an avenue to explore.

PR #290, opened 2019-08-21
Add a pbswift_files output group and use PathToUnderscores

steeve:
Add a new `pbswift_files` output group in order to consume the generated `pbswift` files in macros for instance

Switch from `FileNaming=FullPath` to `FileNaming=PathToUnderscores`
When compiling `pbswift` files, the linker complains if two files have the same name. This fixes it.


tinder-maxwellelliott:
@steeve is there any way to accomplish the `PathToUnderscores` change via a feature flag or something? Looking for options here as well since I of course have duplicated proto filenames....

PR #298, opened 2019-08-29
Make rules_swift work with virtual import directories.

lberki:
This is necessary for the rule set to work with
`--incompatible_generated_protos_in_virtual_imports` (and also with
`proto_library` rules with `strip_import_prefix` / `import_prefix`)

### DIFF
--- a/swift/internal/proto_gen_utils.bzl
+++ b/swift/internal/proto_gen_utils.bzl
@@ -105,6 +105,7 @@ def _generated_file_path(name, extension_fragment, proto_source_root, proto_file
         extension_fragment: An extension fragment that precedes `.swift` on the end of the
             generated files. In other words, the file `foo.proto` will generate a file named
             `foo.{extension_fragment}.swift`.
+        proto_source_root: The source root for the `.proto` file.
         proto_file: The `.proto` file whose generated `.swift` path should be computed.
 
     Returns:
--- a/swift/internal/proto_gen_utils.bzl
+++ b/swift/internal/proto_gen_utils.bzl
@@ -15,9 +15,9 @@
 """Utility functions shared by rules/aspects that generate sources from .proto files."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
-load(":utils.bzl", "workspace_relative_path")
+load(":utils.bzl", "proto_import_path")
 
-def declare_generated_files(name, actions, extension_fragment, proto_srcs):
+def declare_generated_files(name, actions, extension_fragment, proto_source_root, proto_srcs):
     """Declares generated `.swift` files that correspond to a list of `.proto` files.
 
     Args:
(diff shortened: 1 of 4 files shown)


thomasvl:
From the bazel issue, I thought the right way of doing this was going to be using the root from the provider and not hardcoding things?

lberki:
Do I understand correctly that I meticulously told you how to do things the right way then I did them the wrong way anyway? I'm talented like that...

Fixed, PTAL.

lberki:
I'd be happy to, but how do I know if I succeeded? The tests pass with this change, too.

thomasvl:
> I'd be happy to, but how do I know if I succeeded? The tests pass with this change, too.

This is one of the problems trying to write tests for rules/bazel.  You need a custom workspace with another repo reference that uses the rules.  They are pretty hard to set up, so most rules sets don't seem to have them.

lberki:
So what do I do? Is there any prior art as to what needs to work? (I thought `//examples/...` would be enough, which is what Bazel's BuildKite instance does)

lberki:
Updated the above places and checked whether rules work when imported from a remote repository and with generated `.proto` files. Please take another look!

swift/internal/utils.bzl
```diff
@@ -251,3 +251,20 @@ def workspace_relative_path(file):
     """
```
thomasvl: Is anything left using workspace_relative_path or should it be made private to this file?

lberki: You cannot know that -- anyone can `load()`s this file and use this function from it. If you wish, I can remove it, but be aware that it's a potential incompatible change.

thomasvl: > You cannot know that -- anyone can `load()`s this file and use this function from it. If you wish, I can remove it, but be aware that it's a potential incompatible change.

Actually, it is in an `internal` directory, so we're fine breaking folks that directly tried to use it.  It just needs to be not used by anything else within _rules_swift_.  The directory naming is all we can do until bazel has visibility for _bzl_ files.

swift/internal/proto_gen_utils.bzl
```diff
@@ -88,7 +91,7 @@ def register_module_mapping_write_action(name, actions, module_mappings):
 
     return mapping_file
 
-def _generated_file_path(name, extension_fragment, proto_file = None):
+def _generated_file_path(name, extension_fragment, proto_source_root, proto_file = None):
     """Returns the short path of a generated `.swift` file corresponding to a `.proto` file.
 
     The returned workspace-relative path should be used to declare output files so that they are
```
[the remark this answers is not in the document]
lberki: My understanding is that this is still true -- the format is `PACKAGE/NAME.protoc_gen_EXTENSION_swift/IMPORT_PATH.<something>`, to which all of this is true. Except that the path now contains the import path and not the workspace-relative path of the source file.

swift/internal/utils.bzl
lberki: So be it then!

Issue #266, opened 2019-06-25
Add support for Xcode 11's XCTUnwrap


keith:
As of Xcode 11 there are some new XCTest interfaces that are exposed with a new XCTest.swiftmodule. Xcode now passes a new argument for test bundles to this path:

```
-I /Applications/Xcode-11.0.0b2.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/usr/lib
```

To support this I think we'll have to do something like this https://github.com/bazelbuild/rules_swift/blob/770b1fe98693631e72a3b9cd5df4e2ae3d9c76ea/swift/internal/xcode_swift_toolchain.bzl#L134-L138

But for the swiftc opts https://github.com/bazelbuild/rules_swift/blob/770b1fe98693631e72a3b9cd5df4e2ae3d9c76ea/swift/internal/xcode_swift_toolchain.bzl#L142-L171

Here's a small repro project: [xcunwrap.zip](https://github.com/bazelbuild/rules_swift/files/3323431/xcunwrap.zip)

With Xcode 11 installed, run `./repro.sh` to see the issue

keith:
In the beta 5 release notes this is kind of marked as a known issue:

```
The XCTUnwrap API is only available in primary test bundle targets and not in other libraries or frameworks. (51117167)
Workaround: Move any library code that makes use of XCTUnwrap to your primary test bundle target or manually modify the following build settings in affected targets:
SYSTEM_FRAMEWORK_SEARCH_PATHS = "$(inherited) $(PLATFORM_DIR)/Developer/Library/Frameworks"
LIBRARY_SEARCH_PATHS = "$(inherited) $(PLATFORM_DIR)/Developer/usr/lib"
SWIFT_INCLUDE_PATHS = "$(inherited) $(PLATFORM_DIR)/Developer/usr/lib"
```

keith:
Seems like that was only for Xcode

PR #300, opened 2019-08-30
Pass new test library directory to builds

keith:
With Xcode 11 there is a new location where there are Swift specific
libraries and swiftmodules. Specifically they contain the new
`XCTUnwrap` function. This change passes these arguments to the compiler
and the linker.

For iOS tests this also requires https://github.com/bazelbuild/rules_apple/pull/572

Fixes https://github.com/bazelbuild/rules_swift/issues/266

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -37,15 +37,21 @@ load(
 load(":providers.bzl", "SwiftToolchainInfo")
 
 def _swift_developer_lib_dir(platform_framework_dir):
-    """Returns the directory contain extra Swift specific developer libraries
+    """Returns the directory containing extra Swift developer libraries.
 
     Args:
-        platform_framework_dir: The developer platform framework directory for the current platform
+        platform_framework_dir: The developer platform framework directory for
+            the current platform.
 
     Returns:
-        The directory containing Swift specific development libraries and swiftmodules
+        The directory containing extra Swift-specific development libraries and
+        swiftmodules.
     """
-    return paths.join(paths.dirname(paths.dirname(platform_framework_dir)), "usr", "lib")
+    return paths.join(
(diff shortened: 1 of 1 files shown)


swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -35,6 +36,17 @@ load(
...
+def _swift_developer_lib_dir(platform_framework_dir):
+    """Returns the directory contain extra Swift specific developer libraries
+
+    Args:
+        platform_framework_dir: The developer platform framework directory for the current platform
+
+    Returns:
+        The directory containing Swift specific development libraries and swiftmodules
+    """
+    return paths.join(paths.dirname(paths.dirname(platform_framework_dir)), "usr", "lib")
```
keith: I'd love to hear any other suggestions for constructing this path 😬

allevato: Sigh, I wish they just gave us `platform_dir` without the framework business. This is probably the best for now, and there will be cleanup opportunities in the future if we can ever move those placeholder strings and utility functions to apple_support or something.

keith: sg

keith:
@allevato should this one have merged automatically?

keith:
(I rebased it)

allevato:
It needed another internal approval and got lost in the shuffle. Thanks for pinging!

keith:
idk what happened here but it got merged 90995572723ed47cbf2968480db250e97a9f5894

keith:
We should revert this because it breaks UI tests since they don't bundle this library in the runner app but it's expected to be there

keith:
More conversation about it in slack

keith:
Spoke too soon, it's broken with or without this, so I think we should keep it

PR #303, opened 2019-09-05
Add a rapid development feature to make it easier to build custom Swift compilers and throw them at the build rules without building a complete toolchain.

swiple-rules-gardener:
Add a rapid development feature to make it easier to build custom Swift compilers and throw them at the build rules without building a complete toolchain.

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -143,16 +143,23 @@ def _default_linker_opts(
 
     return linkopts
 
-def _default_swiftc_copts(apple_fragment, apple_toolchain, target):
+def _default_swiftc_copts(
+        apple_fragment,
+        apple_toolchain,
+        target,
+        toolchain_root):
     """Returns options that should be passed by default to `swiftc`.
 
     Args:
         apple_fragment: The `apple` configuration fragment.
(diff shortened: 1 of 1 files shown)


swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -163,14 +170,24 @@ def _default_swiftc_copts(apple_fragment, apple_toolchain, target):
         apple_toolchain.platform_developer_framework_dir(apple_fragment),
     ]
 
+    if toolchain_root:
+        copts.extend([
+            "-resource-dir",
+            "{developer_dir}/Toolchains/{toolchain}.xctoolchain/" +
+            "usr/lib/swift".format(
+                developer_dir = apple_toolchain.developer_dir(),
```
keith: should this be `toolchain_root`?

allevato: No, as the comment further down states, this (rare) feature is for when you just want to use a different custom built compiler binary rather than a full toolchain. In that case, you need to pass the resource dir of the _original_ toolchain so that the libraries are found correctly, otherwise the frontend will try to search relative to the location of that custom binary, where they probably don't exist.

I'll toss another comment up here that clarifies that.

PR #304, opened 2019-09-05
Capture swiftc parseable output to file

kastiglione:
This change adds support for `-parseable-output`, which prints a sequence of size delimited json events to `stderr`. For more info, see the apple/swift docs: [docs/DriverParseableOutput.rst](https://github.com/apple/swift/blob/master/docs/DriverParseableOutput.rst).

Currently, using `-parseable-output` results in the json being printed to Bazel's `stdout`. There are two problems:

1. the volume of parseable output is relatively quite large – output from Bazel and other tools becomes much harder to see
2. the parseable output is meant for other tools, and mixing it with all other bazel output makes it very difficult to extract

The solution here is to change the worker to capture `stdout` and `stderr` separately, but only if the `-parseable-output` flag is present. Then, the captured `stderr` is written to a log file. Consumers can then make use of this log file, in whatever way they need. In this case, only the `stdout` will be printed to Bazel's output stream, and this has shown to be empty in my observations.

Without the `-parseable-output` flag, behavior remains as it was: all `stdout` and `stderr` are captured by the worker and returned to bazel, where it prints it to the console.

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -594,6 +594,17 @@ def _compile(
     additional_outputs = []
 
     args = actions.args()
+
+    parseable_output_file = derived_files.parseable_output(
+        actions = actions,
+        target_name = target_name,
+    )
+    additional_outputs.append(parseable_output_file)
+
+    args.add("-Xwrapped-swift=-parseable-output-path={}".format(
+        parseable_output_file.path
+    ))
+
     if is_feature_enabled(
         feature_configuration = feature_configuration,
         feature_name = SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE,
--- a/swift/internal/derived_files.bzl
+++ b/swift/internal/derived_files.bzl
@@ -134,6 +134,18 @@ def _objc_header(actions, target_name):
     """
     return actions.declare_file("{}-Swift.h".format(target_name))
 
+def _parseable_output(actions, target_name):
+    """TODO
+
+    Args:
+        actions: The context's actions object.
+        target_name: The name of the target being built.
+
+    Returns:
+        The declared `File`.
+    """
+    return actions.declare_file("{}.compile.log".format(target_name))
+
 def _partial_swiftmodule(actions, target_name, src):
     """Declares a file for a partial Swift module created during compilation.
 
@@ -297,6 +309,7 @@ derived_files = struct(
     module_map = _module_map,
     modulewrap_object = _modulewrap_object,
     objc_header = _objc_header,
+    parseable_output = _parseable_output,
     partial_swiftmodule = _partial_swiftmodule,
     reexport_modules_src = _reexport_modules_src,
     static_archive = _static_archive,
--- a/tools/common/path_utils.cc
+++ b/tools/common/path_utils.cc
@@ -30,16 +30,6 @@ std::string Dirname(const std::string &path) {
   return path.substr(0, last_slash);
 }
 
-std::string PathJoin(const std::string &tail, const std::string &head) {
-  if (tail.back() == '/') {
-    return PathJoin(tail.substr(0, tail.size() - 1), head);
-  }
-  if (head.front() == '/') {
-    return PathJoin(tail, head.substr(1));
-  }
-  return tail + '/' + head;
-}
-
 std::string ReplaceExtension(const std::string &path,
                              const std::string &new_extension,
                              bool all_extensions) {
--- a/tools/common/path_utils.cc
+++ b/tools/common/path_utils.cc
@@ -30,6 +30,16 @@ std::string Dirname(const std::string &path) {
   return path.substr(0, last_slash);
 }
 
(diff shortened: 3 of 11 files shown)


kastiglione:
Some thoughts and questions on implementation:

1. @Keith suggested the worker could always add `-parseable-output` and always capture it to file, instead of making this conditional on the presence of `-parseable-output`
2. @Keith also suggested that the worker could extract from the json, _just_ the diagnostic message(s) and return those as the output of the `WorkResponse`
3. This currently writes the parseable output to an undeclared file, should it be a declared output?
4. The path of the log file is conveniently (arbitrarily) made as a sibling to the output file map, this is perhaps a hack. Should it change and if so to what?
5. Instead of creating a pipe, polling it, reading into a buffer, writing to an `ostringstream`, and then writing to file, a less round about and more direct implementation could be to set the child process's `stderr` descriptor to that of the destination log file. I followed the current implementation, but I thought I'd mention this thought.

kastiglione:
I deferred adding docstrings until getting some initial feedback.

allevato:
From a quick glance, this looks nice. Feel free to go ahead and start documenting the hell out of it. 😄  I'll go ahead and answer your questions below, and give it a deeper review based on those answers/changes.

Out of curiosity, what's your use case? I have a couple situations in mind where this could come in handy (combined with serialized `.dia` files, but _let's not go there yet_), but I'd like to understand more about what your goal is to make sure I'm reviewing it in the right context.

To answer your questions above:

> 1. @keith suggested the worker could _always_ add `-parseable-output` and always capture `stderr` to file, instead of making this conditional on the presence of `-parseable-output`
> 2. @keith also suggested that the worker could extract from the json, _just_ the diagnostic message(s) and return those as the output of the `WorkResponse`

Since the worker is _already_ collecting all of the `stderr` output in memory and shoving it into the response proto, this sounds totally reasonable—we'd just be moving to a different way of grabbing the same string, more or less. Making it unconditional sounds like the way to go, to reduce the number of code paths that we have to worry about maintaining.

> 3. This currently writes the parseable output to an undeclared file, should it be a declared output?

Yes, otherwise Bazel won't download it from remote build environments after the action has finished executing.

> 4. The path of the log file is conveniently (arbitrarily) made as a sibling to the output file map, this is perhaps a hack. Should the log path change, and if so to what?

Since we're going to make it a declared output, add a new function to the `derived_files` module. My off-the-top-of-my-head recommendation would be `"{}.compile.log".format(target_name)`. Then, you'll need to introduce a new `-Xwrapped-swift` flag to pass the path into the worker, something like `-Xwrapped-swift=-parseable-output-path=<whatever>`.

> 5. Instead of creating a pipe, polling it, reading into a buffer, writing to an `ostringstream`, and then writing to file, a less roundabout and more direct implementation could be to set the child process's `stderr` descriptor to that of the destination log file. I followed the current implementation, but I thought I'd mention this thought.

Let's keep it in memory first. If we're going to extract the diagnostic output from the JSON and return that in the work response, we'll want to instantiate the JSON object without reading it back from disk first. (If there are any opportunities to reduce buffer churn, we can look into that separately, perhaps.)

kastiglione:
> Out of curiosity, what's your use case? I have a couple situations in mind where this could come in handy (combined with serialized `.dia` files, but _let's not go there yet_), but I'd like to understand more about what your goal is to make sure I'm reviewing it in the right context.

Warning: I'm going there 😉 

cc @DavidGoldman 

In our Xcode integration, I have prototyped the following:

1. Use this branch
2. Add `--swiftcopt=-serialize-diagnostics`
3. Set the `SWIFT_EXEC` build setting to the path of a custom script
3. The custom script waits for for the parseable output file, and writes it to `stderr`

This enables exactly what you said, for Xcode to pick up the `.dia` file built by `swiftc`. In particular this enables Fix-Its. From a Run Script, Xcode will only pick up build errors, not the suggested fixes.

It does open the door to other custom Fix-Its too, when you want to go there haha.

This also opens up a (back)door into Xcode's build log. The frontend actions displayed in Xcode's build log are driven from the parseable output. It would be possible to simulate some amount of "native looking" logging by sending custom events to the parseable output.

kastiglione:
Related, @Keith has filed https://bugs.swift.org/browse/SR-11431

brentleyjones:
@kastiglione where did you end up with this? Having Fix-Its work in Xcode would be wonderful 😃.

kastiglione:
@brentleyjones soon, I hope. I wanna say the week after next.

kastiglione:
I'm back at this. I've pushed some commits but no rush to look, I have to rebase against `master`, test it more, and document things.

brentleyjones:
:wave: Just pinging this to refresh people's memories of it.

brentleyjones:
@kastiglione Do y'all still use a version of this in your Xcode integration?

kastiglione:
Update: I don't plan to work on this in the near future.

PR #315, opened 2019-09-25
Add support for swiftinterface files to swift_library.

swiple-rules-gardener:
Add support for swiftinterface files to swift_library.

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -31,6 +31,7 @@ load(
     "SWIFT_FEATURE_DEBUG_PREFIX_MAP",
     "SWIFT_FEATURE_ENABLE_BATCH_MODE",
     "SWIFT_FEATURE_MODULE_MAP_HOME_IS_CWD",
+    "SWIFT_FEATURE_SUPPORTS_LIBRARY_EVOLUTION",
     "SWIFT_FEATURE_USE_RESPONSE_FILES",
     "features_for_build_modes",
 )
@@ -374,6 +375,10 @@ def _xcode_swift_toolchain_impl(ctx):
     # Xcode 10.2 implies Swift 5.0.
     if _is_xcode_at_least_version(xcode_config, "10.2"):
         requested_features.append(SWIFT_FEATURE_DEBUG_PREFIX_MAP)
+
+    # Xcode 11.0 implies Swift 5.1.
+    if _is_xcode_at_least_version(xcode_config, "11.0"):
+        requested_features.append(SWIFT_FEATURE_SUPPORTS_LIBRARY_EVOLUTION)
 
     command_line_copts = _command_line_objc_copts(ctx.fragments.objc) + ctx.fragments.swift.copts()
 
--- a/swift/BUILD
+++ b/swift/BUILD
@@ -2,6 +2,8 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
 
+load("@bazel_skylib//rules:common_settings.bzl", "bool_setting")
+
 exports_files([
     "swift.bzl",
 ])
@@ -16,4 +18,10 @@ filegroup(
     visibility = [
         "//:__pkg__",
     ],
+)
+
+# Configuration setting for enabling the generation of swiftinterface files.
+bool_setting(
+    name = "emit_swiftinterface",
+    build_setting_default = False,
 )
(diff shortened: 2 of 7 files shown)


swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -375,6 +377,12 @@ def _xcode_swift_toolchain_impl(ctx):
     if _is_xcode_at_least_version(xcode_config, "10.2"):
         requested_features.append(SWIFT_FEATURE_DEBUG_PREFIX_MAP)
 
+    # Xcode 11.0 implies Swift 5.1.
+    if _is_xcode_at_least_version(xcode_config, "11.0"):
+        # SWIFT_FEATURE_LIBRARY_EVOLUTION is required for SWIFT_FEATURE_SWIFTINTERFACE.
+        requested_features.append(SWIFT_FEATURE_LIBRARY_EVOLUTION)
+        requested_features.append(SWIFT_FEATURE_SWIFTINTERFACE)
```
keith: Shouldn't we not be enabling this by default?

keith:
I think it would be great if as part of this we wired this up for use with normal swift_library targets and have them depend on this instead of the swiftmodule files from their dependencies. This will greatly reduce unnecessary recompiles when changing private / internal members of modules

swift/internal/xcode_swift_toolchain.bzl
allevato: That's currently being addressed.

allevato:
> I think it would be great if as part of this we wired this up for use with normal swift_library targets and have them depend on this instead of the swiftmodule files from their dependencies.

Since swiftinterface files currently require library evolution, that would change semantics in undesirable ways, like requiring uses of enums from external modules to have `@unknown default`, even when all targets are built from source (e.g., in a monorepo) in the same build graph and there's no need for it.

> This will greatly reduce unnecessary recompiles when changing private / internal members of modules

That alone won't do it; we also need to separate the object generation and module generation into separate actions. But today, that's risky: for Swift modules with large Objective-C dependency subgraphs, the compiler spends the vast majority of its execution time (95% or more) in name binding, with most of that time being spent in ClangImporter. We have some internal targets where that execution time is 1–2 minutes, with almost all of that churning through C headers.

Splitting the compile into two actions would mean that we'd have two actions doing that same churn. It might be counterbalanced by better caching in the incremental case, but it really drags down the non-incremental case.

I'll be more comfortable with that approach if we can precompile C dependencies into PCMs and propagate those instead of the text module maps and headers first.

The approach of propagating swiftinterface files instead of swiftmodules also ends up pushing more work (typechecking, some SILGen and optimization) into the `N` downstream dependencies instead of the single compilation of the module itself, so you could end up regressing performance depending on how much inlineable code you have. Can you cite anything concrete for "greatly reduc[ing]" recompiles, and the conditions that holds under?

keith:
> Since swiftinterface files currently require library evolution, that would change semantics in undesirable ways, like requiring uses of enums from external modules to have @unknown default, even when all targets are built from source (e.g., in a monorepo) in the same build graph and there's no need for it.

Definitely a huge downside, but I figure if we go ahead and wire this up now, off by default, once this is fixed on the Swift side we can drop the library evolution flag and everything will work as we'd like

> That alone won't do it; we also need to separate the object generation and module generation into separate actions. 

Why's this? AFAICT you can make the current swift compile depend on the swiftinterfaces from its dependencies instead of the swiftmodules as this works as expected without splitting this action?

> so you could end up regressing performance depending on how much inlineable code you have

This is definitely interesting, I'd have to wire this up to try it

> Can you cite anything concrete for "greatly reduc[ing]" recompiles, and the conditions that holds under?

The problem we face today is we have a module tree with 500 or so modules, and if we change specific things that affect the swiftmodule output, which is almost anything (adding new private types, affecting internal / private function definitions) that files changes, therefore retriggering all the actions above it. In this case this cascades to the point where adding a private type in a low level module causes a full recompile of everything. Ideally Swift's incremental compilation wouldn't produce anything new in this case, and therefore even if you triggered the compile actions they would virtually be no-ops, but as the Swift team has said they're conservative so they recompile more often than not in our experience. If we depended on the swiftinterface we should be able to avoid this case entirely.

allevato:
> The problem we face today is we have a module tree with 500 or so modules, and if we change specific things that affect the swiftmodule output, which is almost anything (adding new private types, affecting internal / private function definitions) that files changes, therefore retriggering all the actions above it.

The private declarations are present in the swiftmodule for the purposes of debugging and diagnostics. Some things, like private stored properties, also have effects on ABI. If you get rid of all that information in your self-contained build from source, you're going to severely degrade your debugging experience, change the semantics of your code, and a number of other things are unlikely to not work the way you expect.

keith:
[](url)

keith:
I filed an issue tracking this conversation https://github.com/bazelbuild/rules_swift/issues/316

PR #329, opened 2019-10-10
feat: initial support for swift package manager

thesayyn:
I've put some effort to make rules swift to work with the swift package manager.

However, there are some edge cases that need to be cleared.

I would like to put some effort to make this usable but I need your opinions and reviews.

1- swift build command does not generate .a files maybe I don't need it at all. However,  I'm generating it with `ar` utility. (I don't know if this is the right way to do so)
2- I think we should pass some arguments to swift build to build dependencies with correct arch & cpu

Important: managed_directories support was introduced with Bazel 26.0 and we need to make use of it to speed up package resolve operation. However, it fails under CI. (It was working fine in local env)


thesayyn:
Ok now CI failing due to missing Process.run func in swift 4.2. Tomorrow I will put some conditions to support swift 4.2

ladislas:
any update? :)

thesayyn:
It has been a long time since I have been waiting for a review. I guess this feature is not their top priority right now.

eonil:
Any update?

keith:
I think supporting some SwiftPM setup is a good idea. I do think for the V1 we'd at least have to handle the other flag configuration you mentioned here. Likely [rules_foreign_cc](https://github.com/bazelbuild/rules_foreign_cc) can be a model for how to do that. If you want to keep going with this please submit a new PR that's updated!

adam-rocska:
@keith : Soooooo.... no SPM on the horizon yet?

PR #345, opened 2019-10-26
Stop declaring module swiftdeps output

kastiglione:
Module `.swiftdeps` files have thus far been declared outputs, but there are reasons they should not be. This change removes the module `.swiftdeps` from declared outputs. This matches the source file `.swiftdeps`, which are already not declared.

The module `.swiftdeps` don't contain paths, but do contain other build environment information: timestamps and a hash of command line arguments. This info varies across machines. Another potential issue is these  module`.swiftdeps` files are both an input and an output to `swiftc` actions, but rules_swift declares them only as outputs. No actions declare them as inputs.

Unless module `.swiftdeps` can be marked as `no-remote`, then making them undeclared seems to be the best approach.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -225,15 +225,6 @@ def declare_compile_outputs(
 
         output_map[src.path] = struct(**src_output_map)
 
-    # Output the module-wide `.swiftdeps` file, which is used for incremental
-    # builds.
-    swiftdeps = derived_files.swift_dependencies_file(
-        actions = actions,
-        target_name = target_name,
-    )
-    other_outputs.append(swiftdeps)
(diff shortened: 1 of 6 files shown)


allevato:
Thanks! I have an internal issue tracking this because those files set off some of our non-determinism detectors, but I haven't had a chance to circle back around to it yet.

kastiglione:
> non-determinism detectors

That's cool!

allevato:
> That's cool!

It's really just "two builds and a diff" 😛

allevato:
I had to make a small tweak and one other internal change, which apparently made our tooling not connect the commit that was pushed out back to this PR. This change was committed as 31bae7a5ccf36347abcdb089627aa0f89178a97e, so I'm closing this PR out.

PR #348, opened 2019-10-30
Add -Osize support

keith:
This adds a feature to use `-Osize` instead of `-O` which apparently
should rarely have a major impact on "normal" apps' performance
https://forums.swift.org/t/performance-impact-of-swift-compiler-optimization-level/30170/2


keith:
Thanks!

allevato:
I don't know why it took this long—I merged this internally 3 days ago...

PR #349, opened 2019-11-06
Use -filelist when linking swift_library

AlessandroPatti:
The list of object files can potentially become very long. Linking should be performed using `-filelist` and a parameter file.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -57,7 +57,11 @@ def _register_static_library_link_action(
     )
     args = actions.args()
     args.add_all(command_line)
-    args.add_all(objects)
+    args.add("-filelist")
+    objects_args = actions.args()
+    objects_args.set_param_file_format("multiline")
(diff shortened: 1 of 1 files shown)


allevato:
This change assumes that the static archiver always supports the `-filelist` flag, which isn't correct—`libtool` does on Darwin, but `ar` on Linux does not.

rules_swift delegates entirely to the C++ toolchain for linking, but right now there's a little bit of mixed logic because for static linking we're invoking the low-level direct action APIs. With a little bit of refactoring we should instead call `cc_common.create_linking_context_from_compilation_outputs` and grab the static library from the result of that call. Then, if it still doesn't do the right thing with regard to filelists for large inputs, the fix needs to be made in Bazel (either in CROSSTOOL, or in the implementation of the C++ rules).

AlessandroPatti:
@allevato Thanks for the feedback. I am closing this and will open another PR with more platform independent solution.

Issue #346, opened 2019-10-28
[Request] Name generated header after module name if provided


thii:
Currently the generated header of a `swift_library` is named after the target name. I'd like to request to change it to be named after the `module_name` attribute if provided, and `name` otherwise.

Our use case:
We build a mixed Obj-C and Swift static framework by building a `swift_library` target with the `-import-underlying-module` flag, and use an `objc_library` target as the wrapper target, with both target share the same module name. Since a target name has to be unique within a package, we name the `swift_library` target after the module name (for instance `MyLibrary`) in order to get the generated header named `MyLibrary-Swift.h` so that we can import it from the Obj-C part using `#import <MyLibrary/MyLibrary-Swift.h>`. Then for the wrapper target, we're having to name it something else (`MyLibrary_umbrella`) in order to make it work.

The build command for the wrapper target would be:
```
bazel build //path/to/MyLibrary:MyLibrary_umbrella
```

and the `deps` declaration for other targets that depend on this would be:

```
deps = [
    "//path/to/MyLibrary:MyLibrary_umbrella"
]
```

With this change, the above build command and `deps` declaration would be:

```
bazel build //path/to/MyLibrary
```

```
deps = [
    "//path/to/MyLibrary"
]
```

allevato:
This would be a breaking change, so it's not something we could do lightly—at a minimum, it could require some amount of migration for our internal users if they have any libraries with custom module names used from Objective-C (though we try to discourage using the `module_name` attribute entirely).

Doing this would also make the name of the generated header less predictable to the end user. Today, if you have a target `//foo/bar:baz`, then you know your Objective-C code can `#import "foo/bar/baz-Swift.h"`. Making that conditional on the module name makes that behavior harder to explain and document.

Ultimately I think the issue here isn't one that should be solved by renaming the generated header, but by having first-class mixed Swift/Objective-C support in the same module so that these kinds of hacks aren't necessary.

thii:
Thanks for the detailed explanation. That totally makes sense to me. Is first-class mixed Swift/Objective-C support in the same module in the roadmap 🙂?

For the time being, would it be possible to loose the spec a bit by allowing a custom name for the generated header via a `--features` flag?

mpalczew:
or how about just adding a new field to override the header name.

segiddins:
[...]

PR #350, opened 2019-11-07
Allow specifying a custom generated header name

segiddins:
This is my first time contributing to this repo, and I couldn't find any guidelines on where or how to add tests.

This is an attempt to implement https://github.com/bazelbuild/rules_swift/issues/346

### DIFF
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -243,12 +243,6 @@ directly to `copts`, only using this feature in the rare case that a library
 needs to propagate a symbol up to those that depend on it.
 """,
             ),
-            "generated_header_name": attr.string(
-                doc = """\
-The name of the generated Objective-C interface header.
-If not provided, it defaults to `${target_name}-Swift.h`.
-""",
-            ),
             "module_name": attr.string(
                 doc = """\
 The name of the Swift module being built.
@@ -1045,6 +1039,12 @@ explicitly called by code in the binary; for example, if you rely on runtime
 checks for protocol conformances added in extensions in the library but do not
 directly reference any other symbols in the object file that adds that
 conformance.
+""",
+            ),
+            "generated_header_name": attr.string(
+                doc = """\
+The name of the generated Objective-C interface header.
+If not provided, it defaults to `${target_name}-Swift.h`.
 """,
             ),
         },
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -243,6 +243,12 @@ directly to `copts`, only using this feature in the rare case that a library
 needs to propagate a symbol up to those that depend on it.
 """,
             ),
+            "generated_header_name": attr.string(
+                doc = """\
+The name of the generated Objective-C interface header.
+If not provided, defaults to `${target_name}-Swift.h`.
+""",
+            ),
             "module_name": attr.string(
                 doc = """\
 The name of the Swift module being built.
@@ -498,6 +504,7 @@ def _compile(
         copts = [],
         defines = [],
         deps = [],
+        generated_header_name = None,
         genfiles_dir = None):
     """Compiles a Swift module.
 
@@ -530,6 +537,8 @@ def _compile(
         deps: Dependencies of the target being compiled. These targets must
             propagate one of the following providers: `CcInfo`, `SwiftInfo`, or
             `apple_common.Objc`.
+        generated_header_name: The name of the generated Objective-C interface header.
+            If not provided, defaults to `${target_name}-Swift.h`.
         genfiles_dir: The Bazel `*-genfiles` directory root. If provided, its
             path is added to ClangImporter's header search paths for
             compatibility with Bazel's C++ and Objective-C rules which support
@@ -705,6 +714,7 @@ def _compile(
         generated_header = derived_files.objc_header(
             actions = actions,
             target_name = target_name,
+            header_name = generated_header_name,
         )
         args.add("-emit-objc-header-path")
         args.add(generated_header)
--- a/swift/internal/api.bzl
+++ b/swift/internal/api.bzl
@@ -246,7 +246,7 @@ needs to propagate a symbol up to those that depend on it.
             "generated_header_name": attr.string(
                 doc = """\
 The name of the generated Objective-C interface header.
-If not provided, defaults to `${target_name}-Swift.h`.
+If not provided, it defaults to `${target_name}-Swift.h`.
 """,
             ),
             "module_name": attr.string(
@@ -538,7 +538,7 @@ def _compile(
             propagate one of the following providers: `CcInfo`, `SwiftInfo`, or
             `apple_common.Objc`.
         generated_header_name: The name of the generated Objective-C interface header.
-            If not provided, defaults to `${target_name}-Swift.h`.
+            If not provided, it defaults to `${target_name}-Swift.h`.
         genfiles_dir: The Bazel `*-genfiles` directory root. If provided, its
             path is added to ClangImporter's header search paths for
             compatibility with Bazel's C++ and Objective-C rules which support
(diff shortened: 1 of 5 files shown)


segiddins:
(Working on getting added to our CLA google group)

segiddins:
@googlebot I signed it!

swift/internal/api.bzl
```diff
@@ -240,6 +240,12 @@ Each string is prepended with `-D` and added to the command line. Unlike
 it, so use this attribute with caution. It is preferred that you add defines
 directly to `copts`, only using this feature in the rare case that a library
 needs to propagate a symbol up to those that depend on it.
+""",
+            ),
+            "generated_header_name": attr.string(
+                doc = """\
+The name of the generated Objective-C interface header.
+If not provided, defaults to `${target_name}-Swift.h`.
```
thii: The markdown documentation needs to be updated too. I think it's not auto-generated for now.

swift/internal/api.bzl
```diff
@@ -240,6 +240,12 @@ Each string is prepended with `-D` and added to the command line. Unlike
 it, so use this attribute with caution. It is preferred that you add defines
 directly to `copts`, only using this feature in the rare case that a library
 needs to propagate a symbol up to those that depend on it.
+""",
+            ),
+            "generated_header_name": attr.string(
```
allevato: This probably belongs in `library_rule_attrs`, not in `compilation_attrs`. Putting it in `compilation_attrs` means it gets inherited by `swift_{binary,test}`, where it's not needed. The only odd case is `swift_module_alias`, but it looks like we don't need it there either—I just did a manual `swiftc` check and a module containing nothing but `@_exported import Foo` doesn't generate a useful header, so supporting it in that rule is misleading too.

So, let's move this to `library_rule_attrs` and remove it from the `compile` call in the non-library rules.

segiddins: Will move 👍

segiddins: @allevato moved it

swift/internal/swift_library.bzl
```diff
@@ -157,6 +157,7 @@ def _swift_library_impl(ctx):
         defines = ctx.attr.defines,
         deps = deps + private_deps,
         feature_configuration = feature_configuration,
+        generated_header_name = ctx.attr.generated_header_name,
```
allevato: One more thing that occurs to me before we merge this: we should probably have some basic validation of this attribute. Off the top of my head:

* Require it to end in `.h`
* Forbid it to contain `/`
* Anything else?

segiddins: I’d check whether the swift compiler has that validation, and if so probably not bother duplicating it here? If not, those are the only things that come to mind. 

I’m away from a computer the next couple of weeks so I won’t be able to add that validation until I get back.

allevato: There are still reasons to do more validation than the Swift compiler does, since we have to declare those files through Bazel's APIs. In the `/` case, for example, it's beneficial to not let the user declare files that are nested arbitrarily underneath the output directory (e.g., so they don't collide with child packages) and also prevents them from passing something with `../` segments.

Even if the compiler validates (I don't know that it does), waiting for that means waiting until the execution phase, which means you could get through 95% of your build before it dies, whereas validating here means we get effectively immediate failure during analysis, before any actions are executed.

segiddins: Added validations

segiddins:
@allevato I think the CLA bot is complaining because of your commit via suggested changes

allevato:
Yeah, you could work around that by squashing (I don't need to retain authorship of the suggestions) but since I have to pull the commit internally anyway to test it ultimately won't matter.

Sorry for the delay on this—I've been in an internal team event all week, so I'll likely take a look at this on Monday.

swift/internal/api.bzl
```diff
@@ -530,6 +531,8 @@ def _compile(
         deps: Dependencies of the target being compiled. These targets must
             propagate one of the following providers: `CcInfo`, `SwiftInfo`, or
             `apple_common.Objc`.
+        generated_header_name: The name of the generated Objective-C interface header.
```
allevato: Please wrap to 80 columns for consistency with the rest of the code base.

swift/internal/derived_files.bzl
```diff
@@ -122,17 +122,26 @@ def _modulewrap_object(actions, target_name):
...
 
     Returns:
         The declared `File`.
     """
-    return actions.declare_file("{}-Swift.h".format(target_name))
+    if header_name:
+      if not header_name.endswith(".h"):
+        fail("The generated objc header name for {} must end in '.h', given '{}'" % (target_name, header_name))
+      if header_name.find("/"):
+        fail("The generated objc header name for {} cannot contain a '/', given '{}'" % (target_name, header_name))
```
allevato: Please wrap these lines so that they fit within 80 columns for consistency with the rest of the code base.

thii:
Can we have this landed on the next release? :)

ob:
Any updates on this?

swift/internal/derived_files.bzl
```diff
@@ -122,17 +122,28 @@ def _modulewrap_object(actions, target_name):
...
         target_name: The name of the target being built.
+        header_name: The name of the header being declared.
 
     Returns:
         The declared `File`.
     """
-    return actions.declare_file("{}-Swift.h".format(target_name))
+    if header_name:
+        if not header_name.endswith(".h"):
+            fail(("The generated objc header name for {} must end in" +
```
thii:
```suggestion
            fail("The generated objc header name for '{target_name}' must end" +
                 " in '.h', given '{header_name}'".format(
                     target_name = target_name,
                     header_name = header_name,
                 ))
```

swift/internal/derived_files.bzl
```diff
@@ -122,17 +122,28 @@ def _modulewrap_object(actions, target_name):
...
+                  " '.h', given '{}'") % (target_name, header_name))
+        if header_name.find("/"):
```
thii:
```suggestion
        if "/" in header_name:
```

swift/internal/derived_files.bzl
```diff
@@ -122,17 +122,28 @@ def _modulewrap_object(actions, target_name):
...
+            fail(("The generated objc header name for {} cannot contain a " +
```
thii:
```suggestion
            fail("The generated objc header name for '{target_name}' cannot" +
                 " contain a '/', given '{header_name}'".format(
                     target_name = target_name,
                     header_name = header_name,
                 ))
```

allevato:
Sorry for the delay on this; in addition to being stretched a bit thin right now, I've been in the middle of refactoring these rules to support some future work that we need. I'll look at adding this change in as part of that so that you don't have to rebase it in on top of a completely changed implementation.

kastiglione:
@allevato I was wondering with Keith what was driving the refactoring. Are you able to comment on the referenced "future work"?

allevato:
> @allevato I was wondering with Keith what was driving the refactoring. Are you able to comment on the referenced "future work"?

The old design was based around the assumption that the only action we cared about was `SwiftCompile`. When I prototyped the support to integrate PCMs into the dependency graph, it became clear that there were places where I collected info from providers where I'd need different subsets of flags/input files for the `-emit-pcm` action vs. the compile action, and trying to parameterize those functions to say "do these things but not _these_ things" didn't scale at all.

By refactoring the actions to be more granular, adding PCM actions will be cleaner because all I'll have to do is add the PCM action name to the appropriate subset of action configurators, and then wire in the action itself.

It'll also pave the way for other improvements we've talked about but had to put off, like splitting out module compilation vs. object compilation.

kastiglione:
thanks!

PR #354, opened 2019-11-26
Uniquify prebuilt framework search paths

kastiglione:
Prevent duplicate framework search paths when there are multiple frameworks in a single directory.

See also https://github.com/bazelbuild/rules_swift/pull/334 which similarly enabled `uniquify` for import search paths.


allevato:
Thanks for calling this out! Since I'm in the middle of refactoring these actions, I'll drop this change in as part of that.

PR #363, opened 2019-12-19
Use depset(direct=...) not depset(items=...)

### LINKED ISSUE #360: Flag --incompatible_disable_depset_items will break rules_swift in Bazel 1.2.1
Incompatible flag --incompatible_disable_depset_items will break rules_swift once Bazel 1.2.1 is released.

Please see the following CI builds for more information:

* [:ubuntu: 18.04 (OpenJDK 11)](<a href="https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/342#0fbe3972-6997-420b-b2cd-dea5648f20cc" target="_blank">:ubuntu: 18.04 (OpenJDK 11)</a>)
* [:darwin: (OpenJDK 8)](<a href="https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/342#b1beb5e7-45d0-4b24-8d89-546b227b2ef1" target="_blank">:darwin: (OpenJDK 8)</a>)

Questions? Please file an issue in https://github.com/bazelbuild/continuous-integration

**Important**: Please do NOT modify the issue title since that might break our tools.

swiple-rules-gardener:
Use depset(direct=...) not depset(items=...)

Fixes #360

RELNOTES: None

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -176,7 +176,7 @@ def declare_compile_outputs(
             indexstore = None,
             other_outputs = [],
             output_groups = {
-                "compilation_outputs": depset(items = [out_obj]),
+                "compilation_outputs": depset(direct = [out_obj]),
             },
             output_objects = [out_obj],
         )
@@ -232,7 +232,7 @@ def declare_compile_outputs(
 
     args = ["-output-file-map", output_map_file]
     output_groups = {
-        "compilation_outputs": depset(items = output_objs),
+        "compilation_outputs": depset(direct = output_objs),
     }
 
     # Configure index-while-building if requested. IDEs and other indexing tools


kastiglione:
fwiw this change wasn't necessary, these output groups aren't used, here's a PR to remove them https://github.com/bazelbuild/rules_swift/pull/336

allevato:
Well, some of those output groups *should* be used, rather than removed; it's an oversight that I didn't pipe them all the way through to the outputs of the rules themselves, which is why I haven't addressed #336 yet. I had started refactoring some things to do that, but got pulled in another direction and haven't come back around to it yet, but I also didn't want to close out #336 until I had the alternative in place.

kastiglione:
How will the "compilation_outputs" output group be used? By other rules reading from the provider? If so, why an output group provider and not something more semantic?

I'm only aware of output groups being used as a "pull" mechanism for users to cause the build to run additional actions that generate addition output, but for compilation output, the actions and output already "push" these outputs.

kastiglione:
the summary of my comment is: those output groups seem like they would provide no value, even if they were to be wired up.

allevato:
The C++ rules populate the `compilation_outputs` output group with the object files, and that output group name is one that Bazel documents specifically [here](https://github.com/bazelbuild/bazel/blob/7202d109f611ac662860d2bb69633ce8e2e5183a/src/main/java/com/google/devtools/build/lib/analysis/OutputGroupInfo.java#L80-L84), and some IDE integrations use it as a way of collecting those intermediate outputs.

I was working on aligning the Swift rules to do the same thing for consistency, but I need to make other changes (like I should also be including the `.swiftmodule` files in the group, not just the object files).

PR #371, opened 2020-01-03
Start refactoring action registration to be more "modular".

swiple-rules-gardener:
Start refactoring action registration to be more "modular".

Inspired loosely by CROSSTOOL, this lets us split out various parts of toolchain and action configuration so that they can be shared among multiple actions. This will be critical when the build rules start supporting actions other than SwiftCompile, such as emitting precompiled modules for C/Obj-C dependencies.

There are two concepts: a tool config and an action config. A tool config describes an executable and some of its properties (its environment, execution requirements, etc.). This means these can now differ between different tools in the same toolchain (previously, all tools shared the same environment and exec reqs). An action config describes how args and inputs are decided for an action, based on which features are set in the feature configuration.

I've tried to keep this change from getting too large, so only the toolchain-specific compiler flags and the ModuleWrap and AutolinkExtract actions have been migrated over to the new logic. The Compile action has been mostly left alone, except where necessary to use the new APIs, and will be migrated separately.

RELNOTES: None.

### DIFF
--- a/swift/internal/actions.bzl
+++ b/swift/internal/actions.bzl
@@ -15,49 +15,135 @@
 """Functions for registering actions that invoke Swift tools."""
 
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
-load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//lib:partial.bzl", "partial")
 load("@bazel_skylib//lib:types.bzl", "types")
+load(":features.bzl", "are_all_features_enabled")
+load(":toolchain_config.bzl", "swift_toolchain_config")
 
-def _get_swift_driver_mode_args(driver_mode, swift_toolchain):
-    """Gets the arguments to pass to the worker to invoke the Swift driver.
+# The names of actions currently supported by the Swift build rules.
+swift_action_names = struct(
+    # Extracts a linker input file containing libraries to link from a compiled
+    # object file to provide autolink functionality based on `import` directives
+    # on ELF platforms.
+    AUTOLINK_EXTRACT = "SwiftAutolinkExtract",
+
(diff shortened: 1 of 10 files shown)


kastiglione:
Interesting!

What are the next steps?

Are these reviewed internally at google? I'm wondering how much this is set, how much this is already reviewed, etc?

allevato:
In general, these PRs are usually the state of a change before/during review, since our CI workflow pushes them out to GitHub so they can be tested with Bazel to make sure nothing breaks there because of subtle differences.

kastiglione:
Thanks. Would it be possible for review to happen on Github? Just thinking it would be great to have all that context/history available publicly.

kastiglione:
Why is modular quoted 😄?

allevato:
> Thanks. Would it be possible for review to happen on Github? Just thinking it would be great to have all that context/history available publicly.

Unfortunately no; our workflows for rules_{apple,swift} and apple_support are such that we use our internal tooling for reviews and push the results out to GitHub, and those workflows don't support easily copying the review discussion/history over (or preserving the diffs between commits that get pushed to GitHub, since it uses a force push). We treat these as PRs, rather than pushing directly to master, to get the CI checks but also to raise their visibility to external users more than they would be if we just updated the branch behind the scenes.

The main reasons for these workflows being the way they are are that many of these changes also include Google-only files that don't get pushed out, and since so many of our own apps are affected by changes to the build rules that doing the review on GitHub and then reconciling that internally would be a bit painful.

In an ideal world we'd be better able to bridge the two "worlds", but for now the team maintaining these rules doesn't have the resources to do so.

> Why is modular quoted 😄?

I don't know if I have a good answer for that one 😄

kastiglione:
thanks, I didn't think it would be possible, but worth asking. On the plus side, everything is thoroughly commented.

PR #374, opened 2020-01-09
Fix Swift build error when packages have spaces in path

thii:


allevato:
In general this is something that would probably be nice to support, but there are a couple caveats:

* This would make our label-to-module-name mangling slightly different from the one used by `objc_library`, when they're intended to be the same so that it's predictable regardless of what kind of target your dependency is. AFAIK, an `objc_library` in a directory with spaces in the path would still break.

* Last I recall, Bazel had various issues with spaces in paths where things could randomly break and it was generally not recommended to do that. Has that changed? I wasn't able to dig up any documentation in the affirmative or negative about it.

segiddins:
You're still not allowed a space in any path used in a runfile

thii:
Spaces are allowed in target names (https://docs.bazel.build/versions/2.0.0/build-ref.html#name), and in iOS in general, it's not uncommon to see spaces in paths, so it would be nice to have this supported (from both `swift_library` and `objc_library`). Though most people specify custom module names for their modules so this isn't something critical at the moment.

allevato:
But the title of this PR refers to files with "spaces in [the] path", though, and the path to the file determines the label's _package,_ not the target name. According to the [section about package naming on the same page you linked](https://docs.bazel.build/versions/2.0.0/build-ref.html#package-names-package-name), space is not one of the explicitly permitted characters in a package name.

While it's true that Apple paths often contain spaces, if you try to put those in your workspace and Bazel doesn't completely support them (as @segiddins mentions above, it appears to still be a known issue for runfiles at least: https://github.com/bazelbuild/bazel/issues/4327), then you could be opening yourself up to potential problems later. I'm not sure we should implement a "promise we can't keep yet," essentially.

Regarding the `:target` name part of the label, though spaces are permitted there, the target name is also entirely specified by the BUILD file author so it's under your control; it doesn't have to specifically match any existing path. There are also a *lot* of currently permitted characters in the `:target` name that we don't currently support, simply because there hasn't been a need yet, and it would be good to consider all of those holistically. Simply replacing them *all* with an underscore doesn't feel like the right solution, but I also don't know what else we can do that's easy for human authors to translate mentally.

thii:
Closing since this is a wrong fix — spaces aren't allowed in a package name.

PR #375, opened 2020-01-09
Add vfsoverlay feature for improved compilation performance

kastiglione:
For Swift modules with large transitive dependency sets, compilation performance can be sped up by replacing all `-I` flags – which is one per Bazel package, with a single `-I` flag and a `-vfsoverlay`.

In my measurements, I saw compilation time drop ~30% for a module that's near the root of our app's dependency tree. That module had >300 transitive `swift_library` dependencies.

Modules with few dependencies don't directly gain anything, but when building a Swift app, changing a module can mean recompiling all of its transitive reverse dependencies. This change improves the performance of these cascading rebuilds.

Another piece of anecdata: Incremental compilation after adding `private let i = 1` to a class, resulting in about a dozen modules being rebuilt:

Without vfsoverlay: ~215s
With vfsoverlay: ~166s

Almost a minute off.

For a reference point, this is using vfsoverlays to provide what header maps provide in Clang.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1978,6 +1978,10 @@ roots:
     name: \"{}\"
     contents:""".format(virtual_import_search_path)
 
+    if not swiftmodules:
+        contents += " []"
+        return contents
+
     for swiftmodule in swiftmodules:
         contents += """
       - type: file
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -39,6 +39,7 @@ load(
     "SWIFT_FEATURE_ENABLE_BATCH_MODE",
     "SWIFT_FEATURE_ENABLE_LIBRARY_EVOLUTION",
     "SWIFT_FEATURE_ENABLE_TESTING",
+    "SWIFT_FEATURE_ENABLE_VFSOVERLAYS",
     "SWIFT_FEATURE_FASTBUILD",
     "SWIFT_FEATURE_FULL_DEBUG_INFO",
     "SWIFT_FEATURE_IMPLICIT_MODULES",
@@ -64,6 +65,8 @@ load(
     "get_providers",
     "struct_fields",
 )
+
+VIRTUAL_IMPORT_SEARCH_PATH = "/build_bazel_rules_swift/swiftmodules"
 
 # The number of threads to use for WMO builds, using the same number of cores
 # that is on a Mac Pro for historical reasons.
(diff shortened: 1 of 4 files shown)


kastiglione:
Build failures are because I wrote this based on an earlier commit, so that we can use it as a patch. I will fix the build failures after getting feedback on the approach.

swift/internal/compiling.bzl
```diff
@@ -544,6 +576,75 @@ def write_objc_header_module_map(
         output = output,
     )
 
+def _write_swfitc_vfsoverlay(actions, module_name, swiftmodules):
```
rmaz: typo

swift/internal/compiling.bzl
```diff
@@ -20,15 +20,27 @@ load(":derived_files.bzl", "derived_files")
 load(":providers.bzl", "SwiftInfo")
 load(":utils.bzl", "collect_cc_libraries", "get_providers")
 
-def collect_transitive_compile_inputs(args, deps, direct_defines = []):
+VIRTUAL_IMPORT_SEARCH_PATH = "/bazel_build_rules_swift/swiftmodules"
```
thii:
```suggestion
VIRTUAL_IMPORT_SEARCH_PATH = "/build_bazel_rules_swift/swiftmodules"
```

brentleyjones:
I would love to have this feature. Thanks @kastiglione for upstreaming it!

keith:
@allevato is there anything we can do to this to get it merged?

kastiglione:
Let me know what queations I can answer is any.

kastiglione:
Rebased and included an update by @Keith.

swift/internal/compiling.bzl
```diff
@@ -1606,6 +1660,79 @@ def _write_objc_header_module_map(
...
+        swiftmodules: The `List` of swiftmodule `File`s to include in the
+             virtual file system.
+
+    Returns:
+        A string containing the YAML representation of the vfsoverlay.
+    """
+    # These explicit vfsoverlay settings ensure the VFS actually improves
+    # performance. Without these, the performance is worse.
+    contents = """\
+version: 0
```
segiddins, line 1711: would it maybe be preferable to use `struct.to_json` here instead of manually constructing a string ?

kastiglione: I thought the vfsoverlays supported only yaml? If they support json then yes that would be better 👍

rmaz: Xcode uses JSON for VFS overlays but calls them `.yaml` 🙈

thomasvl: Just because I've run into this before -- yaml is a subset of json - https://yaml.org/spec/1.2/spec.html#id2759572

rmaz: Oh nice, I did not know this

kastiglione: I'm surprised that `struct`s can have keys with hyphens.

kastiglione: done, it works

kastiglione:
another friendly ping @allevato

kastiglione:
I just remeasured. For our build, this saves roughly 37% on swift compilation.

allevato:
Gonna pull this in for some testing.

kastiglione:
I'll rebase and resolve.

swift/internal/compiling.bzl
```diff
@@ -1879,6 +1926,78 @@ def _write_objc_header_module_map(
...
+            "external-contents": swiftmodule.path,
+        }
+        for swiftmodule in swiftmodules
+    ]
+
+    # These explicit vfsoverlay settings ensure the VFS actually improves
+    # performance. Without these, the performance is worse.
+    overlay = {
+        "version": 0,
+        "case-sensitive": False,
```
kastiglione, line 1987: I don't why I did this as case _insensitive_. Maybe it should be `case-sensitive: True`.

PR #379, opened 2020-01-15
Remove the `swift.index_while_building` feature.

swiple-rules-gardener:
Remove the `swift.index_while_building` feature.

This was never used; for each of the use cases we originally planned, we ended up going in different directions, so now it just adds complexity to the build rules. For local builds, just pass `-index-store-path` as a manual `--swiftcopt` to get an indexstore that is shared across all actions. For remote indexing, use an aspect that invokes the compiler in parallel to perform indexing (hand-waving details).

RELNOTES: None.


allevato:
Looks like this is still in use by some external folks right now, so I'll leave it in and work around it in the migration.

keith:
Thanks for leaving it! For more context for others we use this since this is the only way (right now) to get indexes to be stored in the remote cache so other users can pull those down and use [index-import](https://github.com/lyft/index-import/) to post process them as needed

kastiglione:
Thanks Tony. For what it's worth, as long as `swiftc` produces indexes, we'll want to cache them from bazel. 

> For remote indexing, use an aspect that invokes the compiler in parallel to perform indexing (hand-waving details).

Can you elaborate on why you suggest this? It defeats the (optimization) purpose of in-process index generation, so outside of RBE, it would slow down builds.

Note that this isn't just remote builds: declaring indexstores in the build means a local disk cache can provide branch-snapshotted indexes.

kastiglione:
Tony what do you think about having some sort of community test suite?

PR #380, opened 2020-01-16
Migrate object file outputs and the output file map to the new action model, and push declaration of all compilation outputs into the same helper method.

swiple-rules-gardener:
Migrate object file outputs and the output file map to the new action model, and push declaration of all compilation outputs into the same helper method.

RELNOTES: None.


swift/internal/compiling.bzl
```diff
@@ -383,6 +395,13 @@ def compile_action_configs():
             actions = [swift_action_names.COMPILE],
             configurators = [_module_name_configurator],
         ),
+
+        # Configure index-while-building.
```
kastiglione: Thanks!

PR #381, opened 2020-01-17
Remove duplicate quote include search path

thii:
`bazel-genfiles` has been merged to `bazel-bin` since the flipping of
the `--incompatible_merge_genfiles_directory` flag. This removes one
when both are present.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -486,7 +486,9 @@ def _output_dirs_clang_search_paths_configurator(prerequisites, args):
             "-iquote{}".format(prerequisites.bin_dir.path),
         ])
 
-    if prerequisites.genfiles_dir:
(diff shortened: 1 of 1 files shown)


allevato:
This will be fixed more uniformly in my next change that completes the migration of the compilation actions to the new model.

A complete fix is a little more involved because the `"."` and `genfiles`/`bin` include paths can also get propagated by `cc_*` targets. If we add them manually, and let the `cc_*` targets add their include paths, we still get dupes. Instead of adding them directly by calling `args.add`, it turns out that it's better to just add them to the same `depset` that has all the other `-iquote`s collected, and then Bazel handles the deduping for us regardless of whether they came from our rules or transitively from some target.

thii:
Closing as this was resolved in 02c0ceb0fd1e966710c3f803bfae5136909de197.

Issue #338, opened 2019-10-19
linkopts are backwards


mpalczew:
adding link opts to a `swift_library` doesn't work as expected. 

*Expected*
```
linkopts = ["-weak_framework Combine"]
```
This fails because it treats this as one flag.  There is no option '-weak_framework Combine'
```
linkopts = ["-weak_framework", "Combine"]
```
This also fails because the options are passed in as 'Combine' '-weak_framework' 
*Workaround*
```
linkopts = ["Combine", "-weak_framework"]
```

allevato:
More information is needed. What rule is doing the linking in this case, a `swift_binary` or something from rules_apple like `ios_application`?

If it's the latter, we have to use `ObjcProvider` to propagate linker flags, which is known to be buggy because it improperly deduplicates flags under certain circumstances, and I wonder if that's what's happening here. Are you using `-weak_framework` anywhere else in your build graph?

(Your expected case is wrong though—a single string should be expected to be treated as a single flag without additional processing, otherwise flags with spaces in them wouldn't be possible.)

mpalczew:
Fair enough on the expected case. I'm using the link opts in my swift_library. My ios_ui_test won't link without this change.

allevato:
Can you provide some more information? Do you have any other `swift_libraries` or `objc_libraries` in the dependency graph that also provide custom `linkopts` (especially `-weak_framework`)? Can you provide a more complete snapshot of your targets, or a small/redacted repro case?

mpalczew:
I can probably throw something together, will take a little time. 
I'm basically doing something like. 

swift_lib  A
swift_lib A_test_lib   dep -> A
ios_ui_test A_test dep -> A_test_lib

brentleyjones:
I just ran into this as well. We needed to reverse the ordering to get weak linking working.

keith:
This is the entire config you need to repro this issue:

```
load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test")
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")

swift_library(
    name = "foo",
    srcs = ["foo.swift"],
    linkopts = [
        "-weak_framework",
        "CoreNFC",
    ],
)

ios_unit_test(
    name = "test",
    deps = ["foo"],
    minimum_os_version = "11.0",
)
```

keith:
Here's a fix https://github.com/bazelbuild/rules_swift/pull/383

[...]

PR #383, opened 2020-01-27
Fix swift_library linkopts ordering for ObjC

keith:
Previously this list would be interpreted backwards

Fixes https://github.com/bazelbuild/rules_swift/issues/338

This is also how this is done in rules_apple https://github.com/bazelbuild/rules_apple/blob/53282252b393e451634d4f410d555df3f295c00e/apple/internal/swift_support.bzl#L84

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1391,7 +1391,7 @@ def new_objc_provider(
     if objc_header:
         objc_provider_args["header"] = depset(direct = [objc_header])
     if linkopts:
-        objc_provider_args["linkopt"] = depset(direct = linkopts)
+        objc_provider_args["linkopt"] = depset(direct = linkopts, order = "topological")
 
     force_loaded_libraries = [
         archive


thomasvl:
https://docs.bazel.build/versions/master/skylark/lib/depset.html

< "topological" (formerly "link"): A topological ordering from the root down to the leaves. There is no left-to-right guarantee.

Doesn't sound like it really ensures it.

keith:
I tried preorder and postorder, both caused a bazel crash, so since I also saw this in rules_apple I figured this must be the right one. I couldn't find anything in the blame on that (didn't go super deep) any idea why it's like that there?

swift/internal/compiling.bzl
```diff
@@ -1391,7 +1391,7 @@ def new_objc_provider(
     if objc_header:
         objc_provider_args["header"] = depset(direct = [objc_header])
     if linkopts:
-        objc_provider_args["linkopt"] = depset(direct = linkopts)
+        objc_provider_args["linkopt"] = depset(direct = linkopts, order = "topological")
```
kastiglione: I haven't had to use the order parameter, but I have some questions. A quick look at the docs make me wonder if this should be `preorder`? The docs say `preorder` was formerly called "native_link", and `topological` was called "link", so both were used for linking. I don't know how "native_link" and "link" differed.

The other thing is that `topological` says there's no left-to-right guarantee, but `preorder` does, and it seems like that would be desirable?

https://docs.bazel.build/versions/master/skylark/lib/depset.html

kastiglione: Oh I see the above comments now

keith:
Ah it seems like the exception is actually intentional, validating that this ordering is topological:

```
Caused by: java.lang.IllegalArgumentException: Order mismatch: preorder != topological
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:440)
        at com.google.devtools.build.lib.collect.nestedset.NestedSetBuilder.addTransitive(NestedSetBuilder.java:142)
        at com.google.devtools.build.lib.rules.objc.ObjcProvider$Builder.uncheckedAddTransitive(ObjcProvider.java:1034)
        at com.google.devtools.build.lib.rules.objc.ObjcProvider$Builder.addElementsFromSkylark(ObjcProvider.java:1153)
        at com.google.devtools.build.lib.rules.objc.AppleSkylarkCommon.newObjcProvider(AppleSkylarkCommon.java:187)
```

allevato:
Huh. I'm surprised we weren't already using `topological` here. Looking at the code in Bazel, it appears that when you merge a `topological` and a `stable` depset, one of them gets flipped. Even their `direct` members, which I thought stayed ordered the same relative to each other and it was only the _traversal_ order that changed.

This situation still sucks because fundamentally, `ObjcProvider` is broken and there's nothing we can do about it without fixing the implementation in Bazel (which isn't worth it, since there's work to migrate to the C++ providers instead). Linkopts shouldn't be depsets at all, because it means you'll still get deduplicated flags in some situations. If you have two `swift_libraries` and one has `["-weak_framework", "Foo"]` and the other has `["-weak_framework", "Bar"]`, you'll end up with `["-weak_framework", "Foo", "Bar"]`, or some other possible ordering.

So this PR buys users a little bit of false hope/security, but I guess it also lets a little bit more stuff work, so 🤷‍♂  Let's go ahead and pull it in.

keith:
Sounds good, here's a better workaround for others, but I guess this is still nice to fix some cases https://github.com/bazelbuild/rules_swift/issues/338#issuecomment-578951441

PR #386, opened 2020-02-05
Make rules_swift compatible with --incompatible_load_proto_rules_from_bzl

Yannic:


allevato:
Thanks for this PR. This will need a little more work first (from my end) before we can merge it because we'll have to update our file transformations to load the definitions from the correct paths within Google's monorepo as well.

keith:
This has been fixed on master

PR #388, opened 2020-02-11
Allow cacheable_swiftmodule feature independent of compilation mode

### LINKED ISSUE #398: Absolute paths embedded in .swiftmodule
When performing distributed builds (in which case an action's execution path is essentially random to avoid collision) or when sharing a cache between different users `.swiftmodule` files contain absolute paths. This leads to cascade rebuilds of all reverse dependencies even if no changes ave
I have tried performing the build with `--features=debug_prefix_map_pwd_is_dot` and `--features=swift.cacheable_swiftmodules`, but it does not seem to affect this behavior.
Looking at the `swiftmodule` in text mode, the values of `--working-directory` and`--fdebug-prefix-map` also embedded with absolute paths in their values. Also, it contains absolute action output paths of all swift dependencies.

This is likely an issue with the swift compiler itself, and out of scope of this repository. In that case I would appreciate any input on what could be done.

kastiglione:
We noticed that `swiftc` command lines were containing offsetting flags when enabling the `cacheable_swiftmodules` feature.

```
-Xfrontend -no-serialize-debugging-options -Xfrontend -serialize-debugging-option
```

The cause is the `not_features` used in this case: https://github.com/bazelbuild/rules_swift/blob/657eda57a097980848dbb3c880b94faeddaa3cd1/swift/internal/compiling.bzl#L211-L223

Which says to add `-serialize-debugging-option` using this logic:

```py
if not (opt and cacheable_swiftmodules):
   # use -serialize-debugging-option
```

but the logic should be:

```py
if not (opt or cacheable_swiftmodules):
   # use -serialize-debugging-option
```

This fix changes `not_features` to match `features` in structure and meaning. Now, the `or` can be expressed using a list of list of feature names:

```
not_features = [
    [SWIFT_FEATURE_OPT],
    [SWIFT_FEATURE_CACHEABLE_SWIFTMODULES],
]
```

Fixes #398 

The change that introduced the regression was https://github.com/bazelbuild/rules_swift/pull/376

### DIFF
--- a/swift/internal/actions.bzl
+++ b/swift/internal/actions.bzl
@@ -108,10 +108,15 @@ def _apply_action_configs(
         # features, check those as well and possibly decide to not apply the
         # configurators based on those.
         if should_apply_configurators and action_config.not_features:
-            should_apply_configurators = not are_all_features_enabled(
-                feature_configuration = feature_configuration,
-                feature_names = action_config.not_features,
-            )
+            # The configurators will not be applied if any of the
+            # `not_features` exclusion lists are entirely enabled.
+            for feature_names in action_config.not_features:
+                if are_all_features_enabled(
+                    feature_configuration = feature_configuration,
+                    feature_names = feature_names,
+                ):
+                    should_apply_configurators = False
+                    break
 
         # If one of the feature lists is completely satisfied, invoke the
         # configurators.
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -217,8 +217,8 @@ def compile_action_configs():
                 ),
             ],
             not_features = [
-                SWIFT_FEATURE_OPT,
-                SWIFT_FEATURE_CACHEABLE_SWIFTMODULES,
+                [SWIFT_FEATURE_OPT],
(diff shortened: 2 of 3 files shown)


kastiglione:
The buildifier failures are an existing issue:


<pre><code><a href="https://github.com/kastiglione/rules_swift/blob/7176c9d7ae5d03e6c411b0a3d8e6ce721d268bc8/swift/internal/swift_grpc_library.bzl#L205">swift/internal/swift_grpc_library.bzl:205</a>: <a href="https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#list-append">list-append</a>: Prefer using ".append()" to adding a single element list.</pre></code>

kastiglione:
The buildifier issue is fixed in https://github.com/bazelbuild/rules_swift/pull/387

kastiglione:
@allevato any thoughts on this?

allevato:
I was hoping we wouldn't need the additional complexity of multiple not-features lists, but it looks unavoidable here (and in another case I'm going to hit in a change I'm working on). Thanks for the fix!

kastiglione:
Thanks, yeah it would be better if it didn't have the additional complexity, but I didn't see how else to solve this case.

PR #396, opened 2020-02-21
Pass dependencies' module maps in `SwiftInfo` in a structured manner.

swiple-rules-gardener:
Pass dependencies' module maps in `SwiftInfo` in a structured manner.

Currently, module maps are retrieved in two ways—the `SwiftInfo.transitive_modulemaps` field and the `apple_common.Objc.module_map` field—and these are uniformly passed to ClangImporter using `-fmodule-map-file`. To support precompiled modules, we'll need to pass both the module name and the precompiled module file in the same unit, and in some cases we may still need to pass the module map and compilation context for a module instead of a PCM (if some `swift_library` opts out of consuming PCMs for its dependencies).

Thus, we add a new field—`SwiftInfo.transitive_modules`—that is a `depset` of structs. Each struct represents a single module (Swift and/or Clang), and we put the modules for both `cc_library` and `objc_library` dependencies there so that the compilation logic handles them uniformly.

This change lays the groundwork for Swift module artifacts to be migrated to this new structured format as well, but this will be done in a separate change.

We still have to maintain the `SwiftInfo.transitive_modulemaps` field for compatibility with Tulsi for the time being, but that field is now deprecated.

RELNOTES: None.


swift/internal/non_swift_target_aspect.bzl
```diff
@@ -183,6 +267,14 @@ def _non_swift_target_aspect_impl(target, aspect_ctx):
             swift_toolchain = swift_toolchain,
             target = target,
         )
+    elif apple_common.Objc in target:
+        return _handle_objc_target(
```
gyfelton, line 271: @allevato Can we consider adding a feature to bypass this processing here in the future? The reason is that we end up having to skip the generated module map from here in this PR for rules_ios: https://github.com/bazel-ios/rules_ios/pull/45 🙏

PR #401, opened 2020-03-11
Add swiftmodule and swiftdoc to output_groups

brentleyjones:


brentleyjones:
Re-submiited as https://github.com/bazelbuild/rules_swift/pull/421 from a different fork.

PR #411, opened 2020-04-17
Replace + (plus) in the generated module name with _ (underscore)

yusuke024:
We had a problem trying to import `objc_library` dependency with a `+` in the package name (the package was category-only source) since `import` statement in Swift doesn't support special characters other than `_`. I can confirm that the generated module name contains `+` by looking at the generated module map file.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -777,7 +777,7 @@ def derive_module_name(*args):
 
     package_part = (package.lstrip("//").replace("/", "_").replace("-", "_")
         .replace(".", "_"))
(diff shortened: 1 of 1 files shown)


yusuke024:
I just realized that this doesn't solve the problem I'm having.
Is the modulemap files of the targets listed in `swift_library.deps` generated by `objc_library`?
If that's the case do you have any suggestions?

yusuke024:
Just realized that this doesn't solve the problem I'm having since the modulemap files of the targets in deps are generated by `objc_library` and is outside the responsibility of the code in this repo.

PR #412, opened 2020-04-17
Fix incorrect doc for `no_generated_module_map` feature

gyfelton:
According to the actual implementation [line 1091 of compiling.bzl](https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/compiling.bzl#L1091), deciding whether to generate module map depends on whether we have generated headers to start with. In practice we have this feature enabled without the need to pass `no_generated_header`, therefore the documentation is stated the opposite.

**I know I suppose to create an issue but this is trivial (?) that I decide to just open up a PR instead. Let me know what should I do since I am newbie to contributing to open source :)**


gyfelton:
@googlebot I signed it!

PR #413, opened 2020-04-20
Add feature to use absolute source file paths when compiling

brentleyjones:
This feature is useful if you need `#file`/`#filePath` to return an absolute path, which is required by Xcode for some functionality.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -29,6 +29,7 @@ load(":debugging.bzl", "ensure_swiftmodule_is_embedded")
 load(":derived_files.bzl", "derived_files")
 load(
     ":feature_names.bzl",
+    "SWIFT_FEATURE_ABSOLUTE_SOURCE_FILES",
     "SWIFT_FEATURE_CACHEABLE_SWIFTMODULES",
     "SWIFT_FEATURE_COMPILE_STATS",
     "SWIFT_FEATURE_COVERAGE",
@@ -529,7 +530,24 @@ def compile_action_configs():
                 swift_action_names.COMPILE,
                 swift_action_names.PRECOMPILE_C_MODULE,
             ],
+            configurators = [
+                _absolute_source_files_configurator,
+                swift_toolchain_config.add_arg(
+                    "-Xwrapped-swift=-use-absolute-paths",
+                ),
+            ],
+            features = [SWIFT_FEATURE_ABSOLUTE_SOURCE_FILES],
+        ),
+    )
+
+    action_configs.append(
+        swift_toolchain_config.action_config(
+            actions = [
+                swift_action_names.COMPILE,
+                swift_action_names.PRECOMPILE_C_MODULE,
+            ],
             configurators = [_source_files_configurator],
+            not_features = [SWIFT_FEATURE_ABSOLUTE_SOURCE_FILES],
         ),
     )
 
@@ -863,6 +881,20 @@ def _module_name_configurator(prerequisites, args):
 def _stats_output_dir_configurator(prerequisites, args):
     """Adds the compile stats output directory path to the command line."""
     args.add("-stats-output-dir", prerequisites.stats_directory.path)
+
+def _make_path_absolute(source_file):
+    """Adds a placeholder prefix for the PWD that will be replaced by swift_runner."""
+    return "__BAZEL_PWD__/{}".format(source_file.path)
+
+def _absolute_source_files_configurator(prerequisites, args):
+    """Adds source files, with absolute paths, to the command line and required inputs."""
+    args.add_all(
+        prerequisites.source_files,
+        map_each = _make_path_absolute,
+    )
+    return swift_toolchain_config.config_result(
+        inputs = prerequisites.source_files,
+    )
 
 def _source_files_configurator(prerequisites, args):
     """Adds source files to the command line and required inputs."""
@@ -1462,9 +1494,14 @@ def _declare_compile_outputs(
     else:
         # Otherwise, we need to create an output map that lists the individual
         # object files so that we can pass them all to the archive action.
+        write_absolute_source_file_paths = is_feature_enabled(
+            feature_configuration = feature_configuration,
+            feature_name = SWIFT_FEATURE_ABSOLUTE_SOURCE_FILES,
+        )
         output_info = _declare_multiple_outputs_and_write_output_file_map(
             actions = actions,
             emits_partial_modules = output_nature.emits_partial_modules,
+            write_absolute_source_file_paths = write_absolute_source_file_paths,
             srcs = srcs,
             target_name = target_name,
         )
@@ -1506,6 +1543,7 @@ def _declare_compile_outputs(
 def _declare_multiple_outputs_and_write_output_file_map(
         actions,
         emits_partial_modules,
+        write_absolute_source_file_paths,
         srcs,
         target_name):
     """Declares low-level outputs and writes the output map for a compilation.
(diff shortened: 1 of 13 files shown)


swift/internal/feature_names.bzl
```diff
@@ -61,6 +61,11 @@ SWIFT_FEATURE_COMPILE_STATS = "swift.compile_stats"
 # builds.
 SWIFT_FEATURE_DEBUG_PREFIX_MAP = "swift.debug_prefix_map"
 
+# If enabled, Swift compilation actions will pass absolute paths, resolved with
+# realpath, for source files. This is allow `#file` to match Xcode's behavior.
```
keith, line 65: Note the `#file` behavior probably isn't a great callout here since the default is changing in Swift to be relative in Xcode as well

brentleyjones: `#filePath` isn't changing to be relative though, is it?

brentleyjones: I mentioned both in the commit message, I should mention both here as well.

keith: Not AFAIK

brentleyjones: I adjusted the comment to make it clear I'm talking about `#filePath`.

swift/internal/compiling.bzl
```diff
@@ -1239,12 +1271,29 @@ def _declare_multiple_outputs_and_write_output_file_map(
...
-    )
+    if write_absolute_source_file_paths:
+        raw_output_map_file = actions.declare_file("{}.output_file_map.raw.json".format(target_name))
+        actions.write(
+            content = struct(**output_map).to_json(),
+            output = raw_output_map_file,
+        )
+        actions.run_shell(
+            arguments = [raw_output_map_file.path, output_map_file.path],
+            command = 'perl -MCwd -pe \'s|__BAZEL_PWD__/([^"]*)|Cwd::realpath($1)|ge\' < $1 > $2',
```
kastiglione, line 1287: I wonder if this should happen in the worker which already parses the output file map.

kastiglione:
Thanks for opening this PR!

swift/internal/compiling.bzl
brentleyjones: I didn't even notice `output_file_map.cc`. I agree, and I'll make that change.

brentleyjones: I've now updated this to be processed in the worker. The code could probably use some cleanup (I don't normally write C++), but it works.

Changes:
- Additional work is only done if `-Xwrapped-swift=-use-absolute-paths` is passed to the worker
- Only `__BAZEL_PWD__/` at the start of a path is processed
- Paths in the output file map are processed by the worker instead of in starlark

brentleyjones:
@googlebot I fixed it.

brentleyjones:
@googlebot I consent

brentleyjones:
@googlebot I consent.

brentleyjones:
@googlebot I fixed it.

tools/common/file_system.cc
```diff
@@ -29,6 +29,18 @@
 
 #include "tools/common/path_utils.h"
 
+namespace {
+
+static std::string RealPath(const std::string &src) {
+  // Passing path,null causes realpath to allocate the buffer of the correct size.
+  char *buffer = realpath(src.c_str(), nullptr);
+  std::string realpath(buffer);
```
segiddins, line 37: too bad we're not using c++17, `std::filesystem::canonical` would handle this

tools/common/string_utils.cc
```diff
@@ -40,3 +40,41 @@ bool MakeSubstitutions(std::string *arg,
 
   return changed;
 }
+
+std::string Unescape(const std::string &arg) {
```
segiddins, line 44: maybe a quick doc comment about what exactly this unescapes?

tools/worker/compile_without_worker.cc
```diff
@@ -14,12 +14,117 @@
...
+
+  auto consumer = [&](const std::string &arg) {
+    if (arg == "-output-file-map") {
+      // Peel off the `-output-file-map` argument, so we can rewrite it if
+      // necessary later.
+    } else if (prev_arg == "-output-file-map") {
+      output_file_map_path = arg;
+    } else if (arg == "-Xwrapped-swift=-use-absolute-paths") {
+      use_absolute_paths = true;
+      full_args.push_back(arg);
```
segiddins, line 63: wouldn't we not want to push this onto the list of args, since swift won't understand it?

brentleyjones: We need it for logic in `SwiftRunner::ProcessPossibleResponseFile`. We are only looking for it here so we can determine if we need to process the output_file_map.

tools/common/string_utils.cc
brentleyjones: I do in the header, similar to `MakeSubstitutions`. Is there something more you think I should do?

brentleyjones: This is also just a move from `swift_runner.cc`.

tools/worker/swift_runner.cc
```diff
@@ -162,6 +124,12 @@ bool SwiftRunner::ProcessPossibleResponseFile(
         });
   }
 
+  // If we have to transform source file paths, do another pass on the arguments
+  if (use_absolute_paths_) {
+    std::transform(
```
brentleyjones, line 129: This fails to build on the linux bazel-ci.

@keith C++ and linux are areas i'm not well versed in. What would you suggest as an alternative here?

brentleyjones:
It doesn't look like we have tests covering the runner. While that is an area we will want to cover eventually, is it fine that we don't have runner tests for this? I'll look at adding tests for stuff leading up to the runner.

tools/worker/swift_runner.cc
kastiglione: > error: no member named 'transform' in namespace 'std'

huh, I wonder why. You could instead do a `for` loop here and manually implement it.

kastiglione: oh, add `#include <algorithm>`?

brentleyjones: Thanks @kastiglione!

keith:
Are there no issues with cache invalidation for remote builds with this? Do the absolute paths never make it into swiftmodules at least?

brentleyjones:
Since it's a feature that should only be enabled on specific targets (we only enable them on test targets, and only in our Xcode builds), the cache invalidation is minimal. The feature argument, the new argument to the worker, and the resulting swiftmodules/.os/.as are all changed and result in different caches. I do say in the feature comment: "Remote or sandboxed builds probably shouldn't enable this feature."

keith:
Hrm yea so I was thinking you'd use this on _all_ local build targets since there's also an issue that with the new build system jumping to diagnostics doesn't work unless the paths are absolute. But I guess that has to be solved elsewhere since you don't actually want to bake those paths in to the binaries, you just want to process the output.

Note you could probably do the same output processing for the XCTest case, although it's definitely nasty.

The annoyance about this for our use case is we have many helper modules that can contain XCTest failure methods, so we'd end up having to exclude a huge number of targets from being cacheable. But just a thought.

I filed FB8451256 about making Xcode support relative paths in test failure output because I think that's a pretty fair use case.

brentleyjones:
> Note you could probably do the same output processing for the XCTest case, although it's definitely nasty.

How? It's not something that is reported via stdout/stderr. Maybe something deep inside of Xcode?

keith:
Yea turns out swizzling XCTest to make the paths absolute seems to work for this case. Still productionizing it for us

brentleyjones:
@keith Now that Xcode 12 has shipped, care to share details on your solution? Maybe we can add it to https://github.com/ios-bazel-users/ios-bazel-users? I would rather go that route than landing this feature.

brentleyjones:
Closing this, assuming your way will work. I'll reopen if we find another need for it.

PR #420, opened 2020-04-22
Remove propagating cc libs with objc provider

keith:
When using a rule tree like this:

```
ios_static_framework
swift_library
objc_library
cc_library
```

Including these libraries in the objc provider leads them to being
included twice in the produced static binary. It appears the comment
here was outdated and that apple_binary does collect libraries directly
from the CcInfo.


keith:
This breaks other cases where you have more objc libs that depend on the swift libs, now tracking this issue in https://github.com/bazelbuild/bazel/issues/11223

PR #421, opened 2020-04-23
Add swiftmodule and swiftdoc to output_groups

brentleyjones:
Some Xcode integrations need these, and by having them as output groups (along with the use of aspects) remote build without the bytes can be enabled properly.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1890,9 +1890,19 @@ def output_groups_from_compilation_outputs(compilation_outputs):
             compilation_outputs.stats_directory,
         ])
 
+    if compilation_outputs.swiftdoc:
+        output_groups["swiftdoc"] = depset([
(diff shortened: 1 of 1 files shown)


keith:
can you rebase this one to get green CI here

brentleyjones:
All green.

kastiglione:
Related to this, I think starting with Xcode 11.5 (swift 5.2.4), Xcode produces `.swiftsourceinfo`. For context, see: https://forums.swift.org/t/proposal-emitting-source-information-file-during-compilation/28794

I'm not sure if (or when) Xcode-bazel integration would need the `.swiftsourceinfo` files, but this PR reminded me of seeing them the other day. Our swiftc stub needed to handle them to support the new build system.

keith:
We should add that one if it becomes important in the future :+1:

PR #425, opened 2020-04-28
Add skylib toolchain to WORKSPACE

keith:
This is required to run some of the tests


keith:
https://github.com/bazelbuild/rules_swift/pull/426

PR #427, opened 2020-04-28
Allow cacheable swiftmodules option with opt

keith:
We've seen cases with envoy-mobile where, when producing a static framework artifact, neither serialize debugging options flag is passed, but `-fdebug-prefix-map=/private/var/tmp/...` is embedded in the swiftmodule file from the C++ libraries it depends on. This appears to be because the default in the Swift compiler prefers embedding them in some cases.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -235,7 +235,6 @@ def compile_action_configs():
                 ),
             ],
             features = [SWIFT_FEATURE_CACHEABLE_SWIFTMODULES],
-            not_features = [SWIFT_FEATURE_OPT],
         ),
         swift_toolchain_config.action_config(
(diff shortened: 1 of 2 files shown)


kastiglione:
The context of why this was disabled for `opt` originally:

At the time this was written, the existing `-serialize-debugging-options` was inside of a `if dbg or fastbuild:` block, and so that's where I added `-no-serialize-debugging-options`. However in hindsight the `no-serialize` addition should not have been conditional on compilation mode. Instead it should have been placed outside of the `if dbg or fastbuild`.

See here:
https://github.com/bazelbuild/rules_swift/pull/296/files#diff-9cbe8d63c6561748aa94fe743f6178b2R266-R281

PR #428, opened 2020-04-29
Add feature to disable default -D flags

keith:
In some cases the default `-D` flags are not desirable since users
prefer to control them themselves.


keith:
If preferred I could also make this a default-on feature like `enable_d_flags` and then users could disable that

segiddins (approved):
👍  if this is still something you have a usecase for

brentleyjones (approved):
I think we should have this. I'm torn on if it should be default or not.

keith:
if anyone want this happy to accept a pr

PR #429, opened 2020-04-30
Allow swift_library to have empty srcs

keith:
This mirrors the behavior of `cc_library` to allow `swift_library`
targets to have empty `srcs`. This is valuable when you want to
optionally have source files based on some `select()` statement.

### DIFF
--- a/swift/internal/swift_library.bzl
+++ b/swift/internal/swift_library.bzl
@@ -20,7 +20,6 @@ load(
     "find_swift_version_copt_value",
     "new_objc_provider",
     "output_groups_from_compilation_outputs",
-    "swift_library_output_map",
 )
 load(
     ":feature_names.bzl",
@@ -147,6 +146,16 @@ def _swift_library_impl(ctx):
     else:
         deps = ctx.attr.deps
         private_deps = []
+
+    if not srcs:
+        return [
+            swift_common.create_swift_info(
+                # Note that private_deps are explicitly omitted here; they should
+                # not propagate.
+                swift_infos = get_providers(deps, SwiftInfo),
+                swift_version = find_swift_version_copt_value(copts),
+            ),
+        ]
 
     compilation_outputs = swift_common.compile(
         actions = ctx.actions,
@@ -290,7 +299,6 @@ dependent for linking, but artifacts/flags required for compilation (such as
     doc = """\
 Compiles and links Swift code into a static library and Swift module.
 """,
-    outputs = swift_library_output_map,
     implementation = _swift_library_impl,
     fragments = ["cpp"],
 )
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1844,27 +1844,6 @@ def output_groups_from_compilation_outputs(compilation_outputs):
 
     return output_groups
 
-def swift_library_output_map(name, alwayslink):
-    """Returns the dictionary of implicit outputs for a `swift_library`.
-
-    This function is used to specify the `outputs` of the `swift_library` rule;
-    as such, its arguments must be named exactly the same as the attributes to
-    which they refer.
-
-    Args:
-        name: The name of the target being built.
-        alwayslink: Indicates whether the object files in the library should
-            always be always be linked into any binaries that depend on it, even
-            if some contain no symbols referenced by the binary.
-
-    Returns:
-        The implicit outputs dictionary for a `swift_library`.
-    """
-    extension = "lo" if alwayslink else "a"
-    return {
-        "archive": "lib{}.{}".format(name, extension),
-    }
-
 def _write_objc_header_module_map(
         actions,
         module_name,


swift/internal/swift_library.bzl
```diff
@@ -148,6 +147,16 @@ def _swift_library_impl(ctx):
...
 
+    if not srcs:
+        return [
+            swift_common.create_swift_info(
+                # Note that private_deps are explicitly omitted here; they should
+                # not propagate.
+                swift_infos = get_providers(deps, SwiftInfo),
+                swift_version = find_swift_version_copt_value(copts),
+            ),
+        ]
```
keith, line 158: We might need to propagate an objc provider here too

thomasvl:
> This mirrors the behavior of `cc_library` to allow `swift_library`
> targets to have empty `srcs`. This is valuable when you want to
> optionally have source files based on some `select()` statement.

From a build graph/performance pov, wouldn't it be better to `select()` the library in/out of the `deps` of the thing depending on it?  Otherwise, are you removing all these sources, but still compiling/linking all of that removed code's `deps`, which would likely be wasted time/bloat in the final application?

keith:
> the library in/out of the deps of the thing depending on it?

It might be, but the major downside here is that in complex dependency graphs you'd have to know to do this for every single dependent.

> but still compiling/linking all of that removed code's deps, which would likely be wasted time/bloat in the final application?

How we handle deps here is a good question, it seems like `cc_library` chooses to still propagate those, but I can see it both way, I just wanted to mirror it's behavior. I honestly think in our macro when we `select`ed out the sources we would likely make the `deps` empty in that case as well to solve what you said.

keith:
Sounds like this isn't something folks want.

kastiglione:
> when we `select`ed out the sources we would likely make the `deps` empty in that case as well

this sounds like something that could be in `swift_library`. Let's say a `swift_library` has no sources, is there any reason at that point to build deps? Shouldn't that be a cut off valve?

I think there's a lot to be said for the idea of `select`ing on the sources, in one place, rather than `select`ing on `deps` in many places.

segiddins:
> Sounds like this isn't something folks want.

This is _definitely_ something that we'd want and would find useful

brentleyjones:
Same, we would also want this.

allevato:
The reason for this is deliberate; there's a 1:1 correspondence between `swift_library` targets and Swift modules. If there are no sources, there can be no module.

Why does this matter? If you have `swift_library` A depending on `swift_library` B, presumably A also imports B somewhere, right? If it didn't, you wouldn't need the dependency. But if you remove the sources from B and don't compile anything, then you no longer have a module, and A can't import B.

So if you're conditionally selecting empty or non-empty `srcs` for a module, you'd have to *also* conditionally control the `import` statement in all of the sources that depend on that module. You're either controlling that through a `-D` flag like `#if I_GOT_B` that you're already selecting elsewhere, or you're using `#if canImport(B)`, in which case your build graph is filled with lies because someone looking at A sees that you have B _unconditionally_ in its deps but then, under some configurations, it turns out that it isn't actually a module and thus it can't be imported. (This is why the analogy to `cc_library` doesn't apply; in C++, you just include headers from your deps based on their filename; they're either in the `hdrs` attribute or not they're not. There's no notion of "did this thing compile an artifact that I can use?")

A `srcs`-less `swift_library` would also be useless in a future world where we want to enforce strict-deps to prevent Swift's overly-permissive behavior of letting people import transitive dependencies.

> Let's say a `swift_library` has no sources, is there any reason at that point to build deps? Shouldn't that be a cut off valve?

That's the thing—the behavior you just described is inconsistent with the build graph you'd be writing. If A depends on B and B depends on C, there's a path from A to C, so having the compilation and linking of C be based on whether B has _sources_ is a bit too magical. Availability of deps should be controlled in deps where it's obvious to the user what's going on.

segiddins:
It matters for us since we're packaging swift modules & objective C modules into a single framework. If there's no swift sources, the framework can still be imported. Same if there are no objc headers. As it stands right now, we need a loading time check for any `.swift` files, instead of being able to go through a common code path regardless of what the set of passed-in `srcs` is. The inconsistency with `cc_library` and `objc_library` makes our macros significantly more complicated than they would otherwise need to be, and it forces us to have conditionals that, in my opinion, shouldn't be necessary.

allevato:
The conditionals you're describing sound like exactly what I'd expect someone to do, because you shouldn't create targets in your build graph that don't actually do anything. You say that the conditionals are unnecessary, but that's not really true—they are necessary but you'd rather have them pushed down into the rules so that somebody else has to manage the complexity.

Ultimately though, the problem arising from your use case of a mixed language framework problem is a separate issue. `swift_library` doesn't support mixed language modules, and it's not clear to me that the right solution for that is to force `swift_library` to make design choices that it otherwise wouldn't make solely to shoulder the burden of that use case, instead of writing a _new_ rule that uses the `swift_common` and `cc_common` APIs to handle compilation of both languages along with all the other subtleties associated with building a mixed language module. (Caveat: once `cc_common` supports Objective-C compilation, which it doesn't today.)

That last caveat may be frustrating, but adding capabilities to `swift_library` that we don't want long-term as a short-term workaround for deficiencies elsewhere is a recipe for maintenance burden and difficult-to-pay-off technical debt.

segiddins:
I'd think that targets in the build graph that do nothing shouldn't really be an issue, since they're not creating any actions?

allevato:
There's still an analysis cost associated with them, and they still show up in queries. Fundamentally, it's a question of whether you want your build graph to accurately represent your build. If you have targets that do nothing, even if they don't execute anything, they make it harder to reason about your graph.

PR #439, opened 2020-06-03
Allow generated headers with path separators

michaeleisel:
E.g., I work on a large project which has always used `#import <SomeModule/SomeModule-Swift.h>` for swift-to-objc, but using `SomeModule/SomeModule-Swift.h` as the generated header for our bazel migration causes it to fail

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1598,11 +1598,11 @@ def _declare_validated_generated_header(actions, generated_header_name):
     Returns:
         A `File` that should be used as the output for the generated header.
     """
-    if "/" in generated_header_name:
+    """if "/" in generated_header_name:
         fail(
             "The generated header for a Swift module may not contain " +
             "directory components (got '{}').".format(generated_header_name),
-        )
+        )"""
 
     extension = paths.split_extension(generated_header_name)[1]
     if extension != ".h":
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1588,8 +1588,7 @@ def _declare_multiple_outputs_and_write_output_file_map(
 def _declare_validated_generated_header(actions, generated_header_name):
     """Validates and declares the explicitly named generated header.
 
-    If the file does not have a `.h` extension or conatins path separators, the
-    build will fail.
+    If the file does not have a `.h` extension, the build will fail.
 
     Args:
         actions: The context's `actions` object.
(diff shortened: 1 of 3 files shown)


michaeleisel:
i've signed the cla

michaeleisel:
@googlebot I signed it!

michaeleisel:
@googlebot I fixed it.

michaeleisel:
@keith @kastiglione are you guys the ones to reach out to about discussing this? if not, who might be?

keith:
@allevato is the person to talk to

keith:
do you have more context on why this was disabled in the first place? I'm assuming there was some reason so it would be nice to be clear on the trade-offs

segiddins:
The original conversation on the restriction came from https://github.com/bazelbuild/rules_swift/pull/350#discussion_r348136234

keith:
Ah nice, thanks for that context. I definitely get the restriction imposed there

michaeleisel:
for us, we're dealing with an existing xcodebuild setup that relies on generated swift headers to be at certain include paths, and so not having this means we have to work around that which gets hairy iirc

segiddins:
Looks like `ctx.actions.declare_file` already does validation: (from another project) `the output artifact 'tests/framework/foo' is not under package directory 'tests/framework/platforms' for target '//tests/framework/platforms:framework_macos_101000_ios_100000_tvos_100000_watchos_30200'` when passing in `../foo` as the path. on the other hand, it looks like you _can_ `declare_file` into a subpackage...

Overall, I think it's probably OK to remove the restriction, especially since it's a thing you can do in Xcode.

thii:
There're some ways to work around this now:

- Replace `#import <SomeModule/SomeModule-Swift.h>` by `#import "SomeModule-Swift.h"` in your code-base. Xcode works with this too.
- Generate and use a header map
- Re-organize your code-base so that the `SomeModule` target is at the `//SomeModule` package. Because `BINDIR` is included by default, you can use `#import <SomeModule/SomeModule-Swift.h>`.

michaeleisel:
it's tough in this case because bazel support for this library is experimental, so we don't want to make sweeping changes for the sake of bazel yet. BUILD.bazel files are generated dynamically based on xcodegen, the standard there. header maps are a road i don't want to go down yet. my disposition is still to merge. who can we ping that has merge permission?

michaeleisel:
@keith thoughts on merging, based on @segiddins comment?

keith:
@googlebot I consent.

PR #440, opened 2020-06-05
Fix passing -num-threads with -wmo

keith:
Previously the assumption was that if `-num-threads` is `1`, only a
single object is emitted. It turns out in swiftc there is a difference
between 0 and 1 to this value, so even with 1 multiple objects are
emitted.

Without this change linking fails when passing this flag like:

```
bazel build examples/xplatform/xctest --swiftcopt=-num-threads --swiftcopt=1 --swiftcopt=-wmo
```

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -2029,7 +2029,10 @@ def _emitted_output_nature(is_wmo_implied_by_features, user_compile_flags):
     # `-num-threads 12` to the command line. We need to stage that as our
     # initial default here to ensure that we return the right value if the user
     # compile flags don't otherwise override it.
-    num_threads = _DEFAULT_WMO_THREAD_COUNT if is_wmo else 1
+    #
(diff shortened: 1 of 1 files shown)


kastiglione:
> It turns out in swiftc there is a difference between 0 and 1 to this value, so even with 1 multiple objects are emitted

worth a comment in the code in my opinion

swift/internal/compiling.bzl
```diff
@@ -2029,7 +2029,10 @@ def _emitted_output_nature(is_wmo_implied_by_features, user_compile_flags):
     # `-num-threads 12` to the command line. We need to stage that as our
     # initial default here to ensure that we return the right value if the user
     # compile flags don't otherwise override it.
-    num_threads = _DEFAULT_WMO_THREAD_COUNT if is_wmo else 1
+    #
+    # 0 is the only option that makes swift emit a single object file, anything
+    # greater and it uses that number of threads to emit multiple objects
```
kastiglione, line 2034: 👍

PR #444, opened 2020-07-02
Add loads for {proto_library,cc_proto_library} to //third_party/bazel_protos

Yannic:


Yannic:
Oopsi, forgot that I already sent #386

PR #462, opened 2020-08-07
Run tests on CI

keith:


keith:
Current issues:

- [ ] macOS tests are failing (locally as well)
- [ ] Linux Swift version is outdated, so some tests fail https://github.com/bazelbuild/continuous-integration/pull/1015

keith:
Going to enable the Linux tests in a follow up PR after the docker image change is deployed

keith:
https://github.com/bazelbuild/rules_swift/pull/463

Issue #305, opened 2019-09-05
Add .clang-format settings file


kastiglione:
I ran `clang-format` on the changes I made in #304, but it resulted in a few changes that did not match the existing formatting/style of the rest of the code. For example the indentation of `public:` and `private:`.

Adding a `.clang-format` file would solve this.

allevato:
A `.clang-format` that's just `BasedOnStyle: Google` _should_ be sufficient here, as long as public clang-format does the same thing as ours internally.

PR #467, opened 2020-08-11
Add .clang-format

keith:
Fixes https://github.com/bazelbuild/rules_swift/issues/305


thii:
I would run clang-format on the current codebase and apply those changes too.

keith:
https://github.com/bazelbuild/rules_swift/pull/470

PR #469, opened 2020-08-11
Test clang format

keith:


keith:
this binary doesn't exist on CI. too bad. not a huge deal since it's not like there are too many of these, they have too much code, or they change too much.

PR #470, opened 2020-08-11
Run clang-format on everything

keith:


keith:
Also seeing if this binary exists on CI so we could validate https://github.com/bazelbuild/rules_swift/pull/469

PR #473, opened 2020-08-13
Use a parameter file when linking Swift libraries when the host machine is macOS

thii:
This is to prevent the linking failures if a `swift_library` target has
too many files that can cause its linking args to become longer than the
system's command length limits. This does not change the behavior when
building on Linux yet, since `ar` does not support passing arguments
using a parameter file.

Redo of https://github.com/bazelbuild/rules_swift/pull/349 with platform-awareness. This is critical if you have a very large Swift target.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -58,7 +58,15 @@ def _register_static_library_link_action(
     )
     args = actions.args()
     args.add_all(command_line)
-    args.add_all(objects)
+
+    filelist_args = actions.args()
+    if swift_toolchain.linker_supports_filelist:
+        args.add("-filelist")
+        filelist_args.set_param_file_format("multiline")
+        filelist_args.use_param_file("%s", use_always = True)
+        filelist_args.add_all(objects)
+    else:
+        args.add_all(objects)
 
     env = cc_common.get_environment_variables(
         action_name = CPP_LINK_STATIC_LIBRARY_ACTION_NAME,
@@ -73,7 +81,7 @@ def _register_static_library_link_action(
     execution_requirements = {req: "1" for req in execution_requirements_list}
 
     actions.run(
-        arguments = [args],
+        arguments = [args, filelist_args],
         env = env,
         executable = archiver_path,
         execution_requirements = execution_requirements,
--- a/swift/internal/swift_toolchain.bzl
+++ b/swift/internal/swift_toolchain.bzl
@@ -168,6 +168,12 @@ def _swift_toolchain_impl(ctx):
         toolchain_root,
     )
 
+    # The static archiver on Linux doesn't currently support the `-filelist`
+    # flag. This should be handled by the C++ toolchain once we migrate away
+    # from invoking the linker directly for static linking.
+    # See https://github.com/bazelbuild/rules_swift/pull/349#issuecomment-550353580.
+    linker_supports_filelist = True if ctx.attr.os == "darwin" else False
+
     # Combine build mode features, autoconfigured features, and required
     # features.
     requested_features = features_for_build_modes(ctx)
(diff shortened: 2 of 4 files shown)


swift/internal/linking.bzl
```diff
@@ -73,7 +95,7 @@ def _register_static_library_link_action(
     execution_requirements = {req: "1" for req in execution_requirements_list}
 
     actions.run(
-        arguments = [args],
+        arguments = [args, filelist_args],
```
brentleyjones, line 98: Is there a way to add `filelist_args` to `args`, then we won't need to expose `filelist_args` outside of the `if`?

swift/internal/linking.bzl
```diff
@@ -58,7 +70,17 @@ def _register_static_library_link_action(
     )
     args = actions.args()
     args.add_all(command_line)
-    args.add_all(objects)
+    filelist_args = actions.args()
+
+    # Use a parameter file when building on macOS. `ar` on Linux does not
+    # support this yet, so we're passing parameters to the command line as is.
+    if _is_apple_platform_toolchain(swift_toolchain):
```
thomasvl, line 77: Drive by - this is defeating some of the toolchain abstraction.  Normally the toolchain would be the thing exposing if it support `-filelist` rather than code here figuring out what toolchain is being used.

keith: +1 to this, it actually even looks like this is done sometimes in the macOS crosstool, can we make that apply to this use case upstream as well?

swift/internal/linking.bzl
thii: It doesn't seem like it's possible.

swift/internal/linking.bzl
thii: This was more like a workaround, until we can migrate to using `cc_common.create_linking_context_from_compilation_outputs`. I'll add a new attribute to `SwiftToolchainInfo` instead.

swift/internal/linking.bzl
```diff
@@ -58,7 +58,15 @@ def _register_static_library_link_action(
...
-    args.add_all(objects)
+
+    filelist_args = actions.args()
+    if swift_toolchain.linker_supports_filelist:
+        args.add("-filelist")
+        filelist_args.set_param_file_format("multiline")
+        filelist_args.use_param_file("%s", use_always = True)
+        filelist_args.add_all(objects)
+    else:
+        args.add_all(objects)
```
keith, line 69: can't args files be nested? so you could `args.add(filelist_args)`?

swift/internal/swift_toolchain.bzl
```diff
@@ -168,6 +168,12 @@ def _swift_toolchain_impl(ctx):
         toolchain_root,
     )
 
+    # The static archiver on Linux doesn't currently support the `-filelist`
+    # flag. This should be handled by the C++ toolchain once we migrate away
+    # from invoking the linker directly for static linking.
+    # See https://github.com/bazelbuild/rules_swift/pull/349#issuecomment-550353580.
+    linker_supports_filelist = True if ctx.attr.os == "darwin" else False
```
keith, line 175: does this toolchain ever get used on darwin? I think you can only get the xcode one there?

keith, line 175: This could be:

```bzl
    linker_supports_filelist = ctx.attr.os == "darwin"
```

swift/internal/linking.bzl
keith: ah I see this was asked below as well

swift/internal/swift_toolchain.bzl
thii: I feel so dumb 🤦🏼‍♂️

thii: Ahh yeah

keith: haha, nbd, one of those things you don't notice without fresh eyes

thii: https://github.com/bazelbuild/rules_swift/pull/479

PR #478, opened 2020-08-31
Read the `custom_malloc` configuration field in `swift_{binary,test}` to use an allocator provided by the `--custom_malloc` flag, if any.

keith:
PiperOrigin-RevId: 329317960

### DIFF
--- a/swift/internal/swift_binary_test.bzl
+++ b/swift/internal/swift_binary_test.bzl
@@ -78,6 +78,15 @@ into the binary. Possible values are:
             "_cc_toolchain": attr.label(
                 default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
             ),
+            # A late-bound attribute denoting the value of the `--custom_malloc`
+            # command line flag (or None if the flag is not provided).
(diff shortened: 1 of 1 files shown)


swift/internal/swift_binary_test.bzl
```diff
@@ -215,10 +224,8 @@ def _swift_linking_rule_impl(
 
     # If a custom malloc implementation has been provided, pass that to the
     # linker as well.
-    if ctx.attr.malloc:
-        additional_linking_contexts.append(
-            ctx.attr.malloc[CcInfo].linking_context,
-        )
+    malloc = ctx.attr._custom_malloc or ctx.attr.malloc
```
segiddins, line 227: will this always exist? should it be `if malloc:` still ?

keith: seems like it always does, even if it's an empty library https://github.com/bazelbuild/bazel/blob/39bc97eab295bddb35b38bfc4a2ff3d2b15d034e/tools/osx/crosstool/BUILD.tpl#L27-L29

PR #482, opened 2020-09-03
Update bzl_library targets

segiddins:
Mostly so depending on them from a stardoc rule works now we load rules_proto

Mostly autogenerated & then tweaked based on the gazelle in bazelbuild/bazel-skylib#271

### DIFF
--- a/swift/internal/BUILD
+++ b/swift/internal/BUILD
@@ -5,7 +5,7 @@ licenses(["notice"])
 bzl_library(
     name = "actions",
     srcs = ["actions.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":features",
         ":toolchain_config",
@@ -17,7 +17,7 @@ bzl_library(
 bzl_library(
     name = "attrs",
     srcs = ["attrs.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":providers",
         "@bazel_skylib//lib:dicts",
@@ -27,7 +27,7 @@ bzl_library(
 bzl_library(
     name = "autolinking",
     srcs = ["autolinking.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":actions",
         ":toolchain_config",
@@ -37,7 +37,7 @@ bzl_library(
 bzl_library(
     name = "compiling",
     srcs = ["compiling.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":actions",
         ":autolinking",
@@ -59,7 +59,7 @@ bzl_library(
 bzl_library(
     name = "debugging",
     srcs = ["debugging.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":actions",
         ":derived_files",
@@ -70,7 +70,7 @@ bzl_library(
 bzl_library(
     name = "derived_files",
     srcs = ["derived_files.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":utils",
         "@bazel_skylib//lib:paths",
@@ -80,7 +80,7 @@ bzl_library(
 bzl_library(
     name = "features",
     srcs = ["features.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":feature_names",
         "@bazel_skylib//lib:collections",
@@ -91,7 +91,7 @@ bzl_library(
 bzl_library(
     name = "linking",
     srcs = ["linking.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":derived_files",
         "@bazel_skylib//lib:collections",
@@ -103,14 +103,14 @@ bzl_library(
 bzl_library(
     name = "module_maps",
     srcs = ["module_maps.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = ["@bazel_skylib//lib:paths"],
 )
 
 bzl_library(
     name = "proto_gen_utils",
     srcs = ["proto_gen_utils.bzl"],
-    visibility = ["//visibility:public"],
+    visibility = ["//swift:__subpackages__"],
     deps = [
         ":utils",
         "@bazel_skylib//lib:paths",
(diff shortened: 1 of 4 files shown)


swift/internal/BUILD
```diff
@@ -3,107 +3,362 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
...
+    deps = [
         "@bazel_skylib//lib:partial",
         "@bazel_skylib//lib:paths",
+        "@bazel_skylib//lib:types",
     ],
 )
 
-# Consumed by Bazel integration tests.
-filegroup(
-    name = "for_bazel_tests",
```
keith, line 98: might still need this filegroup?

keith (approved):

swift/internal/BUILD
segiddins: whoops -- will revert

swift/internal/BUILD
```diff
@@ -3,96 +3,366 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
...
-        "swift_library.bzl",
-        "swift_module_alias.bzl",
-        "swift_proto_library.bzl",
-        "swift_protoc_gen_aspect.bzl",
-    ],
-    visibility = [
-        "//swift:__pkg__",
+    name = "actions",
+    srcs = ["actions.bzl"],
+    visibility = ["//visibility:public"],
```
thomasvl, line 8: Drive by - why make things in the internal directory public, that seems to say it is ok for things to directly depend on them?

segiddins: I think this is what gazelle did to the visibility automatically?

keith: This is an interesting case. So buildifier warns against internal imports from other directories, but you can disable that warning inline. In which case you'd need to import this library for stardoc. I think having these public is nice to not ding folks twice on the same internal visibility issue, but OTOH if you're not using buildifier you may not realize you shouldn't import this (even though the directory name implies it).

keith: Yea I get what you mean, maybe this is a limitation of stardoc more than anything else if we could ignore dependencies maybe that would be good enough. the specific problem we face is because of this issue https://github.com/bazelbuild/rules_apple/issues/319 we have intentionally chosen to load some internal API from rules_apple, taking on the maintenance risk that those could change out from under us, which means for us to generate stardoc for our files, these either have to be public, or we have to run the generation with `--nocheck_visibility`

thomasvl: For stardoc you could still depend on the output public things that transitively provide the files.

keith: you mean that you would find other public bzl_library targets that transitively depend on the same files and import those instead?

keith: FWIW I don't particularly think we should keep these public, I just noticed when trying to use stardoc for our repo that using these internal imports was a blocker unless we used `--nocheck_visibility` or made all these targets public. But really we should solve that issue in another way

keith: Maybe we should file a stardoc / bazel issue about these being public when they're in `internal` directories too?

thomasvl: > you mean that you would find other public bzl_library targets that transitively depend on the same files and import those instead?

yea, it _works_ and you can comment/document it there also.

keith: seems like in the meantime we should check these in with more strict vis, and revisit later if we want

segiddins: I personally think they _should_ remain public.

1) there's no `default_visibility` for the `internal` package, which means the bzl files _are_ effectively public and _can_ be imported
2) changing it so these files are only accessible to a `bzl_library` as deps via another `bzl_library` target  means that anyone trying to use the skylib gazelle won't be able to automate bzl_library generation

thomasvl: Isn't default visibility private? – https://docs.bazel.build/versions/master/visibility.html#visibility-of-a-rule-target

Re the plugin - that only impact someone using the plugin for exactly this module.  Everyone else should be using the .bzl files (and bzl_library-s) that are public and aren't within the internal directory.

segiddins: For rules, sure, but I don't think it is for `load`ing `bzl` files.

> Re the plugin - that only impacts someone using the plugin for exactly this module

i don't think that's true -- the skylib gazelle will add as a dep for an external load the loaded string verbatim, but with the `.bzl` extension stripped. See, for example, the generated deps in this file on the skylib lib files

thomasvl: > For rules, sure, but I don't think it is for `load`ing `bzl` files.

Not sure I follow.  If they are only loading the public `bzl` files, then the matching `bzl_library` it would want to reference would also be public, no?  i.e. - if they aren't loading things from internal, then why would it generate references to these?

> > Re the plugin - that only impacts someone using the plugin for exactly this module
> 
> i don't think that's true -- the skylib gazelle will add as a dep for an external load the loaded string verbatim, but with the `.bzl` extension stripped. See, for example, the generated deps in this file on the skylib lib files

Same. External things shouldn't be loading from internal, so how is this a problem?

Anyway, it is up to you all, I guess I'm missing something.

segiddins: I don't think the stuff in internal is actually internal, if I'm understanding correctly I think that's where the mismatch comes from

keith: that's not enforced in bazel, but buildifier enforces it (which is what i was talking about above) so ideally `internal` things were actually not `load`-able by things outside of their parent directories

keith: So ideally you could treat things there as private API and change as desired, while opening this up might reduce that ability, might end up having folks depend on this stuff more and we might break them

keith: as far as automating these library updates go we can post-process with something like:

```
buildozer 'set visibility //swift:__subpackages__' swift/internal:%bzl_library
```

segiddins: Updated visibility

achew22:
Out of curiosity, what manual tweaks did you need to make after you had fixed relative path generation?

segiddins:
@achew22 changing visibility for the `bzl_library` targets under `test` and `swift/internal`, and changing a couple of references to point to `filegroup` targets inside `@bazel_tools`, since there were no `bzl_library` targets to use as deps there

achew22:
Is there a preexisting filegroup insize of `bazel_tools` that has the right provider to be a `dep` of a `bzl_library`? Maybe I should special case those and have them resolve automatically. Looking at the PR it looks like you only targeted the specific files, which, thankfully, were already exported.

If you remember more specifically what was wrong about the visibility in `test` and `swift/internal`, and especially if you wanna send a PR that has a failing test case, I would really like to learn more.

segiddins:
Nothing was _wrong_, per se, see the discussion up in the history for this PR -- it was decided we didn't want those `bzl_library` targets to be public because they're not intended to be consumed directly outside of this repo (but they are still allowed to be consumed in a "buyers beware, this is unsupported" kind of way).

Maybe I was mis-remembering about the file group -- either way, I _think_ `bzl_library` just cares about `files`, so any `file` or target that exports files in its `DefaultInfo` provider should do the trick iirc.

Either way, I'm not sure what the answer is for how the gazelle generator should handle these cases. Ideally, we'd be able to run gazelle and not have to remember to make these changes every time. On the other hand, it becomes less of a big deal once we switch doc generation over to stardoc, because then we'll essentially be able to test that our `bzl_library` targets are correct.

PR #486, opened 2020-09-11
Remove apple_support dependency

keith:


keith:
Maybe we should use it for https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/xcode_swift_toolchain.bzl#L479 Instead

PR #489, opened 2020-09-15
Implement -coverage-prefix-map feature

keith:
This should ship in Swift 5.3.x https://github.com/apple/swift/pull/32175


keith:
This is in Xcode 12.2b2, so I expect it to be in the next release

segiddins (approved):
Thanks for the tests!

PR #488, opened 2020-09-13
Added helper macro for declaring transitive dependencies

UebelAndre:
I first saw this pattern in [@com_github_grpc//bazel:grpc_extra_deps.bzl](https://github.com/grpc/grpc/blob/master/bazel/grpc_extra_deps.bzl) and thought it was a great way to group dependencies. Should any dependencies be added or removed in future changes, users will more seamlessly be able to update.

Sorry, I didn't file an issue ticket first. I should have read `CONTRIBUTING.md` before opening the pull request 😅. I feel the changes here are small and straight forward enough that any discussion can happen here. Though I expect this to either be rejected or merged potential pending some name changes.

PR #494, opened 2020-09-21
Restore docs until next release

UebelAndre:
@keith Hey, super sorry to ping you a bunch and then have you merge something that's not good. The docs shouldn't have been updated in #488 because because that macro is not in that release and there's no way to know what the commit/sha will be in adavance. This restores docs so users don't end up copy/pasing the wrong thing. Though, in the next release, this should be reverted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,18 @@ load(
 swift_rules_dependencies()
 
 load(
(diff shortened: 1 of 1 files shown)


keith:
Thanks but I created https://github.com/bazelbuild/rules_swift/releases/tag/0.15.0 instead

PR #497, opened 2020-09-22
Add use_default_shell_env to run / run_shell

keith:
This fixes https://github.com/bazelbuild/bazel/issues/12049 and
apparently is the recommendation wherever possible.


keith:
tbd if this is mutually exclusive with passing `env`

PR #498, opened 2020-09-30
Update rules_swift to use the toolchain transition.

keith:
This is phase 2 of of the switch to toolchain transitions. See https://github.com/bazelbuild/bazel/issues/11584 for details.

This shouldn't be merged into rules_swift's master branch until https://cs.opensource.google/bazel/bazel/+/58fdf63d81cc71f0315918b111fc56f4c039f1a5 is in a Bazel release (which should be 3.7, https://github.com/bazelbuild/bazel/issues/12188).

PiperOrigin-RevId: 334610489


keith:
as the original message said we need to wait on this one, but this way we hopefully won't forget

keith:
sounds like this can be merged now it'll just be a no-op

segiddins (approved):

PR #502, opened 2020-10-08
Pass default_shell_env through actions

keith:
Through the investigation of
https://github.com/bazelbuild/bazel/issues/12049 one of the things I
discovered was that when using `actions.run` there are 2 options for
environment variables. `use_default_shell_env = True` is recommended,
but cannot be used if you want to also pass `env` to the actions. To
support Xcode version selection we have to pass `env` with those
variables. Without the default shell env, we only get the environment
variables defined in the crosstool, but not those passed with
`--action_env`. This now adds variables passed as
`--action_env=FOO=BAR`, but not those passed as `--action_env=FOO`
(where the value is supposed to pass through).

This is useful to ensure a few things:

1. The default PATH things are executed with includes /usr/local/bin.
   This can result in pollution of binaries from homebrew. Previously
   there was no way to limit this
2. This should be a good replacement for using custom Swift toolchains.
   Currently those environment variables only apply to some actions
   (excluding those from bazel) using `--action_env=TOOLCHAINS=foo`
   should work better than the current solution (this change can be made
   as a followup)

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -15,6 +15,7 @@
 """Implementation of linking logic for Swift."""
 
 load("@bazel_skylib//lib:collections.bzl", "collections")
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:partial.bzl", "partial")
 load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
@@ -27,6 +28,7 @@ def _register_static_library_link_action(
         cc_feature_configuration,
         objects,
         output,
+        env,
         swift_toolchain):
     """Registers an action that creates a static library.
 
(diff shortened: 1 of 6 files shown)


segiddins:
So I was thinking of an alternative -- instead of trying to merge in the configuration's default shell env, what if we used `env` to add in the env vars we want, and call the executables that way?

keith:
Use which `env` to add them where?

segiddins:
`/usr/bin/env FOO=BAR ... -- executable args...`

keith:
ah, I think piping that through would probably be a bit more difficult than this. I'm not sure we want to compete with bazel's logic for this

brentleyjones:
FYI: I've been using this with `--action_env=TOOLCHAINS=$TOOLCHAINS` to great effect.

brentleyjones (approved):

keith:
https://github.com/bazelbuild/rules_swift/pull/1114

PR #504, opened 2020-10-12
Introduce feature to register a separate action to generate swift derived files

ghvg1313:
theoretically this would speed up incremental builds but cause slightly worse clean build time

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -108,7 +108,7 @@ def compile_action_configs():
             configurators = [_output_object_or_file_map_configurator],
         ),
         swift_toolchain_config.action_config(
-            actions = [swift_action_names.DERIVED_FILES],
+            actions = [swift_action_names.DERIVE_FILES],
             configurators = [_output_swiftmodule_or_file_map_configurator],
         ),
 
@@ -136,13 +136,13 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [_emit_module_path_configurator],
             not_features = [SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION],
         ),
         swift_toolchain_config.action_config(
-            actions = [swift_action_names.DERIVED_FILES],
+            actions = [swift_action_names.DERIVE_FILES],
             configurators = [_emit_module_path_configurator],
         ),
 
@@ -150,7 +150,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [
                 swift_toolchain_config.add_arg("-enable-library-evolution"),
@@ -163,7 +163,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [_emit_module_interface_path_configurator],
             features = [
@@ -176,7 +176,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [_emit_objc_header_path_configurator],
             not_features = [
@@ -185,7 +185,7 @@ def compile_action_configs():
             ],
         ),
         swift_toolchain_config.action_config(
-            actions = [swift_action_names.DERIVED_FILES],
+            actions = [swift_action_names.DERIVE_FILES],
             configurators = [_emit_objc_header_path_configurator],
             not_features = [SWIFT_FEATURE_NO_GENERATED_HEADER],
         ),
@@ -195,7 +195,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [_stats_output_dir_configurator],
             features = [SWIFT_FEATURE_COMPILE_STATS],
@@ -214,7 +214,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [
                 swift_toolchain_config.add_arg("-DDEBUG"),
@@ -224,7 +224,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [
                 swift_toolchain_config.add_arg("-DNDEBUG"),
@@ -238,7 +238,7 @@ def compile_action_configs():
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVED_FILES,
+                swift_action_names.DERIVE_FILES,
             ],
             configurators = [
                 swift_toolchain_config.add_arg("-Onone"),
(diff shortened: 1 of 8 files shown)


ghvg1313:
@googlebot I fixed it.

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -342,6 +343,7 @@ def _all_action_configs(
             swift_toolchain_config.action_config(
                 actions = [
                     swift_action_names.COMPILE,
+                    swift_action_names.DERIVED_FILES,
```
keith, line 346: Should this action be added to all actions that currently contain `.COMPILE`? Or is there a reason it's excluded from some above like the bitcode ones?

swift/internal/feature_names.bzl
```diff
@@ -222,3 +222,8 @@ SWIFT_FEATURE_NO_EMBED_DEBUG_MODULE = "swift.no_embed_debug_module"
 # files where the protoc command line might not be crafted correctly, so it
 # remains opt in.
 SWIFT_FEATURE_GENERATE_FROM_RAW_PROTO_FILES = "swift.generate_from_raw_proto_files"
+
+# If enabled and whole module optimisation is being used, the `*.swiftdoc`,
+# `*.swiftmodule` and `*-Swift.h` are generated with a separate action
+# rather than as part of the compilation.
+SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION = "swift.split_derived_files_generation"
```
keith, line 229: Not sure about the naming of derived files, but I don't have a better suggestion, and I do like that it's not super tied to the current things it outputs.

swift/internal/compiling.bzl
```diff
@@ -219,7 +249,10 @@ def compile_action_configs():
         # If the `swift.opt_uses_wmo` feature is enabled, opt builds should also
         # automatically imply whole-module optimization.
         swift_toolchain_config.action_config(
-            actions = [swift_action_names.COMPILE],
+            actions = [
+                swift_action_names.COMPILE,
+                swift_action_names.DERIVED_FILES
```
keith, line 254: Same question from the other file, should almost all things with `.COMPILE` currently also have this new one? I believe right below this with the serialize debugging options actions we do need to add this one as well?

swift/internal/compiling.bzl
```diff
@@ -1161,18 +1249,38 @@ def compile(
...
-    ]) + compile_outputs.object_files + other_outputs
+    if is_feature_enabled(
+        feature_configuration = feature_configuration,
+        feature_name = SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION,
+    ):
+        all_compile_outputs = compact([
+            # The `.swiftmodule` file is explicitly listed as the first output
+            # because it will always exist and because Bazel uses it as a key for
+            # various things (such as the filename prefix for param files generated
+            # for that action). This guarantees some predictability.
```
keith, line 1260: I think we should move this to the assignment below instead?

swift/internal/compiling.bzl
```diff
@@ -1161,18 +1249,38 @@ def compile(
...
+            compile_outputs.swiftinterface_file,
+            compile_outputs.indexstore_directory,
+            compile_outputs.stats_directory,
+        ]) + compile_outputs.object_files
+        all_derived_outputs = compact([
+            compile_outputs.swiftmodule_file,
+            compile_outputs.swiftdoc_file,
+            compile_outputs.generated_header_file,
+        ]) + other_outputs
```
keith, line 1269: I wonder if we should rename `other_outputs` at some point since it's unclear without reading the implementation whether it should be in this bucket or the other one, but this looks right for now

swift/internal/compiling.bzl
ghvg1313: Good catch, I think I missed some new changes during the cherry-pick.
It seems some of the actions are for xcode 12? Could you help do some sanity check just to make sure this works? We currently don't have a proper xcode 12 branch that's compilable here.

keith: Looks like there are a few more missing:

```patch
-  arguments: "-Onone"
-  arguments: "-Xfrontend"
-  arguments: "-no-serialize-debugging-options"
-  arguments: "-gline-tables-only"
-  arguments: "-Xwrapped-swift=-debug-prefix-pwd-is-dot"
```

these _might_ no matter that much but I think we probably want to keep them as aligned as possible

swift/internal/feature_names.bzl
brentleyjones: The name matches Xcode parlance, so I personally agree with it.

keith: sounds good lets keep it

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -287,6 +287,7 @@ def _all_action_configs(
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
+                swift_action_names.DERIVED_FILES,
```
segiddins, line 290: how should the feature work when using the non-Xcode toolchain?

segiddins (commented):
Overall I'm in favor of this change, but I'd like to see some test coverage with the feature enabled that shows the correct actions being registered, and also show it works as expected with wmo (because wmo changes the action graph that we use)

swift/internal/compiling.bzl
```diff
@@ -95,17 +96,27 @@ def compile_action_configs():
     action_configs = [
         # Emit object file(s).
         swift_toolchain_config.action_config(
-            actions = [swift_action_names.COMPILE],
+            actions = [
+                swift_action_names.COMPILE,
+                swift_action_names.DERIVED_FILES
```
keith, line 101: shouldn't this one not be here? we don't want `-emit-object` to be passed right?

swift/internal/compiling.bzl
```diff
@@ -95,17 +96,27 @@ def compile_action_configs():
...
                 swift_toolchain_config.add_arg("-emit-object"),
             ],
         ),
 
         # Add the single object file or object file map, whichever is needed.
         swift_toolchain_config.action_config(
-            actions = [swift_action_names.COMPILE],
+            actions = [
+                swift_action_names.COMPILE,
+                swift_action_names.DERIVED_FILES
```
keith, line 112: I don't think we want it here either, since the lines right below this you added could add the file map as well?

swift/internal/xcode_swift_toolchain.bzl
keith: I think theoretically it could work the exact same way.

keith: I have a local change that makes this work on Linux, so seems fine as long as we set it up correctly

keith:
@googlebot I consent.

brentleyjones (approved):

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -189,6 +199,7 @@ _FEATURE_CHECKS = {
     SWIFT_FEATURE_ENABLE_BATCH_MODE: _check_enable_batch_mode,
     SWIFT_FEATURE_SUPPORTS_PRIVATE_DEPS: _check_supports_private_deps,
     SWIFT_FEATURE_USE_RESPONSE_FILES: _check_use_response_files,
+    SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES: _check_skip_function_bodies,
```
brentleyjones, line 202: Do we want to alphabetize this?

swift/internal/feature_names.bzl
```diff
@@ -232,3 +232,7 @@ SWIFT_FEATURE_GENERATE_FROM_RAW_PROTO_FILES = "swift.generate_from_raw_proto_fil
 # `*.swiftmodule` and `*-Swift.h` are generated with a separate action
 # rather than as part of the compilation.
 SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION = "swift.split_derived_files_generation"
+
+# If enabled the skip function bodies frontend flag is passed when using derived
+# files generation. This requires Swift 5.2
+SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES = "swift.enable_skip_function_bodies"
```
segiddins, line 238: should this include `derived_files` somewhere in the feature name?
Is the only reason not to default this to on the dependency on Swift 5.2+ ?

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -601,6 +602,10 @@ def _xcode_swift_toolchain_impl(ctx):
         requested_features.append(SWIFT_FEATURE_SUPPORTS_LIBRARY_EVOLUTION)
         requested_features.append(SWIFT_FEATURE_SUPPORTS_PRIVATE_DEPS)
 
+    # Xcode 11.4 implies Swift 5.1.
```
segiddins, line 605: The comment above says

```
This requires Swift 5.2
```

so there's a mismatch somewhere

ghvg1313:
Thanks for adding the tests!

segiddins (approved):

PR #507, opened 2020-10-20
Fix incremental compilation with derived files

keith:
For incremental compilation we pass an output file map to compile
actions. Since we have 2 actions with the derived files feature, they
were fighting to write the same set of files. In this case we now pass a
separate output file map to the derived files action that only contains
the swiftmodules, and one that only contains the objects to the standard
compile.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -785,7 +785,7 @@ def _output_object_or_file_map_configurator(prerequisites, args):
 def _output_swiftmodule_or_file_map_configurator(prerequisites, args):
     """Adds the output file map or single object file to the command line."""
     return _output_or_file_map(
-        output_file_map = prerequisites.output_file_map,
+        output_file_map = prerequisites.derived_files_output_file_map,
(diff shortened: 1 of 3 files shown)


swift/internal/compiling.bzl
```diff
@@ -1848,7 +1873,11 @@ def _declare_multiple_outputs_and_write_output_file_map(
...
             )
             other_outputs.append(partial_module)
-            src_output_map["swiftmodule"] = partial_module.path
+
+            if split_derived_file_generation:
+                derived_files_output_map[src.path] = {"swiftmodule": partial_module.path}
+            else:
+                src_output_map["swiftmodule"] = partial_module.path
 
         output_map[src.path] = struct(**src_output_map)
```
segiddins, line 1882: does this also need to go into the conditional?

keith: no because we still want this to be added because it contains just the `object` in the derived file case, and both `object` and `swiftmodule` in the default case

segiddins: 👍

PR #510, opened 2020-10-21
Fix swift build rules not recognizing "licenses" attribute.

keith:
I'm reluctantly taking this even though no one should use it just so we
can ideally reduce future conflicts with upstream changes.

PiperOrigin-RevId: 338254161
(cherry picked from commit ef51377b00ac29b509733bc9ec20fce9ffe28e33)


keith:
can't take because bazel fails I guess

PR #511, opened 2020-10-21
Add swift.remap_xcode_path feature

keith:
This is analogous to https://github.com/bazelbuild/bazel/commit/bc5883a04ef5148038edd0080e1565a0cb1688b0

It leads to more hermetic binaries and shouldn't have any impact on
debugging.

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -239,4 +239,4 @@ SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES = "swift.skip_function_bodies_for_deri
 
 # If enabled remap the absolute path to Xcode in debug info. When used with
 # swift.coverage_prefix_map also remap the path in coverage data.
-SWIFT_REMAP_XCODE_PATH = "swift.remap_xcode_path"
+SWIFT_FEATURE_REMAP_XCODE_PATH = "swift.remap_xcode_path"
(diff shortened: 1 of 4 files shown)


segiddins (approved):

swift/internal/feature_names.bzl
```diff
@@ -236,3 +236,7 @@ SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION = "swift.split_derived_files_genera
 # If enabled the skip function bodies frontend flag is passed when using derived
 # files generation. This requires Swift 5.2
 SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES = "swift.skip_function_bodies_for_derived_files"
+
+# If enabled remap the absolute path to Xcode in debug info. When used with
+# swift.coverage_prefix_map also remap the path in coverage data.
+SWIFT_REMAP_XCODE_PATH = "swift.remap_xcode_path"
```
brentleyjones, line 242: Do we want `_FEATURE_` in the name?

keith: done

thii:
What happens if `-debug-prefix-map` is passed more than once and they all match? Does the order matter?

keith:
Looks like only the first path matches https://github.com/llvm/llvm-project/blob/0c15a1b4bcefc1eb434e07aa31b7a08109ccd27e/clang/lib/CodeGen/CGDebugInfo.cpp#L479-L489

keith:
Also see https://github.com/bazelbuild/bazel/pull/12354

PR #512, opened 2020-10-27
Migrate to the modern Starlark linker input API.

benjaminp:
See https://github.com/bazelbuild/bazel/issues/10860.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -95,6 +95,7 @@ def _register_static_library_link_action(
     )
 
 def register_libraries_to_link(
+        owning_label,
         actions,
         alwayslink,
         cc_feature_configuration,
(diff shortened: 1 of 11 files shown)


keith:
this looks great, but do you know what the status of this inside google? normally i would expect them to be pushing on migrating to this new API first?

benjaminp:
I'm out in the cold with you. Let's ask @oquenchil.

oquenchil:
google3 is not completely migrated but that's fine, the flag will be flipped in the Bazel source code and we can set the flag to False internally for a while in the bazelrc. There is pressure here because internally we can flip whenever we want whereas with Bazel we'd have to wait till Bazel 5.0 if we miss this one.

@allevato can you please review and import?

PR #521, opened 2020-11-11
Migrate Swift build rules to the new C++ `LinkerInput` APIs.

keith:
Based partially on https://github.com/bazelbuild/rules_swift/pull/512 by @benjaminp.

This change isn't required but should help decrease conflicts with upstream

PiperOrigin-RevId: 341831304
(cherry picked from commit bf9560d39502f33bfdf478199c9a7ef95ad3dc2b)


keith:
Doing https://github.com/bazelbuild/rules_swift/pull/522 instead to avoid any other potential misses

PR #547, opened 2021-01-08
Remove DEBUG flag from fastbuild compilation

omarzl:
I suggest that the flag should be only included in dbg compilation mode and not in fastbuild since I consider it should be used specifically for debugging purposes

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -213,7 +213,7 @@ def compile_action_configs():
             configurators = [
                 swift_toolchain_config.add_arg("-DDEBUG"),
             ],
-            features = [[SWIFT_FEATURE_DBG], [SWIFT_FEATURE_FASTBUILD]],
+            features = [[SWIFT_FEATURE_DBG]],
         ),
         swift_toolchain_config.action_config(
             actions = [


omarzl:
@googlebot I signed it!

keith:
I'm not sure what the history is here, but I guess given that `DEBUG` and `NDEBUG` by convention are not a tri-state, IMO fastbuild makes more sense to be `DEBUG` than `NDEBUG`, and by convention I guess it's fair for folks to assume that at least one of those will always be passed.

More practically this would be a breaking change, and it would diverge with bazel which has this same behavior for objc / c++ https://github.com/bazelbuild/bazel/blob/0f626a472b9fb6ca8a6da17448ba8ff518188870/tools/osx/crosstool/cc_toolchain_config.bzl#L5505-L5506

If this is causing issues for you we should probably cleanup https://github.com/bazelbuild/rules_swift/pull/428 instead and just disable these entirely

omarzl:
I understand your point of view, we are in the transition from xcodebuild to bazel and I am used to only have debug flags for Xcode Debug's configuration, for me seemed strange that this flag is used in fastbuild that's why I open up the pr/discussion.
The pr 428 is a better solution we will have more control over the flags if we don't want the default ones.

Thank you Keith, Im closing this pr then!

keith:
FWIW the difference between dbg and fastbuild is pretty minimal to so you may opt to just never use fastbuild

PR #550, opened 2021-01-15
Allow modules propagated by `SwiftInfo` to indicate whether they are system modules.

keith:
System modules differ from non-system modules in that we don't pass the module map to the compiler via `-fmodule-map-file` for system modules in implicit module builds, only in explicit module builds. This is because module maps passed via `-fmodule-map-file` are always treated as non-system modules, which affects how diagnostics are emitted when their headers are parsed, and some Swift modules like `SwiftOverlayShims` don't declare themselves as `[system]` in their module map file but contain small nullability issues. Such headers *must* be found through a standard header search via a system path like `-isystem` so that they don't cause builds to fail if they treat warnings more severely.

PiperOrigin-RevId: 351896080
(cherry picked from commit 214061922f68fbf4f0f7fcf4df9d54af5b279cf5)

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -935,7 +935,7 @@ def _collect_clang_module_inputs(
         # Add the module map, which we use for both implicit and explicit module
         # builds. For implicit module builds, we don't worry about the headers
         # above because we collect the full transitive header set below.
-        if not types.is_string(module_map):
+        if not module.is_system and not types.is_string(module_map):
             module_inputs.append(module_map)
 
     # If we prefer textual module maps and headers for the build, fall back to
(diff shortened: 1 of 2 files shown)


swift/internal/compiling.bzl
```diff
@@ -949,33 +949,36 @@ def _collect_clang_module_inputs(
...
             `swift_common.create_module`.
+        ignore_system: If `True` and the module is a system module, no flag
+            should be returned. Defaults to `True`.
 
     Returns:
-        The argument to pass to `swiftc` (without the `-Xcc` prefix).
+        A list of arguments, possibly empty, to pass to `swiftc` (without the
+        `-Xcc` prefix).
     """
-    if types.is_string(module):
```
keith, line 962: @thii can you look at this? I'm reverting a bit of https://github.com/bazelbuild/rules_swift//commit/efc75e0cb80186f5c774cb4fc87f2aeba9addad8 here but I don't think this function can get a string anymore? 

Related https://github.com/bazelbuild/rules_swift//commit/d916037ac3a918ce7231c355945246d265cb78a3

thii: LGTM. efc75e0cb80186f5c774cb4fc87f2aeba9addad8 is not needed anymore. I thought the whole thing already got replaced by d916037ac3a918ce7231c355945246d265cb78a3.

keith: great that's what I was hoping, thanks!

PR #555, opened 2021-01-27
Use released protoc binaries where possible

thii:
This updates `rules_proto`, which has been configured to use released
protoc binaries where it can.

This can be a breaking change if users have a call to `protobuf_deps()`
_before_ `swift_rules_extra_dependencies()` in their WORKSPACE. To
fix, remove `load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")`
and `protobuf_deps()` from the WORKSPACE file, and use the following
instead:

```
load(
    "@build_bazel_rules_swift//swift:extras.bzl",
    "swift_rules_extra_dependencies",
)

swift_rules_extra_dependencies()
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -8,18 +8,11 @@ load(
 swift_rules_dependencies()
 
 load(
-    "@build_bazel_apple_support//lib:repositories.bzl",
-    "apple_support_dependencies",
+    "@build_bazel_rules_swift//swift:extras.bzl",
+    "swift_rules_extra_dependencies",
 )
 
-apple_support_dependencies()
-
-load(
-    "@com_google_protobuf//:protobuf_deps.bzl",
-    "protobuf_deps",
-)
-
-protobuf_deps()
+swift_rules_extra_dependencies()
 
 load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
--- a/swift/repositories.bzl
+++ b/swift/repositories.bzl
@@ -99,23 +99,12 @@ def swift_rules_dependencies():
     _maybe(
         http_archive,
         name = "rules_proto",
-        # latest as of 2020-09-01
+        # latest as of 2021-01-27
         urls = [
-            "https://github.com/bazelbuild/rules_proto/archive/40298556293ae502c66579620a7ce867d5f57311.zip",
+            "https://github.com/bazelbuild/rules_proto/archive/a0761ed101b939e19d83b2da5f59034bffc19c12.zip",
         ],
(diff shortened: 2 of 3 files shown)


swift/repositories.bzl
```diff
@@ -38,6 +38,18 @@ def swift_rules_dependencies():
     dependencies of the Swift rules are downloaded and that they are isolated
     from changes to those dependencies.
     """
+    _maybe(
```
brentleyjones, line 41: Does this need to be moved up here, versus where it was? The list will be alphabetical if it's back where it started.

thii: It doesn't seem needed to be moved up. I got some build error at first that made me thought it was because of the order.

brentleyjones (approved):

WORKSPACE
```diff
@@ -8,18 +8,18 @@ load(
...
 
 load(
-    "@com_google_protobuf//:protobuf_deps.bzl",
-    "protobuf_deps",
+    "@build_bazel_apple_support//lib:repositories.bzl",
+    "apple_support_dependencies",
 )
 
-protobuf_deps()
+apple_support_dependencies()
```
keith, line 22: im not sure what i was doing here originally but can't we remove this since it's in the extra deps?

swift/repositories.bzl
```diff
@@ -99,23 +99,12 @@ def swift_rules_dependencies():
...
+            "https://github.com/bazelbuild/rules_proto/archive/a0761ed101b939e19d83b2da5f59034bffc19c12.zip",
         ],
-        sha256 = "37d32b789be90fead9ab108dbe4fe4df463d26c122dc896dc1bf134252d3c49a",
-        strip_prefix = "rules_proto-40298556293ae502c66579620a7ce867d5f57311",
-        type = "zip",
-    )
-
-    _maybe(
-        http_archive,
-        name = "com_google_protobuf",
```
keith, line 113: maybe a bit weird that this is still used but not directly depended on? idk.

segiddins (approved):
Awesome!

jerrymarino (approved):
This is a nice change, I often to get rebuilds of the proto toolchain when using misc bazel repos. +1 to removing the `"@com_google_protobuf//:protobuf_deps.bzl"` if we can 👍

swift/repositories.bzl
thii: It's still a dependency. rules_proto initializes and replaces part of it by prebuilt binaries.

WORKSPACE
thii: Removed.

thii:
Note that this is just for the protoc executables. The swift worker depends on the `protobuf` cc_library, so it still needs to be compiled.

PR #560, opened 2021-01-28
Simplify the paths to module maps generated by the Swift build rules.

keith:
Since we pass these module maps directly to the compiler with `-fmodule-map-file`, we don't need them to be named exactly `module.modulemap`, so the extra subdirectory is also unnecessary.

This change also migrates generation of the modulemap file content for the generated header's module to the same code (`module_maps.bzl`) already used by `swift_clang_module_aspect`, instead of having its own separate code path.

PiperOrigin-RevId: 354324121
(cherry picked from commit 5f51ca9c5149122f41cada6122c61788d880fee9)

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -62,6 +62,7 @@ load(
     "SWIFT_FEATURE_VFSOVERLAY",
 )
 load(":features.bzl", "are_all_features_enabled", "is_feature_enabled")
+load(":module_maps.bzl", "write_module_map")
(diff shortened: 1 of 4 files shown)


brentleyjones (approved):

thii:
The generated module map now reexports all of the imported modules. Is that an intended change?

keith:
@allevato

allevato:
Oof, it was not, thanks for catching that. I can get a fix out soon if you want to wait and stay in sync with `upstream`.

PR #562, opened 2021-02-08
Allow controlling whether to include `export *` in the generated module maps

thii:
The content of the generated module maps was changed in
aa542252a2ea9c11a77b45de314350838d554747, by always adding `export *`
which wasn't previously included. This change reverts that back, and
adds a new parameter - `reexports_all` - to the `write_module_map`
helper, that allows controlling whether it should include `export *` in
the generated module map.


keith:
sounds like this was an unintentional regression https://github.com/bazelbuild/rules_swift/pull/560#issuecomment-775309878

PR #566, opened 2021-02-16
Explicitly set --macos_minimum_os=10.15 to make `swift_{binary,test}` runnable on macOS Catalina (which is the macOS version used by Bazel CI) when building with Xcode 12 or above

thii:
Since there's no way to set the deployment version for
`*_{binary,test}`, this makes rules_swift on CI green again without
having to go back to using Xcode 11.7.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,1 @@
+build --macos_minimum_os=10.15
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

philwo:
We are considering eventually upgrading to macOS Big Sur - would that be an issue then?

thii:
It shouldn't be. The value just needs to be lower or equal to the running OS version.

segiddins (approved):

Issue #561, opened 2021-02-03
Performance issues / useage of `swift.index_while_building`


jerrymarino:
Good morning! I'm again looking at how to use the feature `swift.index_while_building` with Bazel and found it to introduce significant build performance issues. The impact of using this flag is a 300% increase in swift compilation times as opposed to not using it.

From what I can tell, it adds multiple GB of output to the build - as there is a per `swift_library` index, all system modules and deps seem to be indexed O(N swift_library) times. rules_ios uses the program `index-import` which churns trying to import these indicies for [longer than it'd takes](https://github.com/bazel-ios/rules_ios/issues/203) Xcode to index the source code after build. Using a "global index" - a single index mitigates the perf impact, but I'm not sure how this would work with remote caching.

I was curious how "index while building" is working with `rules_swift` and Bazel in general for larger builds. XCHammer doesn't have this feature and works around it by letting Xcode index a focused subset after the build.

brentleyjones:
At Target we used a global index, but only when running locally: https://github.com/bazelbuild/rules_swift/commit/6b77a6bd416586947e6e1b8cd600eb59dd772dfe.

Remote cache hits would still download to each module folder, but post build we would copy those into the global index. Locally built modules would go directly into the global index. we then only `index-import` from the global index, also also importing units and records that changed: https://github.com/lyft/index-import/pull/46/commits/6bb07ce98989a8827cb55385d8e2b879f4667ec9

We personally found that the largest hit was the vast number of files and directories being made (and re-made, since `declare_directory` would delete the indexstore before re-populating it). I wrote up about it here: https://brentley.dev/xcodebuild-vs-bazel-incremental/#multiple-index-stores

---

I would love to see a first class solution for a global index store, but I think it would require changes to Bazel: https://github.com/bazelbuild/bazel/issues/7527

jerrymarino:
@brentleyjones great to hear that you've got this working! Also thanks for taking the time to write up how you're dealing with this, I'm going to give your blog post a read 👍 

> Remote cache hits would still download to each module folder, but post build we would copy those into the global index.

[...]

PR #567, opened 2021-02-20
Implement global index store cache

jerrymarino:
swift and clang write index data based on the status of what resides in
`index-store-path`. When using a transient per-swift-library index, it
writes O( M imports * N libs) indexer data and slows down compilation
significantly: to the tune of 300% slow down and 6GB+ index data in my
testing. Bazel likes to use per `swift_library` data in order to remote
cache them, which needs extra consideration to work with the design of
swift and clang and preserve the performance

This PR does 2 things to make index while building use the original
model of the index-store so we don't have to patch clang and swift for
this yet. When the flag `-Xwrapped-swift-enable-global-index-store` is
set:

1. it writes to a global index cache, `bazel-out/global_index_store`
2. it copies indexstore data (records and units) to where Bazel
   expected them for caching: e.g. the original location with
   `swift.index_while_building`

Note that while this uses a `index-import` binary, it is very different
than how index-import currently works. In the variant this expects, it
is program that can _copy_ index data for specific compiler outputs.
This has the effect fo remote caching the subset that was used in
compilation for this library.

I added a detailed description of this feature in the RFC to add it
there https://github.com/lyft/index-import/pull/53.  In practice,
transitive imports will be indexed by deps and if they aren't cached
then the fallback is Xcode.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -58,6 +58,7 @@ load(
     "SWIFT_FEATURE_SUPPORTS_SYSTEM_MODULE_FLAG",
     "SWIFT_FEATURE_SYSTEM_MODULE",
     "SWIFT_FEATURE_USE_C_MODULES",
+    "SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE",
     "SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE",
     "SWIFT_FEATURE_VFSOVERLAY",
     "SWIFT_FEATURE__NUM_THREADS_0_IN_SWIFTCOPTS",
@@ -805,6 +806,14 @@ def compile_action_configs(
             ],
             features = [SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES],
         ),
+        swift_toolchain_config.action_config(
+            actions = [swift_action_names.COMPILE],
+            configurators = [_global_index_store_configurator],
+            features = [
+                SWIFT_FEATURE_INDEX_WHILE_BUILDING,
+                SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE,
+            ],
+        ),
 
         # Configure index-while-building.
         swift_toolchain_config.action_config(
@@ -1418,6 +1427,12 @@ def _index_while_building_configurator(prerequisites, args):
     """Adds flags for index-store generation to the command line."""
     if not _index_store_path_overridden(prerequisites.user_compile_flags):
         args.add("-index-store-path", prerequisites.indexstore_directory.path)
+
+def _global_index_store_configurator(prerequisites, args):
+    """Adds flags for index-store generation to the command line."""
+    out_dir = prerequisites.indexstore_directory.dirname.split("/")[0]
+    path = out_dir + "/_global_index_store"
+    args.add("-Xwrapped-swift=-global-index-store-import-path=" + path)
 
 def _conditional_compilation_flag_configurator(prerequisites, args):
     """Adds (non-Clang) conditional compilation flags to the command line."""
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -99,10 +99,14 @@ SWIFT_FEATURE_ENABLE_TESTING = "swift.enable_testing"
 SWIFT_FEATURE_FULL_DEBUG_INFO = "swift.full_debug_info"
 
 # If enabled, the compilation action for a target will produce an index store.
+# https://docs.google.com/document/d/1cH2sTpgSnJZCkZtJl1aY-rzy4uGPcrI-6RrUpdATO2Q/
 SWIFT_FEATURE_INDEX_WHILE_BUILDING = "swift.index_while_building"
 
 # If enabled the compilation action will not produce indexes for system modules.
 SWIFT_FEATURE_DISABLE_SYSTEM_INDEX = "swift.disable_system_index"
+
+# Index while building - using a global index store cache
+SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE = "swift.use_global_index_store"
 
 # If enabled, compilation actions and module map generation will assume that the
 # header paths in module maps are relative to the current working directory
--- a/swift/repositories.bzl
+++ b/swift/repositories.bzl
@@ -108,6 +108,27 @@ def swift_rules_dependencies():
         type = "zip",
     )
 
+    # It relies on `index-import` to import indexes into Bazel's remote
+    # cache and allow using a global index internally in workers.
+    # Note: this is only loaded if swift.index_while_building_v2 is enabled
+    _maybe(
+        http_archive,
+        name = "build_bazel_rules_swift_index_import",
+        build_file_content = """\
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
+
+native_binary(
+    name = "index_import",
+    src = "index-import",
+    out = "index-import",
+    visibility = ["//visibility:public"],
+)
+""",
+        canonical_id = "index-import-5.3.2.6",
+        urls = ["https://github.com/MobileNativeFoundation/index-import/releases/download/5.3.2.6/index-import.zip"],
+        sha256 = "61a58363f56c5fd84d4ebebe0d9b5dd90c74ae170405a7b9018e8cf698e679de",
+    )
+
     _maybe(
         swift_autoconfiguration,
         name = "build_bazel_rules_swift_local_config",
--- a/tools/worker/BUILD
+++ b/tools/worker/BUILD
@@ -18,12 +18,18 @@ config_setting(
     },
 )
 
+# Internal hinge for index while building V2 feature
+config_setting(
+    name = "use_global_index_store",
+    values = {
+        "features": "swift.use_global_index_store",
+    },
+)
+
 cc_library(
     name = "compile_with_worker",
     srcs = [
         "compile_with_worker.cc",
-        "output_file_map.cc",
-        "output_file_map.h",
         "work_processor.cc",
         "work_processor.h",
     ],
@@ -34,7 +40,6 @@ cc_library(
         "//tools/common:file_system",
         "//tools/common:path_utils",
         "//tools/common:temp_file",
-        "@com_github_nlohmann_json//:json",
         "@com_google_protobuf//:protobuf",
     ],
 )
@@ -50,13 +55,30 @@ cc_library(
 
 cc_library(
     name = "swift_runner",
-    srcs = ["swift_runner.cc"],
+    srcs = [
+        "output_file_map.cc",
+        "output_file_map.h",
+        "swift_runner.cc",
+    ],
     hdrs = ["swift_runner.h"],
+    copts = select({
+        ":use_global_index_store": [
+            "-DINDEX_IMPORT_PATH=\\\"$(rootpath @build_bazel_rules_swift_index_import//:index_import)\\\"",
+        ],
+        "//conditions:default": [],
+    }),
+    data = select({
+        ":use_global_index_store": [
+            "@build_bazel_rules_swift_index_import//:index_import",
+        ],
+        "//conditions:default": [],
+    }),
     deps = [
         "//tools/common:bazel_substitutions",
         "//tools/common:file_system",
         "//tools/common:process",
         "//tools/common:temp_file",
+        "@com_github_nlohmann_json//:json",
     ],
 )
 
--- a/tools/worker/swift_runner.cc
+++ b/tools/worker/swift_runner.cc
@@ -20,6 +20,7 @@
 #include "tools/common/file_system.h"
 #include "tools/common/process.h"
 #include "tools/common/temp_file.h"
+#include "tools/worker/output_file_map.h"
 
 namespace {
 
@@ -129,13 +130,87 @@ int SwiftRunner::Run(std::ostream *stderr_stream, bool stdout_to_stderr) {
     exit_code = RunSubProcess(rewriter_args, stderr_stream, stdout_to_stderr);
   }
 
+  auto enable_global_index_store = global_index_store_import_path_ != "";
+  if (enable_global_index_store) {
+    OutputFileMap output_file_map;
+    output_file_map.ReadFromPath(output_file_map_path_);
+
+    auto outputs = output_file_map.incremental_outputs();
+    std::map<std::string, std::string>::iterator it;
+
+    std::vector<std::string> ii_args;
+// The index-import runfile path is passed as a define
+#if defined(INDEX_IMPORT_PATH)
+    ii_args.push_back(INDEX_IMPORT_PATH);
+#else
+    // Logical error
+    std::cerr << "Incorrectly compiled work_processor.cc";
+    exit_code = EXIT_FAILURE;
+    return exit_code;
+#endif
+
+    for (it = outputs.begin(); it != outputs.end(); it++) {
+      // Need the actual output paths of the compiler - not bazel
+      auto output_path = it->first;
+      auto file_type = output_path.substr(output_path.find_last_of(".") + 1);
+      if (file_type == "o") {
+        ii_args.push_back("-import-output-file");
+        ii_args.push_back(output_path);
+      }
+    }
+
+    auto exec_root = GetCurrentDirectory();
+    // Copy back from the global index store to bazel's index store
+    ii_args.push_back(exec_root + "/" + global_index_store_import_path_);
+    ii_args.push_back(exec_root + "/" + index_store_path_);
+    exit_code =
+        RunSubProcess(ii_args, stderr_stream, /*stdout_to_stderr=*/true);
+  }
   return exit_code;
 }
+
+// Marker for end of iteration
+class StreamIteratorEnd {};
+
+// Basic iterator over an ifstream
+class StreamIterator {
+ public:
+  StreamIterator(std::ifstream &file) : file_{file} { next(); }
+
+  const std::string &operator*() const { return str_; }
+
+  StreamIterator &operator++() {
+    next();
+    return *this;
+  }
+
+  bool operator!=(StreamIteratorEnd) const { return !!file_; }
+
+ private:
+  void next() { std::getline(file_, str_); }
+
+  std::ifstream &file_;
+  std::string str_;
+};
+
+class ArgsFile {
+ public:
+  ArgsFile(std::ifstream &file) : file_(file) {}
+
+  StreamIterator begin() { return StreamIterator{file_}; }
+
+  StreamIteratorEnd end() { return StreamIteratorEnd{}; }
+
+ private:
+  std::ifstream &file_;
+};
 
 bool SwiftRunner::ProcessPossibleResponseFile(
     const std::string &arg, std::function<void(const std::string &)> consumer) {
   auto path = arg.substr(1);
   std::ifstream original_file(path);
+  ArgsFile args_file(original_file);
+
   // If we couldn't open it, maybe it's not a file; maybe it's just some other
   // argument that starts with "@" such as "@loader_path/..."
   if (!original_file.good()) {
@@ -143,15 +218,17 @@ bool SwiftRunner::ProcessPossibleResponseFile(
     return false;
   }
 
+  // Read the file to a vector to prevent double I/O
+  auto args = ParseArguments(args_file);
+
   // If we're forcing response files, process and send the arguments from this
   // file directly to the consumer; they'll all get written to the same response
   // file at the end of processing all the arguments.
   if (force_response_file_) {
-    std::string arg_from_file;
-    while (std::getline(original_file, arg_from_file)) {
+    for (auto it = args.begin(); it != args.end(); ++it) {
       // Arguments in response files might be quoted/escaped, so we need to
       // unescape them ourselves.
-      ProcessArgument(Unescape(arg_from_file), consumer);
+      ProcessArgument(it, Unescape(*it), consumer);
     }
     return true;
   }
@@ -161,12 +238,10 @@ bool SwiftRunner::ProcessPossibleResponseFile(
   bool changed = false;
   std::string arg_from_file;
   std::vector<std::string> new_args;
-
-  while (std::getline(original_file, arg_from_file)) {
-    changed |=
-        ProcessArgument(arg_from_file, [&](const std::string &processed_arg) {
-          new_args.push_back(processed_arg);
-        });
+  for (auto it = args.begin(); it != args.end(); ++it) {
+    changed |= ProcessArgument(it, *it, [&](const std::string &processed_arg) {
+      new_args.push_back(processed_arg);
+    });
   }
 
   if (changed) {
@@ -182,10 +257,11 @@ bool SwiftRunner::ProcessPossibleResponseFile(
   return changed;
 }
 
+template <typename Iterator>
 bool SwiftRunner::ProcessArgument(
-    const std::string &arg, std::function<void(const std::string &)> consumer) {
+    Iterator &itr, const std::string &arg,
+    std::function<void(const std::string &)> consumer) {
   bool changed = false;
-
   if (arg[0] == '@') {
     changed = ProcessPossibleResponseFile(arg, consumer);
   } else {
@@ -198,8 +274,8 @@ bool SwiftRunner::ProcessArgument(
         consumer(GetCurrentDirectory() + "=.");
         changed = true;
       } else if (new_arg == "-coverage-prefix-pwd-is-dot") {
-        // Get the actual current working directory (the workspace root), which we
-        // didn't know at analysis time.
+        // Get the actual current working directory (the workspace root), which
+        // we didn't know at analysis time.
         consumer("-coverage-prefix-map");
         consumer(GetCurrentDirectory() + "=.");
         changed = true;
@@ -213,7 +289,8 @@ bool SwiftRunner::ProcessArgument(
         temp_directories_.push_back(std::move(module_cache_dir));
         changed = true;
       } else if (StripPrefix("-generated-header-rewriter=", new_arg)) {
-        generated_header_rewriter_path_ = new_arg;
+        changed = true;
+      } else if (StripPrefix("-global-index-store-import-path=", new_arg)) {
         changed = true;
       } else {
         // TODO(allevato): Report that an unknown wrapper arg was found and give
@@ -221,6 +298,28 @@ bool SwiftRunner::ProcessArgument(
         changed = true;
       }
     } else {
+      // Process default arguments
+      if (arg == "-index-store-path") {
+        consumer("-index-store-path");
+        ++itr;
+
+        // If there was a global index store set, pass that to swiftc.
+        // Otherwise, pass the users. We later copy index data onto the users.
+        if (global_index_store_import_path_ != "") {
+          new_arg = global_index_store_import_path_;
+        } else {
+          new_arg = index_store_path_;
+        }
+        changed = true;
+      } else if (arg == "-output-file-map") {
+        // Save the output file map to the value proceeding
+        // `-output-file-map`
+        consumer("-output-file-map");
+        ++itr;
+        new_arg = output_file_map_path_;
+        changed = true;
+      }
+
       // Apply any other text substitutions needed in the argument (i.e., for
       // Apple toolchains).
       //
@@ -236,6 +335,36 @@ bool SwiftRunner::ProcessArgument(
   return changed;
 }
 
+template <typename Iterator>
+std::vector<std::string> SwiftRunner::ParseArguments(Iterator itr) {
+  std::vector<std::string> out_args;
+  for (auto it = itr.begin(); it != itr.end(); ++it) {
+    auto arg = *it;
+    out_args.push_back(arg);
+
+    if (StripPrefix("-Xwrapped-swift=", arg)) {
+      if (StripPrefix("-global-index-store-import-path=", arg)) {
+        global_index_store_import_path_ = arg;
+      } else if (StripPrefix("-generated-header-rewriter=", arg)) {
+        generated_header_rewriter_path_ = arg;
+      }
+    } else {
+      if (arg == "-output-file-map") {
+        ++it;
+        arg = *it;
+        output_file_map_path_ = arg;
+        out_args.push_back(arg);
+      } else if (arg == "-index-store-path") {
+        ++it;
+        arg = *it;
+        index_store_path_ = arg;
+        out_args.push_back(arg);
+      }
+    }
+  }
+  return out_args;
+}
+
 std::vector<std::string> SwiftRunner::ProcessArguments(
     const std::vector<std::string> &args) {
   std::vector<std::string> new_args;
@@ -248,16 +377,19 @@ std::vector<std::string> SwiftRunner::ProcessArguments(
 #endif
 
   // The tool is assumed to be the first argument. Push it directly.
-  auto it = args.begin();
+  auto parsed_args = ParseArguments(args);
+
+  auto it = parsed_args.begin();
   new_args.push_back(*it++);
 
   // If we're forcing response files, push the remaining processed args onto a
   // different vector that we write out below. If not, push them directly onto
   // the vector being returned.
   auto &args_destination = force_response_file_ ? response_file_args : new_args;
-  while (it != args.end()) {
-    ProcessArgument(
-        *it, [&](const std::string &arg) { args_destination.push_back(arg); });
+  while (it != parsed_args.end()) {
+    ProcessArgument(it, *it, [&](const std::string &arg) {
+      args_destination.push_back(arg);
+    });
     ++it;
   }
 
--- a/tools/worker/swift_runner.h
+++ b/tools/worker/swift_runner.h
@@ -97,8 +97,14 @@ class SwiftRunner {
   //
   // This method has file system side effects, creating temporary files and
   // directories as needed for a particular substitution.
-  bool ProcessArgument(const std::string &arg,
+  template <typename Iterator>
+  bool ProcessArgument(Iterator &itr, const std::string &arg,
                        std::function<void(const std::string &)> consumer);
+
+  // Parses arguments to ivars and returns a vector of strings from the iterator.
+  // This method doesn't actually mutate any of the arguments.
+  template <typename Iterator>
+  std::vector<std::string> ParseArguments(Iterator itr);
 
   // Applies substitutions to the given command line arguments, returning the
   // results in a new vector.
@@ -128,6 +134,18 @@ class SwiftRunner {
   // The path to the generated header rewriter tool, if one is being used for
   // this compilation.
   std::string generated_header_rewriter_path_;
+
+  // The path of the output map file
+  std::string output_file_map_path_;
+
+  // The index store path argument passed to the runner
+  std::string index_store_path_;
+
+  // The path of the global index store  when using
+  // swift.use_global_index_store. When set, this is passed to `swiftc` as the
+  // `-index-store-path`. After running `swiftc` `index-import` copies relevant
+  // index outputs into the `index_store_path` to integrate outputs with Bazel.
+  std::string global_index_store_import_path_;
 };
 
 #endif  // BUILD_BAZEL_RULES_SWIFT_TOOLS_WORKER_SWIFT_RUNNER_H_


jerrymarino:
Fixes bazelbuild/rules_swift#561

tools/worker/work_processor.cc
```diff
@@ -129,7 +162,53 @@ void WorkProcessor::ProcessWorkRequest(
...
+  // Import global indexes if compilation succeeds
+  if (swift_exit_code != EXIT_FAILURE && !bazel_index_store_path.empty()) {
+    std::map<std::string, std::string>::iterator it;
+
+    std::vector<std::string> ii_args;
+
+    // Perhaps add this as an external dep of rules_swift
+    // needs the feature -import-output-file added here
+    // https://github.com/lyft/index-import/pull/53
+    auto index_import_path = "/usr/local/bin/index-import";
```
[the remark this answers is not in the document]
jerrymarino: Yeah I totally agree @segiddins 👍  - I mainly hardcoded this. I wasn't sure if the tool would actually be part of index-import in the end, but will add that once I've got this piece figured out.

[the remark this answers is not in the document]
brentleyjones: It would be nice if it didn't depend directly on `index-import`, but some `executable`. That interface to that executable could be `index-import`, or it could be implemented by something else if the users wanted to.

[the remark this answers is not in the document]
jerrymarino: Yep I like the idea of just having a toolchain input that we can specify: perhaps by default it'd pull in `index-import` as a sensible default if `swift.index_while_building` was enabled

tools/worker/work_processor.cc
```diff
@@ -77,16 +83,41 @@ void WorkProcessor::ProcessWorkRequest(
...
+
     if (!arg.empty()) {
       params_file_stream << arg << '\n';
     }
 
     prev_arg = original_arg;
   }
 
+  // Material within this index store is namespaced to arcitecture, os, etc
+  auto exec_root = GetCurrentDirectory();
```
amberdixon, line 106: This makes sense. But what if instead of taking this approach, we changed the return value of the `_indexstore_directory` method in derived_files.bzl ? Would that not have the same affect?

tools/worker/work_processor.cc
```diff
@@ -77,16 +83,42 @@ void WorkProcessor::ProcessWorkRequest(
...
+
     if (!arg.empty()) {
       params_file_stream << arg << '\n';
     }
 
     prev_arg = original_arg;
   }
 
+  // Material within this index store is namespaced to arcitecture, OS
+  // version, and material // and there isn't a per architecture index.
```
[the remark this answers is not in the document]
jerrymarino: Done 👍

tools/worker/work_processor.cc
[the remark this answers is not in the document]
jerrymarino: Added on latest rev 👍

tools/worker/work_processor.cc
```diff
@@ -77,16 +83,42 @@ void WorkProcessor::ProcessWorkRequest(
       is_wmo = true;
     }
 
+    /// Peel off index-store-path, we will conditionally pop on a new one
+    if (arg == "-index-store-path") {
+      arg.clear();
+    } else if (prev_arg == "-index-store-path") {
+      bazel_index_store_path = arg;
```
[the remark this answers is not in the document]
jerrymarino: No, it flips the bit when `-Xwrapped-swift-enable-global-index-store` - added a comment for clarity

tools/worker/work_processor.cc
jerrymarino: I don't think so - the main reason we run the "index import" binary in the worker is to populate `bazel-out` for remote caching. That material needs to be known to Bazel rather than the global cache which clang and swift use internally. For now, I added more comments to clarify how that works. Also I'm planning to send a consolidated document of the new architecture of "Index while building V2"

swift/repositories.bzl
```diff
@@ -124,3 +124,24 @@ def swift_rules_dependencies():
         swift_autoconfiguration,
         name = "build_bazel_rules_swift_local_config",
     )
+
+    # index_import is used as a swiss army knife of managing llvm indexes and
+    # is pulled in when index_while_building_v2 is enabled
+    _maybe(
+        http_archive,
+        name = "build_bazel_rules_swift_index_import",
```
segiddins, line 120: any reason this wouldn't be com_github_lyft_index_import?

jerrymarino:
In the latest rev I have added the ability to turn in the feature when `swift.index_while_building_v2` is enabled - this will have 0 impact on existing code and makes it easier to test out! 

It won't pull index-import unless that feature is set, and the feature can live adjacent to `swift.index_while_building` until we deprecate /remove the original implementation.

tools/worker/work_processor.cc
```diff
@@ -77,16 +86,49 @@ void WorkProcessor::ProcessWorkRequest(
       is_wmo = true;
     }
 
+    // The output processor is passed as an argument
+    if (arg == "-Xwrapped-swift-output-processor") {
+      arg.clear();
+    } else if (prev_arg == "-Xwrapped-swift-output-processor") {
```
segiddins, line 92: this feels specific to processing index stores, rather than generic output processing, should the name reflect that?

swift/internal/actions.bzl
```diff
@@ -218,6 +219,19 @@ def run_toolchain_action(
         ):
             execution_requirements["supports-workers"] = "1"
 
+        # For index_while_building_v2 we import a subset of the indexed outputs
+        # from the global index store cache into bazel-out. The program that
+        # handles this is an external program
+        if is_feature_enabled(
+            feature_configuration = feature_configuration,
+            feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING_V2,
+        ) and len(swift_toolchain.swift_worker_output_processor[DefaultInfo].files.to_list()):
```
segiddins, line 228: should this use the `executable` instead of `files`? additionally, should we be adding runfiles in anywhere in case the executable isn't a single file?

jerrymarino: I don't know if a way to make `executable` have a conditional dep on a select so it uses a file group. It is only a file group because of that.

swift/repositories.bzl
jerrymarino: Yep - deps are namespaced to `rules_swift` so it can use it's own version

tools/worker/work_processor.cc
jerrymarino: I don't think so - at a technical level it's just a binary that handles `-import-output-file` - the program could do virtually anything

swift/repositories.bzl
```diff
@@ -112,3 +112,24 @@ def swift_rules_dependencies():
...
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
+
+native_binary(
+    name = "index_import",
+    src = "index-import",
+    out = "index-import",
+    visibility = ["//visibility:public"],
+)
+""",
+        # TODO[jmarino]: update once https://github.com/lyft/index-import/pull/53 lands or move to other repo
```
jerrymarino, line 131: @brentleyjones  or @keith  LMK what you all think of this, if we want to go with the "index import" becomes a swiss army knife to manage indexes it will make sense to put this capability in the repo. I plan to end up with a mono-repo containing all code to manage indexes in Bazel so people can easily add it!

jerrymarino: As a heads up, I'll need to determine a repo to put the code that imports remote indexes into a global index: likely an aspect or BEP based program. I liked using `index-import` because it's not specific to anyones rules and is general enough

brentleyjones: Hmm. I was thinking that `swift_library` would have an `index_importer` attr, so we wouldn't have an explicit dependency on `index-import`. And that would be an `executable` type. Users could use a script or `index-import`, but on their terms. That attribute would default to `None` and you would get an error in the rule if it was unset when using the `swift.index_while_building_v2` feature.

jerrymarino: I like encapsulating all internals of this feature and implementing the feature one way to work for all users of `index_while_building_v2` feature. As a user of Bazel, I'd like to flip a feature and have it do something reasonable that works with remote caching out of the box without worrying about the intricacies and footguns. That being said, if you'd like to add a different program with the current rev you can override `build_bazel_rules_swift_index_import` by declaring it before hand

tools/worker/work_processor.cc
amberdixon: I think an alternative here would be to:
1. Add a method to `derived_files.bzl` called `_global_indexstore_directory` which points to the global indexstore.
2. In the `_index_while_building_configurator` method in compiling.bzl, return the global indexstore dir, if this feature is enabled.
3. In `swift_library.bzl`, after callign `swift_common.compile`, add something like this:
```
if compilation_outputs.global_indexstore && index_while_building_v2:
   ctx.actions.run("mv #{compilation_outputs.global_indexstore}  #{compilation_outputs.indexstore}
```

amberdixon: The tricky part here is to get the `mv` command right. You would need to use `compilation_outputs.output_file_map` contents as well to figure out which files get moved out of the global indexstore and into the module-specific indexstore.

jerrymarino: I don't think you can give the action an unhermetic out of band output like the global index or interface with the remote cache with that approach. I added the higher level design here: https://github.com/bazel-ios/rules_ios/pull/212/files

jerrymarino: Also if you used `mv` then it'd put the global index store: e.g. ~500MB of data into each `swift_library`'s remote cache entry. The current codepath reads unit files with LLVM to pull a very small part of the index

jerrymarino: Also if you did:
```
if compilation_outputs.global_indexstore && index_while_building_v2:
   ctx.actions.run("mv #{compilation_outputs.global_indexstore}  #{compilation_outputs.indexstore}
```

wouldn't it have a conflict when other actions are running: either corrupt `compilation_outputs.global_indexstore` or delete it altogheter?

amberdixon: I think you would still need a binary that reads output-file-map and moves files accordingly and cannot just use `mv`. 

Isn't the existing implementation in worker.cc doing the same thing as what I had proposed above? It's writing files to a global index and then moving them over to somewhere else. How is the existing impl avoiding conflicts and behaving non-hermetically, since all compile commands are writing to the same global indexstore?

swift/internal/actions.bzl
segiddins: instead of using a filegroup, we should allow for the `swift_worker_output_processor` in the toolchain to be `None`

swift/internal/compiling.bzl
```diff
@@ -1837,6 +1847,9 @@ def _declare_compile_outputs(
     index_while_building = is_feature_enabled(
         feature_configuration = feature_configuration,
         feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING,
+    ) or is_feature_enabled(
```
segiddins, line 1850: should we error if both are enabled? should we require the non-v2 feature to be enabled if the v2 one is?

swift/internal/feature_names.bzl
```diff
@@ -99,6 +99,10 @@ SWIFT_FEATURE_ENABLE_TESTING = "swift.enable_testing"
 # warnings otherwise.
 SWIFT_FEATURE_FULL_DEBUG_INFO = "swift.full_debug_info"
 
+# Index while building - using a global index store cache
```
segiddins, line 102: other features are documented in terms of `If enabled...` and we should replicate that style here

swift/internal/swift_toolchain.bzl
```diff
@@ -273,6 +274,14 @@ for incremental compilation using a persistent mode.
 """,
                 executable = True,
             ),
+            "_worker_output_processor": attr.label(
```
segiddins, line 277: `executable = True`

swift/internal/swift_toolchain.bzl
```diff
@@ -215,6 +215,7 @@ def _swift_toolchain_impl(ctx):
             root_dir = toolchain_root,
             supports_objc_interop = False,
             swift_worker = ctx.executable._worker,
+            swift_worker_output_processor = ctx.attr._worker_output_processor,
```
segiddins, line 218: `ctx.executable._worker` ?

tools/worker/BUILD
```diff
@@ -77,6 +77,29 @@ cc_binary(
...
+
+
+filegroup(
+   name = "worker_output_processor",
+   # with index_while_building_v2 we want to add a dependency on the tool that
+   # manages the per swift_library index. Otherwise, don't load it
+   # This is mainly an internal capability right now - check
+   # tools/worker/work_processor.cc for usage:
+   # build_bazel_rules_swift_index_import is be provided by the user for now
+   srcs = select({
```
segiddins, line 96: instead of a filegroup, could use `alias` and select on `actual`, returning either index-import or `None`

tools/worker/work_processor.cc
segiddins: except here we're only giving it .o  files, and not any of the other files output by swiftc

tools/worker/work_processor.cc
```diff
@@ -77,16 +86,49 @@ void WorkProcessor::ProcessWorkRequest(
...
     if (!arg.empty()) {
       params_file_stream << arg << '\n';
     }
 
     prev_arg = original_arg;
   }
 
+  // Material within this index store is namespaced to arcitecture, os, etc
+  auto exec_root = GetCurrentDirectory();
+  auto global_index_store = exec_root + "/bazel-out/global_index_store";
```
segiddins, line 119: should this be namespaced under rules_swift somewhere?

tools/worker/BUILD
```diff
@@ -77,6 +77,29 @@ cc_binary(
...
+)
+
+
+filegroup(
+   name = "worker_output_processor",
+   # with index_while_building_v2 we want to add a dependency on the tool that
+   # manages the per swift_library index. Otherwise, don't load it
+   # This is mainly an internal capability right now - check
+   # tools/worker/work_processor.cc for usage:
+   # build_bazel_rules_swift_index_import is be provided by the user for now
```
segiddins, line 95: this feels weird -- either we should provide it in the repositories.bzl, or not have a default in the toolchain at all, imo

segiddins (commented):
We probably should add a test with the feature enabled? Also, will the feature work on Linux? Or is the binary you're using mac only?

swift/internal/actions.bzl
jerrymarino: The current rev uses a `filegroup` to allow `swift_worker_output_processor` in the toolchain to be an empty array and get the same outcome as if it were `None`. The file in the array is _already_ an executable, but inside an array.

I'm happy to make the change but I can't figure out how. I'd prefer it as a `None` overall, any suggestions on how to achive that?

segiddins: added the suggestion in another comment -- use an alias with a select for `actual`, instead of a filegroup

tools/worker/work_processor.cc
jerrymarino: I don't recommend calling it something specific to swift because clang will write to it as well.

swift/internal/swift_toolchain.bzl
jerrymarino: The above line uses worker - this is for the `swift_worker_output_processor`

tools/worker/work_processor.cc
segiddins: I'm not seeing any reference to `global_index_store` in the native cc rules?

swift/internal/swift_toolchain.bzl
segiddins: updated

tools/worker/BUILD
jerrymarino: I left comment earlier [that I didn't think this was possible](https://github.com/bazelbuild/rules_swift/pull/567#discussion_r581331006) 

Specifically, in this code path inside of Bazel `alias.actual` can't be None: https://docs.bazel.build/versions/master/be/general.html#alias.actual

swift/internal/swift_toolchain.bzl
jerrymarino: I made this a fiegroup because I didn't think it was possible to make it an `executable` https://github.com/bazelbuild/rules_swift/pull/567#discussion_r581331006

swift/internal/actions.bzl
jerrymarino: Cool thanks for the suggestion I tried that one this morning to no avail! in this code path inside of Bazel `alias.actual` can't be None: https://docs.bazel.build/versions/master/be/general.html#alias.actual

jerrymarino:
Yeah if you provide it with a linux binary it will work the same on linux. I can do a build of the LLVM binary for linux. Do we want to make compiling a linux binary a blocker for this? `index-import` was never built for linux https://github.com/lyft/index-import/releases/tag/5.3.2.5

tools/worker/work_processor.cc
jerrymarino: This isn't implemented - the plan is written into the rules_ios PR here https://github.com/bazel-ios/rules_ios/pull/212

swift/internal/swift_toolchain.bzl
jerrymarino: Ah thanks - I see what you mean! I added a comment in the thread about why it uses a filegroup and not `executable` https://github.com/bazelbuild/rules_swift/pull/567#discussion_r581331006

tools/worker/work_processor.cc
jerrymarino: Today that works because logic is handled internally in an action binary, like `wrapped_clang` or the `worker`. I've been working to clarify in the design document that this logic is [internal to the worker in rules_swift](https://github.com/bazel-ios/rules_ios/pull/212/files#diff-26cd539e3b611f02c1703c8fce4656a533b33dc08a4479ab5bfa7523260fa292R34).

As far as I knew, it isn't technically possible or gainful to express the paradigm of an internal global store with Bazel's outputs. If you know of a way I'm interested to learn about it. I will further clarify the details in the design document to have a single doc for this code https://github.com/bazel-ios/rules_ios/pull/212/files

amberdixon: Thanks for clarifying the need to write the global indexstore internally within the action!

amberdixon (approved):
+1 to Sam's suggestions to adding tests (if easily possible) and docs. Also, considering the behavior when one, both or neither index_while_building flags are enabled.

I spoke with Jerry offline about the need to use index-import, and I'm onboard with the decision, since building and releasing index-import as a binary and parsing unit files (which is what https://github.com/lyft/index-import/pull/53 is doing) is non-trivial.

jerrymarino:
I like the idea of having some tests for index store integration in general 👍 . 

After syncing up today I'm more in favor of putting the helper binary into `index-import`. I like that this repo already has the eyes of people that know about the indexer. Once we can get [the Xcode indexer working with bazel outputs](https://github.com/bazel-ios/rules_ios/blob/d9841c96f821f0d2926ddff9b664cceb64d87c0e/docs/index_while_building.md#local-optimization---sans-index-import) we may only need the helper binary and hopefully won't have to remap units.

Speaking of building and releasing the binary I didn't notice they had a way to do this yet ? https://github.com/lyft/index-import/releases/tag/5.3.2.5  I think as part of adding a dep on `rules_swift` of the helper program we can setup a buildkite job or action to cut releases on tags.

brentleyjones:
Building index-import requires building part of the swift toolchain, which is pretty expensive. I'm open to ideas though.

jerrymarino:
Cool SGTM 👍  - for local exec we shouldn't need `index-import` to remap indexes but we'll still need it and the Bazel action helper binary to work with remote actions. I've added a line item to create a CI script to release the binary in our roadmap and will also file the issue to rules_ios https://github.com/bazel-ios/rules_ios/pull/212

tools/worker/BUILD
segiddins: bumping this comment

tools/worker/work_processor.cc
segiddins: would it maybe make more sense to pass in the whole output file map, so the output processor could look at the other outputs if it so chooses?

swift/internal/feature_names.bzl
segiddins: also, perhaps it makes sense to give this a more semantic name? "swift.index_while_building_global_indexstore" or something like that?

tools/worker/work_processor.cc
jerrymarino: I don't think so. Clang doesn't use output file maps and the program should work with both clang and swift.

swift/internal/compiling.bzl
jerrymarino: I think that's a lot of complex validation logic but I'm not opposed to that.

swift/internal/feature_names.bzl
jerrymarino: It uses the existing convention `swift.index_while_building` and suffixed it with `v2`. The feature does other things than using a global index store like copying outputs for a compilation and we don't call the original feature `swift.index_while_building_with_per_swift_library_index_stores`

tools/worker/BUILD
jerrymarino: I like that idea, can update to point at the helper binary 👍  https://github.com/bazel-ios/rules_ios/pull/217/files#diff-435e1309098a3a6ba8d1758ed935645d702c31d66bbf05b6f7d026a3d82fdfdcR116

segiddins (commented):
No need to add a linux binary, but we should make sure there's a good error message emitted when enabling the feature on linux (e.g. not relying on the index-import binary just failing to execute)

tools/worker/work_processor.cc
jerrymarino: I think it'd be better to name it something around `index-import` like in your first suggestion then add more complexity and layers of indirection since we don't have a use case for that.

tools/worker/BUILD
jerrymarino: Yeah exactly - we can provide it in `repositories.bzl` - the repository shouldn't be pulled or used in Bazel unless you flip `swift.index_while_building_v2`

jerrymarino: Updated here: https://github.com/bazelbuild/rules_swift/pull/567/files#diff-ea04140a13490474bc161e7f11e1f015dc0780a8142df609568a0a3b05324fddR141

tools/worker/work_processor.cc
jerrymarino: I have removed this from the code altogether.

segiddins (approved):

jerrymarino:
Happy Friday all, I've got this one fully rebased on the latest changes. We've been running it for a few months now and seeing a nice improvement overall. `bazel-out` is down from `20gb` to `7gb` with the change too.

jerrymarino:
cc @keith  @brentleyjones mind giving this a stamp if there's no other changes you think we need?

brentleyjones:
I’ll give this a run/review soon.

tools/worker/work_processor.cc
```diff
@@ -76,16 +83,41 @@ void WorkProcessor::ProcessWorkRequest(
...
+  } else if (!bazel_index_store_path.empty()) {
+    processed_args.push_back("-index-store-path");
+    processed_args.push_back(bazel_index_store_path);
+  }
+
   if (!output_file_map_path.empty()) {
+    output_file_map.ReadFromPath(output_file_map_path);
+
     if (!is_wmo) {
-      output_file_map.ReadFromPath(output_file_map_path);
```
brentleyjones, line 121: nit: Can you remove this newline?
```suggestion
```

tools/worker/work_processor.cc
```diff
@@ -129,7 +161,6 @@ void WorkProcessor::ProcessWorkRequest(
   SwiftRunner swift_runner(processed_args, /*force_response_file=*/true);
 
   int exit_code = swift_runner.Run(&stderr_stream, /*stdout_to_stderr=*/true);
-
```
brentleyjones, line 132: nit: Can you put this newline back?

tools/worker/work_processor.cc
```diff
@@ -143,6 +174,43 @@ void WorkProcessor::ProcessWorkRequest(
     }
   }
 
+  // Import global indexes if compilation succeeds
+  if (exit_code != EXIT_FAILURE && enable_global_index_store) {
+    std::vector<std::string> ii_args;
+
+    // The index-import runfile path is pased as a define
```
brentleyjones, line 181: Why isn't it passed as an argument?

tools/worker/work_processor.cc
```diff
@@ -143,6 +174,43 @@ void WorkProcessor::ProcessWorkRequest(
...
+    #if defined(INDEX_IMPORT_PATH)
+      ii_args.push_back(INDEX_IMPORT_PATH);
+    #else
+      // Logical error
+      std::cerr << "Incorrectly compiled work_processor.cc";
+      exit_code = EXIT_FAILURE;
+    #endif
+
+    // Use the output path map to deterine what compilation ouputs to import
+    auto outputs  = output_file_map.incremental_outputs();
```
brentleyjones, line 191: nit:
```suggestion
    auto outputs = output_file_map.incremental_outputs();
```

tools/worker/work_processor.cc
```diff
@@ -143,6 +174,43 @@ void WorkProcessor::ProcessWorkRequest(
...
+      }
+    }
+
+    // Copy back from the global index store to bazel's index store
+    ii_args.push_back(global_index_store);
+
+    ii_args.push_back(exec_root + "/" + bazel_index_store_path);
+
+    const std::vector<std::string>& ii_args_ = ii_args;
+    // Dupes the output into the action for failure cases
```
brentleyjones, line 210:
```suggestion
    // Dupes the output into the action for failure cases
    const std::vector<std::string>& ii_args_ = ii_args;
```

Why do we need to do this assignment?

swift/internal/feature_names.bzl
```diff
@@ -90,6 +90,10 @@ SWIFT_FEATURE_ENABLE_TESTING = "swift.enable_testing"
...
+# https://docs.google.com/document/d/1cH2sTpgSnJZCkZtJl1aY-rzy4uGPcrI-6RrUpdATO2Q/
+SWIFT_FEATURE_INDEX_WHILE_BUILDING_V2 = "swift.index_while_building_v2"
```
brentleyjones, line 95: Instead of being a mutually exclusive flag, I would make this a flag you have to pass in addition `swift.index_while_building`, maybe something like `swift.use_global_index`. The benefit to doing it this way, is that if it turns out to be the best way to do indexes, we can change the feature to be on by default in the future, so then people only need to pass the existing `swift.index_while_building`.

brentleyjones (commented):
Some comments. I'll test the changes now.

tools/worker/work_processor.cc
```diff
@@ -76,16 +83,41 @@ void WorkProcessor::ProcessWorkRequest(
...
     if (!arg.empty()) {
       params_file_stream << arg << '\n';
     }
 
     prev_arg = original_arg;
   }
 
+  // Material within this index store is namespaced to arcitecture, os, etc
+  auto exec_root = GetCurrentDirectory();
+  auto global_index_store = exec_root + "/bazel-out/global_index_store";
```
brentleyjones, line 108: The way this references `bazel-out` breaks when running from a clean build:

```
WARNING: failed to create one or more convenience symlinks for prefix 'bazel-':
  cannot create symbolic link bazel-out -> /private/var/tmp/_bazel_brentleyjones/0f422387884b3a44e5ecad78262e92a3/execroot/lyftios/bazel-out:  /private/var/tmp/_bazel_brentleyjones/137e1fc04254d0a8e222ea4d6e9bfc0c/execroot/lyftios/bazel-out (File exists)
```

brentleyjones, line 108: Where is this created? I'm not finding it in `bazel-out`.

tools/worker/work_processor.cc
jerrymarino: I think the function is looking for a `const`

tools/worker/work_processor.cc
jerrymarino: The program `swiftc` will run `mkdir -p` on the index store path

swift/internal/compiling.bzl
```diff
@@ -2027,10 +2042,19 @@ def _declare_compile_outputs(
...
     )
+    index_while_building_v2 = is_feature_enabled(
+        feature_configuration = feature_configuration,
+        feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING_V2,
+    )
+
+    if index_while_building_v1 and index_while_building_v2:
+        fail("can't use both swift.index_while_building and swift.index_while_building_v2")
+
+    index_while_building = (index_while_building_v1 or index_while_building_v2)
```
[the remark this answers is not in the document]
jerrymarino: Yeah we can just check `index_while_building` if it's additive

tools/worker/work_processor.cc
jerrymarino: Passing it as an arg required a lot more code and breaking changes to the public toolchain API: expressing and then passing the binary into the toolchain. That was my original approach but got rid of it.

tools/worker/work_processor.cc
jerrymarino: Will do

swift/internal/compiling.bzl
```diff
@@ -1685,6 +1691,15 @@ def compile(
...
         vfsoverlay_file = None
 
+    # For index while building v2 it uses a global index store which is
+    # configured by this flag
+    additional_copts = []
+    if is_feature_enabled(
+        feature_configuration = feature_configuration,
+        feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING_V2,
+    ):
+        additional_copts.append("-Xwrapped-swift=-enable-global-index-store")
```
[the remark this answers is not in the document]
jerrymarino: SGTM - I'm not 100% sure about what happens if you have 2 concurrent swiftc invocations with the same outputs but different args anyhow.

jerrymarino:
Thanks for the review @brentleyjones! I'll go ahead and make the change around `swift.use_global_index`.

It's totally seperate but I think longer term, we'd want to compile `index-import` as an ELF, I've got a WIP github action I can try to cleanup and submit to that repo for releasing

tools/worker/work_processor.cc
brentleyjones: Still, I couldn't find `global_index_store` anywhere. What is an example path for you?

tools/worker/work_processor.cc
brentleyjones: But now if I use this feature globally, but some actions don't use index-import, all swift based actions will be invalidated when updating index-import. I would rather we pass in an argument, unless we can scope this version of the work processor to `swift_action_names.COMPILE + swift.use_global_index`, but the normal version is scoped to all other actions/features.

tools/worker/work_processor.cc
jerrymarino: What I'm saying is that the program `swiftc` runs `mkdir -p with $path/to/` the path you give it via `-index-store-path` if it doesn't exist. In other words, this _is_ where the path is provided to the compiler `swiftc`, which is the program that creates it.

brentleyjones: I understand that. I don't understand why I can't find this global index store on my filesystem after compiling 😆

jerrymarino: @brentleyjones the `bazel-out` directory is in the execroot and symlinked into the WORKSPACE dir.  Likely you haven't got it enabled! What command / repo are you using to build it with? You'll need mainly 2 settings
```
# Use 'worker' strategy for swift compilation
build --strategy=SwiftCompile=worker

# flip on IWBV2
build --features swift.index_while_building_v2
```


You can also test it end to end with [`rules_ios`](https://github.com/bazel-ios/rules_ios/blob/master/.bazelrc) 

```
git clone git@github.com:bazel-ios/rules_ios.git
cd rules_ios/
bazel build tests/ios/app:SwiftLib --apple_platform_type=ios
```

Then running
```
$ du -hs bazel-out/global_index_store/
7.9M    bazel-out/global_index_store/
```

Per library index stores
```
$ find bazel-out/**/ -name \*_swift.indexstore | xargs du -hsc
8.0K    bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-dbg-ST-f15c78a5cd43//bin/tests/ios/app/SwiftLib_swift.indexstore
8.0K    total
```

tools/worker/work_processor.cc
jerrymarino: I thought that feature was supposed to share work between the 2 compiler invocations so it'd noop any extra work anyways - _I generally don't know how it works_!

If you prefer it there for some reason, then it's fine by me. Either way we pack it a runfile, the current version packs it atomically with the worker and it's much less code/easier to rip out in a few years if we can mainline this feature into swiftc.

jerrymarino: I thought that feature was supposed to share work between the 2 compiler invocations so it'd noop any extra work anyways - I generally don't know how it works!

If you prefer it in the toolchain, I can add it in a follow up. Either way we pack it a runfile, the current version packs it atomically with the worker and it's much less code/easier to rip out in a few years if we can mainline this feature into swiftc.

jerrymarino: I thought that feature was supposed to share work between the 2 compiler invocations so it'd noop any extra work anyways - I generally don't know how it works!

If you prefer it in the toolchain, I can add it in a follow up. Either way we pack it a runfile, the current version packs it atomically with the worker and it's easier rip out in a few years if we can mainline this into swiftc.

brentleyjones:
I'll retest again soon, ensuring I have workers enabled.

tools/worker/work_processor.cc
brentleyjones: So the issue was that I didn't have the worker strategy. I'm assuming that is needed for this? That might be why I was getting even larger per-library index stores with this method version v1.

brentleyjones: Here is an example of where I did a lesser form of this change, and it worked for both `local` and `worker`: https://github.com/bazelbuild/rules_swift/commit/6b77a6bd416586947e6e1b8cd600eb59dd772dfe

Ideally this could work for both as well.

brentleyjones:
>It's totally seperate but I think longer term, we'd want to compile index-import as an ELF, I've got a WIP github action I can try to cleanup and submit to that repo for releasing

Sounds good

brentleyjones:
I can confirm this works. For a single configuration we only saw about a 100 MB savings in our 3.9 GB `bazel-out`, but I can see how for some projects it could be a lot more.

What I would like changed before merging though is:

- Going the `swift.use_global_index` route for the feature flag
- Making it work for `local` spawn strategy as well

jerrymarino:
@brentleyjones thanks for taking the time to run it through - it's much appreciated 

> we only saw about a 100 MB savings in our 3.9 GB bazel-out, 

Ahh I could see that, esp depending on indexing system modules, how swiftc classifies modules as system, and if the app is pure swift. The examples I've had were apps heavily mixed with cpp and objc code, importing underlying modules, etc.

I'm getting some cycles to make a pass at this - should have an update soon.

swift/internal/compiling.bzl
```diff
@@ -774,6 +775,13 @@ def compile_action_configs(
             ],
             features = [SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES],
         ),
+        # Note: this *must* go before the index_while_building configurator
+        # as runner code processes this in order.
+        swift_toolchain_config.action_config(
+            actions = [swift_action_names.COMPILE],
+            configurators = [_global_index_store_configurator],
+            features = [SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE],
```
brentleyjones, line 783: I assume we want this to also include `SWIFT_FEATURE_INDEX_WHILE_BUILDING`?
```suggestion
            features = [
                SWIFT_FEATURE_INDEX_WHILE_BUILDING,
                SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE,
            ],
```

swift/internal/compiling.bzl
```diff
@@ -774,6 +775,13 @@ def compile_action_configs(
             ],
             features = [SWIFT_FEATURE_ENABLE_SKIP_FUNCTION_BODIES],
         ),
+        # Note: this *must* go before the index_while_building configurator
+        # as runner code processes this in order.
```
brentleyjones, line 779: Can the runner be adjusted to have this work in any order? I'm less concerned about this file, but more about testing invocations by hand (where order usually doesn't matter).

swift/internal/compiling.bzl
```diff
@@ -2031,6 +2054,14 @@ def _declare_compile_outputs(
         feature_configuration = feature_configuration,
         feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING,
     )
+    use_global_index_store = is_feature_enabled(
+        feature_configuration = feature_configuration,
+        feature_name = SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE,
+    )
+
+    if use_global_index_store and not use_global_index_store:
```
brentleyjones, line 2062:
```suggestion
    if use_global_index_store and not index_while_building:
```

swift/internal/feature_names.bzl
```diff
@@ -90,6 +90,10 @@ SWIFT_FEATURE_ENABLE_TESTING = "swift.enable_testing"
 # warnings otherwise.
 SWIFT_FEATURE_FULL_DEBUG_INFO = "swift.full_debug_info"
 
+# Index while building - using a global index store cache
+# https://docs.google.com/document/d/1cH2sTpgSnJZCkZtJl1aY-rzy4uGPcrI-6RrUpdATO2Q/
+SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE = "swift.use_global_index_store"
```
brentleyjones, line 95: nit: Can we place this after `SWIFT_FEATURE_DISABLE_SYSTEM_INDEX`?

tools/worker/work_processor.cc
jerrymarino: Updated to work with both in latest rev

swift/internal/compiling.bzl
jerrymarino: Sure, we can auto enable this. ATM it requires both

swift/internal/compiling.bzl
jerrymarino: Ahh good catch, thanks for the fix.

swift/internal/compiling.bzl
jerrymarino: It's possible if we want to make another loop over the arguments to get the values. I didn't think its worth processing the args an extra time, since the flags `-Xwrapped-swift=` are made for internal usage in the rules.

swift/internal/feature_names.bzl
jerrymarino: Seems fine.

swift/internal/feature_names.bzl
```diff
@@ -98,16 +98,16 @@ SWIFT_FEATURE_ENABLE_TESTING = "swift.enable_testing"
 # warnings otherwise.
 SWIFT_FEATURE_FULL_DEBUG_INFO = "swift.full_debug_info"
 
-# Index while building - using a global index store cache
-# https://docs.google.com/document/d/1cH2sTpgSnJZCkZtJl1aY-rzy4uGPcrI-6RrUpdATO2Q/
-SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE = "swift.use_global_index_store"
-
 # If enabled, the compilation action for a target will produce an index store.
+# https://docs.google.com/document/d/1cH2sTpgSnJZCkZtJl1aY-rzy4uGPcrI-6RrUpdATO2Q/
```
brentleyjones, line 102: Is this url misplaced?

jerrymarino: It's the docs on how the feature works

tools/worker/swift_runner.h
```diff
@@ -128,6 +135,17 @@ class SwiftRunner {
   // The path to the generated header rewriter tool, if one is being used for
   // this compilation.
   std::string generated_header_rewriter_path_;
+
+  // The path of the global index store  when using
+  // swift.use_global_index_store. When set, this is passed to `swiftc` as the
+  // `-index-store-path`. After running `swiftc` `index-import` copies relevant
+  // index outputs into the `index_store_path` to integrate outputs with Baze.
```
jpsim, line 142:
```suggestion
  // index outputs into the `index_store_path` to integrate outputs with Bazel.
```

jerrymarino:
@googlebot I fixed it

tools/worker/swift_runner.h
```diff
@@ -128,6 +135,17 @@ class SwiftRunner {
   // The path to the generated header rewriter tool, if one is being used for
   // this compilation.
   std::string generated_header_rewriter_path_;
+
+  // The path of the global index store  when using
+  // swift.use_global_index_store. When set, this is passed to `swiftc` as the
+  // `-index-store-path`. After running `swiftc` `index-import` copies relevant
+  // index outputs into the `index_store_path` to integrate outputs with Bazel.
+  std::string global_index_store_import_path_;
```
brentleyjones, line 143: nit: Can this be after `index_store_path_`?

tools/worker/swift_runner.h
```diff
@@ -110,6 +116,7 @@ class SwiftRunner {
   bazel_rules_swift::BazelPlaceholderSubstitutions
       bazel_placeholder_substitutions_;
 
+
```
brentleyjones, line 119: nit: can you remove this extra newline?

tools/worker/BUILD
```diff
@@ -50,13 +55,30 @@ cc_library(
...
-    srcs = ["swift_runner.cc"],
+    srcs = [
+        "output_file_map.cc",
+        "output_file_map.h",
+        "swift_runner.cc",
+    ],
     hdrs = ["swift_runner.h"],
+    copts = select({
+        ":use_global_index_store": [
+            "-DINDEX_IMPORT_PATH=\\\"$(rootpath @build_bazel_rules_swift_index_import//:index_import)\\\"",
```
brentleyjones, line 66: The conversation from before still holds (https://github.com/bazelbuild/rules_swift/pull/567#discussion_r632228239), I just wanted to bring it up and get @keith's opinion on it as well.

My thought was that we should pass in an argument for this, in some way, to prevent non-index producing actions (e.g. `PRECOMPILE_C_MODULE`, `DERIVE_FILES`) from being invalided when `build_bazel_rules_swift_index_import` is updated. I still believe that, though I'm wondering if it can be a fast follow PR.

swift/repositories.bzl
brentleyjones: Also, one can use `bind()` to use the same version in their workspace if they wanted to

brentleyjones (commented):
Thanks @jerrymarino for addressing my comments over this long review period!

I just have a couple more comments, and I need to verify that everything still works, including the new `local` support, before I can approve it.

swift/internal/compiling.bzl
```diff
@@ -2175,6 +2190,14 @@ def _declare_compile_outputs(
...
         feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING,
     )
+    use_global_index_store = is_feature_enabled(
+        feature_configuration = feature_configuration,
+        feature_name = SWIFT_FEATURE_USE_GLOBAL_INDEX_STORE,
+    )
+
+    if use_global_index_store and not index_while_building:
+        fail("can't use swift.use_global_index_store without swift.index_while_building")
+
```
brentleyjones, line 2200: This block should be removed now. The `action_config` above now correctly limits the use of the global index store to only if both features are enabled. There are also times when one will globally enable both of these features, then selectively disable just `SWIFT_FEATURE_INDEX_WHILE_BUILDING`, which can trip up this code path.

tools/worker/BUILD
jerrymarino: It's not clear to me how to actually make that happen in in a way that we can merge this in. At this point I've gotta time-box it. That said, I suspect it'd be beneficial as-is.

I'm also taking a look at adding this into rules_ios and calling the compiler in a simpler way.

jerrymarino:
@brentleyjones ok thanks again for taking the time out of your day to review it! I'm open to submitting more iterative followups if we can get it in 👍 .

tools/worker/BUILD
brentleyjones: Sure. We can do this in a followup.

tools/worker/swift_runner.h
jerrymarino: ✅

tools/worker/swift_runner.h
jerrymarino: ✅

swift/internal/compiling.bzl
jerrymarino: Ah good point, since we flip it on if you've got the new flag, it's unnecessary.

swift/internal/compiling.bzl
```diff
@@ -2175,6 +2190,7 @@ def _declare_compile_outputs(
         feature_configuration = feature_configuration,
         feature_name = SWIFT_FEATURE_INDEX_WHILE_BUILDING,
     )
+
```
brentleyjones, line 2193: nit: please remove

tools/worker/swift_runner.cc
```diff
@@ -129,29 +130,105 @@ int SwiftRunner::Run(std::ostream *stderr_stream, bool stdout_to_stderr) {
     exit_code = RunSubProcess(rewriter_args, stderr_stream, stdout_to_stderr);
   }
 
+  auto exec_root = GetCurrentDirectory();
```
brentleyjones, line 133: Can we move this inside the `if` since it's only used in there?

brentleyjones (approved):
Thanks for all the hard work @jerrymarino!

I have two more small style changes I want, then we can merge it (after getting another approval).

jerrymarino:
@brentleyjones thanks again for the review and @segiddins! I'll follow up with you all soon on this one

swift/repositories.bzl
```diff
@@ -108,6 +108,41 @@ def swift_rules_dependencies():
...
+
+native_binary(
+    name = "index_import",
+    src = "index-import",
+    out = "index-import",
+    visibility = ["//visibility:public"],
+)
+
+native_binary(
+    name = "validate_index",
```
keith, line 128: unused can remove

swift/repositories.bzl
```diff
@@ -108,6 +108,41 @@ def swift_rules_dependencies():
...
+    src = "validate-index",
+    out = "validate-index",
+    visibility = ["//visibility:public"],
+)
+
+native_binary(
+    name = "absolute_unit",
```
keith, line 135: unused can remove

keith (approved):

keith:
note it looks like you have some conflicts

jerrymarino:
Ah looks like more rebase conflicts - taking another look at this.

brentleyjones:
CI seems to be failing.

jerrymarino:
Any ideas on the CI flake? The builds seem to fail on some of the shards:

```
(02:49:58) ERROR: /Users/buildkite/builds/bk-imacpro-20/bazel/rules-swift-swift/examples/xplatform/dispatch/BUILD:5:13: Compiling Swift module examples_xplatform_dispatch_dispatch failed: (Exit 1): worker failed: error executing command
  (cd /private/var/tmp/_bazel_buildkite/5448f2c789b4d9aa3c30b8012d046270/execroot/build_bazel_rules_swift && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=11.1 \
    XCODE_VERSION_OVERRIDE=12.4.0.12D4e \
  bazel-out/host/bin/tools/worker/worker swiftc @bazel-out/darwin-fastbuild/bin/examples/xplatform/dispatch/examples_xplatform_dispatch_dispatch.swiftmodule-0.params)
Execution platform: @local_config_platform//:host
<unknown>:0: error: error opening 'bazel-out/darwin-fastbuild/bin/_swift_incremental/examples/xplatform/dispatch/dispatch_objs/main.swift.swiftdeps' for output: No such file or directory
LLVM ERROR: IO failure on output stream: Bad file descriptor
swift_worker: Could not copy bazel-out/darwin-fastbuild/bin/_swift_incremental/examples/xplatform/dispatch/dispatch_objs/main.swift.o to bazel-out/darwin-fastbuild/bin/examples/xplatform/dispatch/dispatch_objs/main.swift.o (errno 2)
(02:49:58) ERROR: /Users/buildkite/builds/bk-imacpro-20/bazel/rules-swift-swift/examples/xplatform/dispatch/BUILD:5:13 Linking examples/xplatform/dispatch/dispatch failed: (Exit 1): worker failed: error executing command
  (cd /private/var/tmp/_bazel_buildkite/5448f2c789b4d9aa3c30b8012d046270/execroot/build_bazel_rules_swift && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=11.1 \
    XCODE_VERSION_OVERRIDE=12.4.0.12D4e \
  bazel-out/host/bin/tools/worker/worker swiftc @bazel-out/darwin-fastbuild/bin/examples/xplatform/dispatch/examples_xplatform_dispatch_dispatch.swiftmodule-0.params)
```

jerrymarino:
It looks like the bazel CI issue was a transient flake - I bumped the timestamp on the commit and now it's green

brentleyjones:
@jerrymarino Can you work on a followup PR to switch the `DEFINE` for `index-import` to a passed in argument? Thanks!

jerrymarino:
@brentleyjones you've got it. I'm planning to tackle this soon. Realistically I should have something towards mid to end of next week

tools/worker/BUILD
brentleyjones: @jerrymarino Did you give any more thoughts to this?

brentleyjones: https://github.com/bazelbuild/rules_swift/pull/889

PR #569, opened 2021-02-22
Move toolchain- and command-line-provided compiler flags into action configurators.

thii:
This change removes of the `SwiftToolchainInfo.command_line_copts` field, no longer treating those command line flags as an odd special case. It also ensures that flags from the toolchain are passed in a unified manner regardless of which rule registers the action.

For example, Bazel-legacy Objective-C flags determined by compilation mode, like `-fstack-protector`, affect the compatibility of Clang modules and need to be passed both when compiling an explicit module and when compiling the Swift code that consumes those modules.

This required a bit more shuffling around of the way we need to handle differing outptus for WMO, since the `--swiftcopt` flags would no longer be able to be scanned directly after the toolchain configuration.

PiperOrigin-RevId: 355451642
(cherry picked from commit 1bea35e22e18c56818f1c485ab04d2272736ea45)

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -61,6 +61,8 @@ load(
     "SWIFT_FEATURE_USE_C_MODULES",
     "SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE",
     "SWIFT_FEATURE_VFSOVERLAY",
+    "SWIFT_FEATURE__NUM_THREADS_1_IN_SWIFTCOPTS",
+    "SWIFT_FEATURE__WMO_IN_SWIFTCOPTS",
 )
 load(":features.bzl", "are_all_features_enabled", "is_feature_enabled")
 load(":module_maps.bzl", "write_module_map")
@@ -85,11 +87,31 @@ _SWIFTMODULES_VFS_ROOT = "/__build_bazel_rules_swift/swiftmodules"
 # when an API to obtain this is available.
 _DEFAULT_WMO_THREAD_COUNT = 12
 
-def compile_action_configs():
+# Swift command line flags that enable whole module optimization. (This
+# dictionary is used as a set for quick lookup; the values are irrelevant.)
+_WMO_FLAGS = {
+    "-wmo": True,
+    "-whole-module-optimization": True,
+    "-force-single-frontend-invocation": True,
+}
+
+def compile_action_configs(
+        *,
+        additional_objc_copts = [],
+        additional_swiftc_copts = []):
     """Returns the list of action configs needed to perform Swift compilation.
 
     Toolchains must add these to their own list of action configs so that
     compilation actions will be correctly configured.
+
+    Args:
+        additional_objc_copts: An optional list of additional Objective-C
+            compiler flags that should be passed (preceded by `-Xcc`) to Swift
+            compile actions *and* Swift explicit module precompile actions after
+            any other toolchain- or user-provided flags.
+        additional_swiftc_copts: An optional list of additional Swift compiler
+            flags that should be passed to Swift compile actions only after any
+            other toolchain- or user-provided flags.
 
     Returns:
         The list of action configs needed to perform compilation.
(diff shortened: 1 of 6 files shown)


swift/internal/swift_toolchain.bzl
```diff
@@ -89,14 +89,20 @@ def _all_tool_configs(
...
+        additional_swiftc_copts: Additional Swift compiler flags obtained from
+            the `swift` configuration fragment.
+
     Returns:
         A list of action configurations for the toolchain.
     """
     return (
-        compile_action_configs() +
+        compile_action_configs(
+            additional_swiftc_copts = additional_swiftc_copts,
```
thii, line 104: There's a typo in the param name here in the `upstream` branch.

https://github.com/bazelbuild/rules_swift/blob/ab1c5cd0038c76a88738d8aff7397e728ccad3cf/swift/internal/swift_toolchain.bzl#L101

swift/internal/compiling.bzl
```diff
@@ -764,6 +802,40 @@ def compile_action_configs():
...
+                    swift_action_names.COMPILE,
+                    swift_action_names.PRECOMPILE_C_MODULE,
+                ],
+                configurators = [
+                    # TODO(#568): Switch to using lambda when the minimum
+                    # supported Bazel version by rules_swift supports it.
+                    partial.make(
+                        _additional_objc_copts_configurator,
+                        additional_objc_copts,
+                    ),
```
thii, line 818: The `upstream` branch uses lambda here, but we can't use it now as it's not yet available in any Bazel release.

swift/internal/compiling.bzl
```diff
@@ -2340,35 +2471,28 @@ def _emitted_output_nature(is_wmo_implied_by_features, user_compile_flags):
...
-    if num_threads == None or num_threads < 0:
-        fail("The value of '-num-threads' must be a positive integer.")
+    # We check the feature first because that implies that `-num-threads 1` was
+    # present in `--swiftcopt`, which overrides all other flags (like the user
+    # compile flags, which come from the target's `copts`). Only fallback to
+    # checking the flags if the feature is disabled.
+    is_single_threaded = is_feature_enabled(
+        feature_configuration = feature_configuration,
+        feature_name = SWIFT_FEATURE__NUM_THREADS_1_IN_SWIFTCOPTS,
+    ) or _find_num_threads_flag_value(user_compile_flags) == 1
```
thii, line 2492: In the master branch, this is being compared to 0, which seems like a workaround (?) to remove the `-num-threads` flag (since it's equivalent to not passing `-num-threads` at all)? Maybe we should have a feature for it instead of overriding the logic here.

keith: This should be 0 instead. The problem is the 1 is wrong for the intent of this AFAIUI. Swift produces a single object file only if you use 0 as the value here, where if you use 1+ it produces multiple object files even though in the case of 1 it's with a single thread. This behavior in the compiler is useful for weeding out parallelization issues // @allevato

allevato: Right; we haven't applied any fixes like https://github.com/bazelbuild/rules_swift/pull/440 in our version of the rules yet, so I wanted to make sure that this was solely a refactoring change and didn't also sneak in other fixes/behavioral changes. Based on the linked PR, you should change it to 0 to match your existing version of the code.

My preference, though, would be to just do some data collection and see if there's any reason at all to supporting `-num-threads` from a performance point of view, and if not, then wipe it entirely from the rules and add some detection code to ban it, because it would be lovely to simplify the output file selection logic. Have you done anything similar and found that it's worth keeping the support?

keith: In the past we tested a bunch of values for this and didn't see any performance win. Until recently we were actually passing `-num-threads 0` because it had the side effect of working around a code coverage bug https://bugs.swift.org/browse/SR-9145 but AFAICT that is fixed now.

swift/internal/swift_toolchain.bzl
thii: @allevato While you are here, just wanted to point out this was causing the Linux builds to fail.

allevato: Thanks, and sorry for the typo—the `swift_toolchain.bzl` file that you see is an open-source-only file, so we don't actually use it (Google's copy is different) or have any testing on it. So those builds failing was, err, the testing that we do have doing its job 😒 

I've considered just deleting the file from our copy of the rules and not trying to keep it in sync anymore, but in the event of large refactorings, that would leave you all with more work to do to chase those down...

brentleyjones (approved):

PR #574, opened 2021-02-25
Pass `-no-clang-module-breadcrumbs` to frontend jobs by default

thii:
This adds a feature `swift.embed_clang_module_breadcrumbs`, if enabled,
embed absolute paths to the dependent pcm files in the debug info. This
is used as a fallback path for LLDB to reconstruct Clang types from
DWARF, when it wasn't able to import a dependent Clang module from
source. This feature is disabled by default (by passing `-Xfrontend
-no-clang-module-breadcrumbs` to swiftc) to make the compilation
outputs more cacheable.

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -211,6 +211,13 @@ SWIFT_FEATURE_SUPPORTS_PRIVATE_DEPS = "swift.supports_private_deps"
 # system command line limit.
 SWIFT_FEATURE_NO_EMBED_DEBUG_MODULE = "swift.no_embed_debug_module"
 
+# If enabled, embed absolute paths to the dependent pcm files in the debug
+# info. This is used as a fallback path for LLDB to reconstruct Clang types
+# from DWARF, when it wasn't able to import a dependent Clang module from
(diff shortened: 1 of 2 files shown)


swift/internal/feature_names.bzl
```diff
@@ -211,6 +211,13 @@ SWIFT_FEATURE_SUPPORTS_PRIVATE_DEPS = "swift.supports_private_deps"
 # system command line limit.
 SWIFT_FEATURE_NO_EMBED_DEBUG_MODULE = "swift.no_embed_debug_module"
 
+# If enabled, embed absolute paths to the dependent pcm files in the debug
+# info. This is used as a fallback path for LLDB to reconstruct Clang types
+# from DWARF, when it wasn't able to import a dependent Clang module from
+# source. This feature is disabled by default to make the compilation outputs
+# more cacheable.
+SWIFT_FEATURE_EMBED_CLANG_MODULE_BREADCRUMBS = "swift.embed_clang_module_breadcrumbs"
```
keith, line 219: i wonder if we should even have a feature for this unless it comes up as something folks need?

keith (approved):

PR #579, opened 2021-03-12
Add a `generates_header` attribute to `swift_library` to control Objective-C header generation.

keith:
Today, `swift_library` unconditionally generates a header and module map for all targets (unless rarely-used features are applied), even if the library has no APIs exported to Objective-C. This creates a small amount of unnecessary work, but that work grows significantly when explicit modules are used in the build, because we also need to compile the explicit modules for those generated headers _just in case_ some `objc_library` further up in the build graph needs to import the header.

This is the first of a few changes that will eventually require the `generates_header` attribute to be set on targets to have the generated header emitted and propagated.

This change also removes the `swift.no_generated_header` feature, since the attribute's behavior supplants it.

PiperOrigin-RevId: 362562962
(cherry picked from commit e621e5e644c7f62f44241da253fb40ea14672451)

### DIFF
--- a/swift/internal/attrs.bzl
+++ b/swift/internal/attrs.bzl
@@ -271,8 +271,27 @@ a `.h` extension and cannot contain any path separators.
 If this attribute is not specified, then the default behavior is to name the
 header `${target_name}-Swift.h`.
 
-This attribute is ignored if the toolchain does not support generating headers
-or if the target has the `swift.no_generated_header` feature enabled.
+It is an error to specify a value for this attribute when `generates_header` is
+False.
+""",
+                mandatory = False,
+            ),
+            "generates_header": attr.bool(
+                # TODO(b/182493307): Make the default False after migrating all
+                # targets to explicitly specify it when needed.
+                default = True,
+                doc = """\
+If True, an Objective-C header will be generated for this target, in the same
+build package where the target is defined. By default, the name of the header is
+`${target_name}-Swift.h`; this can be changed using the `generated_header_name`
+attribute.
+
+Targets should only set this attribute to True if they export Objective-C APIs.
+A header generated for a target that does not export Objective-C APIs will be
+effectively empty (except for a large amount of prologue and epilogue code) and
+this is generally wasteful because the extra file needs to be propagated in the
+build graph and, when explicit modules are enabled, extra actions must be
+executed to compile the Objective-C module for the generated header.
 """,
                 mandatory = False,
             ),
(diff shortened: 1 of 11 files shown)


test/generated_header_tests.bzl
```diff
@@ -102,15 +74,15 @@ def generated_header_test_suite(name = "generated_header"):
...
+    analysis_failure_test(
         name = "{}_invalid_extension".format(name),
         expected_message = "The generated header for a Swift module must have a '.h' extension",
         tags = [name],
         target_under_test = "@build_bazel_rules_swift//test/fixtures/generated_header:invalid_extension",
     )
 
     # Verify that the build analyzes if a path separator is used.
-    generate_header_and_module_map_provider_test(
+    provider_test(
```
keith, line 85: This one is a bit different from upstream because we made this case work

swift/internal/compiling.bzl
```diff
@@ -1691,10 +1689,7 @@ def compile(
 
     # If a header and module map were generated for this Swift module, attempt
     # to precompile the explicit module for that header as well.
-    if not is_feature_enabled(
```
jerrymarino, line 1694: Minor, should this just error out?

swift/internal/attrs.bzl
```diff
@@ -271,8 +271,27 @@ a `.h` extension and cannot contain any path separators.
...
 header `${target_name}-Swift.h`.
 
-This attribute is ignored if the toolchain does not support generating headers
-or if the target has the `swift.no_generated_header` feature enabled.
+It is an error to specify a value for this attribute when `generates_header` is
+False.
+""",
+                mandatory = False,
+            ),
+            "generates_header": attr.bool(
```
jerrymarino, line 279: Perhaps call this `generate_header`, to more align with `alwayslink` as opposed to `alwayslinks`

jerrymarino (approved):
Saw this in my inbox and the line on explicit module compilation seemed interesting, quick drive by review and nice PR 👍 - using some of these features in rules_ios

thii (approved):

swift/internal/attrs.bzl
brentleyjones: I think we will want to keep whatever name `upstream` uses. Maybe google can change their version? @allevato

allevato: I've migrated a very large number of targets in our monorepo to use this name already, so I have no plans to change it. 🙂

brentleyjones (approved):

PR #581, opened 2021-03-23
Add `swift.global_module_cache_uses_tmpdir` feature

brentleyjones:
This adds a new feature, that can be enabled by passing `--features=swift.use_global_module_cache` and `--features=swift.global_module_cache_uses_tmpdir` to the build command.

This feature, when enabled, will makes the Swift compilation actions use the shared Clang module cache path written to `/private/tmp/__build_bazel_rules_swift/swift_module_cache/REPOSITORY_NAME`. This makes the embedded Clang module breadcrumbs deterministic between Bazel instances, because they are always embedded as absolute paths. Note that the use of this cache is non-hermetic--the cached modules are not wiped between builds, and won't be cleaned when invoking bazel clean; the user is responsible for manually cleaning them.

Additionally, this can be used as a workaround for a bug in the Swift compiler that causes the module breadcrumbs to be embedded even though the `-no-clang-module-breadcrumbs` flag is passed (https://bugs.swift.org/browse/SR-13275).

Since the source path of modulemaps might be different for the same module, (i.e. multiple checkouts of the same repository, or remote execution), multiple modules with different hashes can end up in the cache. This can result in build failures. Don't use this feature with sandboxing (or probably remote execution as well).

---

This is based on my original commit (106add0a329010d358464955a15dc2836d53f13a) and PR #456.

Closes #456.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -506,7 +506,7 @@ def compile_action_configs(
             configurators = [_tmpdir_module_cache_configurator],
             features = [
                 SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE,
-                SWIFT_FEATURE_GLOBAL_MODULE_CACHE_USES_TMPDIR
+                SWIFT_FEATURE_GLOBAL_MODULE_CACHE_USES_TMPDIR,
             ],
             not_features = [SWIFT_FEATURE_USE_C_MODULES],
         ),
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -44,6 +44,7 @@ load(
     "SWIFT_FEATURE_ENABLE_TESTING",
     "SWIFT_FEATURE_FASTBUILD",
     "SWIFT_FEATURE_FULL_DEBUG_INFO",
+    "SWIFT_FEATURE_GLOBAL_MODULE_CACHE_USES_TMPDIR",
     "SWIFT_FEATURE_INDEX_WHILE_BUILDING",
     "SWIFT_FEATURE_LAYERING_CHECK",
     "SWIFT_FEATURE_MODULE_MAP_HOME_IS_CWD",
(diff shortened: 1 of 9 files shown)


brentleyjones:
@thii I made some changes to how #456 worked, based on my testing with remote execution (where I use this with a slight local tweak).

swift/internal/compiling.bzl
```diff
@@ -942,6 +958,18 @@ def _global_module_cache_configurator(prerequisites, args):
...
 
+def _tmpdir_module_cache_configurator(prerequisites, args):
+    """Adds flags to enable a stable tmp directory module cache."""
+
+    args.add(
+        "-module-cache-path",
+        paths.join(
+            "/private/tmp/__build_bazel_rules_swift",
+            "swift_module_cache",
+            prerequisites.workspace_name,
```
brentleyjones, line 969: The workspace name here is to give some form of isolation around different builds. Without this you have a higher chance of having the same modulemap compiled from different locations. Locally we add something else to this for remote execution, so I could be convinced to use a build setting or define instead.

swift/internal/feature_names.bzl
```diff
@@ -158,6 +158,29 @@ SWIFT_FEATURE_USE_C_MODULES = "swift.use_c_modules"
...
+# `-no-clang-module-breadcrumbs` flag is passed
+# (https://bugs.swift.org/browse/SR-13275).
+#
+# Since the source path of modulemaps might be different for the same module,
+# (i.e. multiple checkouts of the same repository, or remote execution),
+# multiple modules with different hashes can end up in the cache. This can
+# result in build failures. Don't use this feature with sandboxing (or
+# probably remote execution as well).
+#
+# This feature requires `swift.use_global_module_cache` to be enabled.
```
brentleyjones, line 181: This felt more right then having them be mutually exclusive, or this feature working regardless of the other being enabled.

swift/internal/compiling.bzl
thii: Would it be possible to upstream the change needed for remote execution?

thii: The `swift.use_global_module_cache` feature doesn't share cache between different configurations (so, different `bin_dir`s). Should this have the same behavior?

https://github.com/bazelbuild/rules_swift/blob/2a20daa7ce087195b3fe0dbde78f7c5209526946/swift/internal/compiling.bzl#L939-L943

brentleyjones: I originally had that, but found that it's safe to share the cache in this way, since opt vs dbg will get different hash folders, and this saves space on remote execution worker machines (where it has a multiplier effect). It still uses the workspace to silo based on project though.

brentleyjones: >Would it be possible to upstream the change needed for remote execution?

The change is listening to a execution service specific environment variable, and appending a magic path here, say `_EXECUTOR_ID_`, which gets replaced in the swift worker. I'll see how generic I can make that so it would work with any service that exposes a variable like that.

thii: I was thinking if it's safe to share the cache between platforms (for example, iOS and watchOS). I can't test this now but I'll report back if it turns out to be an issue for us.

thii (approved):

swift/internal/compiling.bzl
```diff
@@ -927,6 +943,18 @@ def _global_module_cache_configurator(prerequisites, args):
             paths.join(prerequisites.bin_dir.path, "_swift_module_cache"),
         )
 
+def _tmpdir_module_cache_configurator(prerequisites, args):
+    """Adds flags to enable a stable tmp directory module cache."""
+
+    args.add(
+        "-module-cache-path",
+        paths.join(
+            "/private/tmp/__build_bazel_rules_swift",
```
keith, line 952:
```suggestion
            "/tmp/__build_bazel_rules_swift",
```

if there's no difference I think using this would be preferred since I view that symlink as a bit of an implementation detail, but maybe there is?

PR #596, opened 2021-04-19
Sort headers in the module maps generated by `swift_clang_module_aspect`, and rewrite generation to use a multi-line `Args` object as a file writer instead of building up an analysis time string.

keith:
With this approach, each "arg" (or each entry in a list/depset of args) is treated as its own line in the output file; we use the `format_each` and `map_each` parameters to provide any additional text that should surround that value on the line.

More importantly, this adds support for expanding tree artifacts (directories) if one is provided in a compilation context.

PiperOrigin-RevId: 367221120
(cherry picked from commit 5f7af6972ca83d703fa1858a6bb568d2e300ed33)

### DIFF
--- a/swift/internal/module_maps.bzl
+++ b/swift/internal/module_maps.bzl
@@ -14,8 +14,6 @@
 
 """Logic for generating Clang module map files."""
 
-load("@bazel_skylib//lib:paths.bzl", "paths")
-
 def write_module_map(
         actions,
         module_map_file,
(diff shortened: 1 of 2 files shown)


keith:
ha seems like nested functions are a newer feature than what is currently released in bazel

thii:
Maybe it's worth requesting the change for that feature to be cherry-picked into 4.1, if it's not in there yet.

Issue #468, opened 2020-08-11
Add docs generation


keith:
https://github.com/bazelbuild/stardoc

alexeagle:
Maybe I can help with this, I just added stardoc to rules_docker

keith:
that would be awesome!

alexeagle:
By the way, now that this uses stardoc, it looks a lot better on our "g3doc for rulesets" 
https://docs.aspect.dev/rules_swift
😎

PR #598, opened 2021-04-30
Add first stardoc API docgen target

alexeagle:
See #468


keith (approved):
did you want to add something more to this?

keith:
ah looks like the tests reveal some issues

alexeagle:
Hmm, interesting despite all my usual tricks, stardoc fails when running on CI due to inability to locate our self-references. It does run locally so I will make a follow-up change to get the tests running. At least in the meantime it's an improvement that you have a generator hooked up.

alexeagle:
CI failure seems to be pre-existing

PR #615, opened 2021-05-04
Disable layering checks when compiling the explicit module for a Swift generated header.

keith:
The Swift compiler determines which module to import a symbol from based on the module that defines that symbol. Due to modular re-exports (which are particularly common among system frameworks), this may not be the same framework that the user imported from Swift and added to their dependencies.

(For example, most users will import and depend on `Foundation` to use `NSObject`, but Swift will import it from the module it is actually defined in, `ObjectiveC`.)

PiperOrigin-RevId: 371919747
(cherry picked from commit d8a381c0174d3a445389c3decc3727d41b80b255)

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -474,6 +474,21 @@ def compile_action_configs(
     ]
 
     #### Flags controlling how Swift/Clang modular inputs are processed
+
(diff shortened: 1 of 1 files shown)


keith:
depends on lambdas again, might wait on this one since it seems low priority

PR #623, opened 2021-05-28
Add the ability to dump ast to a file

segiddins:
Some tooling (namely https://github.com/square/Cleanse/tree/master/cleansec) operates over dumped swift AST files.
In order to dump AST files, (almost) complete `swiftc` invocations are required, in addition to the `-dump-ast` flag.
Because almost the same set of arguments as module compilation are used for dumping the AST, it is nearly impossible for a bazel user to reconstruct them by hand (or via a standalone rule).

This new output group (and action) serves to dump AST files for each swift source file in a `swift_library` swift module.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1903,6 +1903,7 @@ def compile(
     )
 
     return struct(
+        ast_files = compile_outputs.ast_files,
         generated_header = compile_outputs.generated_header_file,
         generated_module_map = compile_outputs.generated_module_map_file,
         indexstore = compile_outputs.indexstore_directory,
@@ -1916,7 +1917,6 @@ def compile(
         swiftdoc = compile_outputs.swiftdoc_file,
         swiftinterface = compile_outputs.swiftinterface_file,
         swiftmodule = compile_outputs.swiftmodule_file,
-        ast_files = compile_outputs.ast_files,
     )
 
 def precompile_clang_module(
@@ -2323,6 +2323,7 @@ def _declare_multiple_outputs_and_write_output_file_map(
     # action outputs although they are not processed further.
     other_outputs = []
 
+    # AST files that are available in the swift_ast_file output group
     ast_files = []
 
     for src in srcs:
@@ -2365,8 +2366,8 @@ def _declare_multiple_outputs_and_write_output_file_map(
     )
 
     return struct(
-        object_files = output_objs,
         ast_files = ast_files,
+        object_files = output_objs,
         other_outputs = other_outputs,
         output_file_map = output_map_file,
     )
@@ -2643,6 +2644,11 @@ def output_groups_from_compilation_outputs(compilation_outputs):
     """
     output_groups = {}
 
+    if compilation_outputs.ast_files:
+        output_groups["swift_ast_file"] = depset(
+            compilation_outputs.ast_files,
+        )
+
     if compilation_outputs.indexstore:
         output_groups["swift_index_store"] = depset([
             compilation_outputs.indexstore,
@@ -2662,11 +2668,6 @@ def output_groups_from_compilation_outputs(compilation_outputs):
         output_groups["swiftmodule"] = depset([
             compilation_outputs.swiftmodule,
         ])
-
-    if compilation_outputs.ast_files:
-        output_groups["swift_ast_file"] = depset(
-            compilation_outputs.ast_files,
-        )
 
     return output_groups
 
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -142,6 +142,19 @@ def compile_action_configs(
             configurators = [_output_swiftmodule_or_file_map_configurator],
         ),
 
+        # Dump ast files
+        swift_toolchain_config.action_config(
+            actions = [swift_action_names.DUMP_AST],
+            configurators = [
+                swift_toolchain_config.add_arg("-dump-ast"),
+                swift_toolchain_config.add_arg("-suppress-warnings"),
+            ],
+        ),
+        swift_toolchain_config.action_config(
+            actions = [swift_action_names.DUMP_AST],
+            configurators = [_output_ast_path_or_file_map_configurator],
+        ),
+
         # Emit precompiled Clang modules, and embed all files that were read
         # during compilation into the PCM.
         swift_toolchain_config.action_config(
@@ -257,6 +270,7 @@ def compile_action_configs(
             actions = [
                 swift_action_names.COMPILE,
                 swift_action_names.DERIVE_FILES,
+                swift_action_names.DUMP_AST,
             ],
             configurators = [
                 swift_toolchain_config.add_arg("-DDEBUG"),
@@ -267,6 +281,7 @@ def compile_action_configs(
             actions = [
                 swift_action_names.COMPILE,
                 swift_action_names.DERIVE_FILES,
+                swift_action_names.DUMP_AST,
             ],
             configurators = [
                 swift_toolchain_config.add_arg("-DNDEBUG"),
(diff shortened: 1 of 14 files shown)


thii (approved):

swift/internal/derived_files.bzl
```diff
@@ -17,6 +17,22 @@
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load(":utils.bzl", "owner_relative_path")
 
+def _ast(actions, target_name, src):
+    """Declares a file for an intermediate object file during compilation.
```
brentleyjones, line 21: The comment is incorrect (copied from `_intermediate_object_file`). Can you also be more specific in the comments for the args and returns?

jerrymarino:
I think there's more to running this frontend action in Bazel than dumping the args: it allows us to schedule the actions with Bazel. That being said, I think we could run different frontend actions with an aspect and simplify the code. What are the tradeoffs of doing that approach? 

I was also wondering about the use cases forit. I imagine that I'd want to run this frontend action seperately than a compiler invocation. For example, I'd like to extract ASTs but not run code generation for object files. If that's possible maybe we should disable object file generation.  Is that possible?

jerrymarino:
I think there's more to running this frontend action in Bazel than dumping the args: it allows us to schedule the actions with Bazel. That being said, I think we could run different frontend actions with an aspect and simplify the code. What are the tradeoffs of doing that approach? 

I was also wondering about the use cases of this feature. I imagine that I'd want to run this frontend action separately than an code emitting compiler invocation. For example, I'd like to extract ASTs but not run code generation for object files. If that's possible maybe we should disable object file generation.  Is that possible?

segiddins:
> Doesn't bazel aquery provide this?

Yes, but not in a way that can easily be integrated into a rule (and we'd like to consume the dumped AST in rules!)

segiddins:
>  I imagine that I'd want to run this frontend action separately than an code emitting compiler invocation. For example, I'd like to extract ASTs but not run code generation for object files. If that's possible maybe we should disable object file generation. Is that possible?

That's what's happening here, since dumping the AST is generated from a separate action.

> That being said, I think we could run different frontend actions with an aspect and simplify the code.

That would almost certainly be _more_ code, since rules_swift exposes no way to get at the configurators used to construct the `swiftc` invocations, and `-dump-ast` uses a subset of the flags used for `-emit-object` or `-emit-swiftmodule`

jerrymarino:
>That's what's happening here, since dumping the AST is generated from a separate action.

Ah so I can specify the output group then and just run the AST actions? That's pretty cool.


> That would almost certainly be more code

It could be nice if that approach can simplifies the core of `rules_swift`: less features, less outputs, and less logic on the compilation side of things, and maybe different code this feature.

FWIW I've seen some code use an aspect to setup a compdb as well strictly based on args and metadata from rules. I'll try to dig that up to get a reference to it to see if it's applicable or helpful to consider if it's viable

jerrymarino:
PS: the aspect was something like this
```py
# print args
def _a_impl(target, ctx):
  for action in target.actions:
    if action.mnemonic == "SwiftCompile":
      action_args = action.argv
      print("ARGS", action_args)
      # TODO: schedule action here
  return []

print_aspect = aspect(
  implementation = _a_impl,
  attr_aspects = ['deps'],
)```

File: swift/internal/compiling.bzl [hunk 2]
```diff
@@ -139,6 +139,19 @@ def compile_action_configs(
             configurators = [_output_swiftmodule_or_file_map_configurator],
         ),
 
+        # Dump ast files
+        swift_toolchain_config.action_config(
+            actions = [swift_action_names.DUMP_AST],
+            configurators = [
+                swift_toolchain_config.add_arg("-dump-ast"),
+                swift_toolchain_config.add_arg("-suppress-warnings"),
```
keith, line 147: not sure about this one, seems like something you should probably pass manually as needed?

tools/worker/swift_runner.cc
```diff
@@ -198,8 +199,8 @@ bool SwiftRunner::ProcessArgument(
         consumer(GetCurrentDirectory() + "=.");
         changed = true;
       } else if (new_arg == "-coverage-prefix-pwd-is-dot") {
-        // Get the actual current working directory (the workspace root), which we
-        // didn't know at analysis time.
+        // Get the actual current working directory (the workspace root), which
+        // we didn't know at analysis time.
```
keith, line 203: looks like you can revert this file

keith (approved):

swift/internal/compiling.bzl
segiddins: probably -- i guess it just means duplicated warnings for the same file/line ?

keith: Ohh, in the case you're building both at once? interesting yea, seems fine to leave

jerrymarino (approved):

tools/worker/swift_runner.cc
jerrymarino: I hit the same issue on my PR and it is due to the master branch not having clang-format up to date on the files. Maybe worth just running clang-format on the repo first and rebasing to get the white space out of PRs when people run it.

tools/worker/work_processor.cc
```diff
@@ -69,11 +70,18 @@ void WorkProcessor::ProcessWorkRequest(
     // necessary later.
     if (arg == "-output-file-map") {
       arg.clear();
+    } else if (arg == "-dump-ast") {
```
jerrymarino, line 73: @segiddins it looks like this doesn't work with the `local` strategy - can you confirm?

 I've refactored some of the argument parsing logic in [this PR](https://github.com/bazelbuild/rules_swift/pull/567) so is now possible to strip off arguments and parse arguments to ivars in the runner.

jerrymarino (changes requested):

tools/worker/work_processor.cc
segiddins: 👍🏻  now handled in both places (runner for avoiding setting WMO flag, work processor for avoiding incremental mode as well)

keith (approved):

swift/internal/compiling.bzl
```diff
@@ -1859,6 +1916,7 @@ def compile(
         swiftdoc = compile_outputs.swiftdoc_file,
         swiftinterface = compile_outputs.swiftinterface_file,
         swiftmodule = compile_outputs.swiftmodule_file,
+        ast_files = compile_outputs.ast_files,
```
brentleyjones, line 1919: nit: can this be at the top to preserve ordering?

swift/internal/compiling.bzl
```diff
@@ -2258,6 +2323,8 @@ def _declare_multiple_outputs_and_write_output_file_map(
     # action outputs although they are not processed further.
     other_outputs = []
 
+    ast_files = []
```
brentleyjones, line 2326: Should this have a comment like the other ones?

swift/internal/compiling.bzl
```diff
@@ -2291,6 +2366,7 @@ def _declare_multiple_outputs_and_write_output_file_map(
 
     return struct(
         object_files = output_objs,
+        ast_files = ast_files,
```
brentleyjones, line 2369: nit: can this be moved to the top?

swift/internal/compiling.bzl
```diff
@@ -2587,6 +2663,11 @@ def output_groups_from_compilation_outputs(compilation_outputs):
             compilation_outputs.swiftmodule,
         ])
 
+    if compilation_outputs.ast_files:
+        output_groups["swift_ast_file"] = depset(
+            compilation_outputs.ast_files,
+        )
+
```
brentleyjones, line 2670: nit: Can this be moved to above `if compilation_outputs.indexstore` to keep it alphabetical?

brentleyjones (approved):

PR #636, opened 2021-06-09
Remove the `-rpath` for the Xcode's developer frameworks directory fr…

segiddins:
…om the Swift toolchain.

It may have been needed in the past with older versions of Swift, but it doesn't appear to be anymore (and it causes problems if the version targeted during the build is mismatched with the simulator/OS version under which the test is running).

PiperOrigin-RevId: 378400910
(cherry picked from commit 92b727ede6af6a09ec4977c4ea5257c3357a5986)

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -264,15 +264,10 @@ def _default_linker_opts(
     if use_system_swift_libs:
         linkopts.append("-L/usr/lib/swift")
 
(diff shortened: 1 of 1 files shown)


keith (approved):
i haven't looked to see if this conflicts with https://github.com/bazelbuild/rules_swift/commit/bf5fccc55f83e0583c193c97cbbc7913ac005715 but if it does we might want to take that one first

segiddins:
Good call, https://github.com/bazelbuild/rules_swift/pull/637

PR #641, opened 2021-06-09
Feature flag to set num-threads to host threads

jerrymarino:
Currently, num-threads is hardcoded to 12. This PR sets the value to the
number of cores return by sysctl, or "host threads". This may improve
vertical scaling in some situations for larger core machines.

### DIFF
--- a/tools/worker/swift_runner.cc
+++ b/tools/worker/swift_runner.cc
@@ -15,6 +15,7 @@
 #include "tools/worker/swift_runner.h"
 
 #include <fstream>
+#include <thread>
 
 #include "tools/common/bazel_substitutions.h"
 #include "tools/common/file_system.h"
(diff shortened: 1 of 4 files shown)


thii (approved):
This is nice, presuming there's no side-effects on reproducibility.

tools/worker/swift_runner.cc
```diff
@@ -318,6 +319,14 @@ bool SwiftRunner::ProcessArgument(
         ++itr;
         new_arg = output_file_map_path_;
         changed = true;
+      } else if (use_host_num_threads_ && arg == "-num-threads") {
```
keith, line 322: this should probably be an error or something instead? like you shouldn't pass this and enable this feature at the same time I don't think?

tools/worker/swift_runner.cc
```diff
@@ -102,7 +103,7 @@ static bool StripPrefix(const std::string &prefix, std::string &str) {
 
 SwiftRunner::SwiftRunner(const std::vector<std::string> &args,
                          bool force_response_file)
-    : force_response_file_(force_response_file) {
+    : force_response_file_(force_response_file), use_host_num_threads_(false) {
```
keith, line 106: since the other feature bools aren't here you shouldn't need this one here either i don't think?

brentleyjones:
`upstream` is going a different way to solve parallelism.

PR #644, opened 2021-06-10
Fix race condition in MakeDirs

jerrymarino:
If there are 2 concurrent processes creating the same directories, the existing code fails. I suspect this is "transient" errors we saw in #  567's final CI run, but this is unrelated to the PR. 

I suspect invalidating the worker or other updates triggered big rebuilds. Myself and a co-worker hit this today when we bumped rules_swift. 

This manifests in many ways when using anything that needs a directory e.g.
```
<unknown>:0: error: error opening 'bazel-out/macos-x86_64-applebin_macos-darwin_x86_64-dbg-ST-15eaacdcbb7d/bin/_swift_incremental/tests/macos/xcconfig/array_key_xcconfig_swift_objs/empty.swift.swiftdeps' for output: No such file or directory
```

Note: I don't have a perfect repro - at a 10% rate: invoking Bazel to build 2 swift libraries in the same BUILD file seemed to do the trick.

### DIFF
--- a/tools/common/file_system.cc
+++ b/tools/common/file_system.cc
@@ -18,6 +18,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <iostream>
 #include <string>
 
 #ifdef __APPLE__
@@ -99,7 +100,13 @@ bool MakeDirs(const std::string &path, int mode) {
   struct stat dir_stats;
   if (stat(path.c_str(), &dir_stats) == 0) {
     // Return true if the directory already exists.
-    return S_ISDIR(dir_stats.st_mode);
+    if (!S_ISDIR(dir_stats.st_mode)) {
+      std::cerr << "error: path exists and isn't a directory (" << path.c_str()
+                << ") \n";
+      return false;
+    } else {
+      return true;
+    }
   }
 
   // Recurse to create the parent directory.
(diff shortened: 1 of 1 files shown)


tools/common/file_system.cc
```diff
@@ -108,5 +115,22 @@ bool MakeDirs(const std::string &path, int mode) {
...
+    int chmod_ret = chmod(path.c_str(), S_IRWXU);
+    if (chmod_ret == 0) {
+      return true;
+    }
+    std::cerr << "error:" << strerror(errno) << ":" << path.c_str() << "\n";
+    return false;
+  }
+
+  // If we can access the directory assume it's valid
+  if (access(path.c_str(), F_OK) == 0) {
```
brentleyjones, line 131: We need to do the same `S_ISDIR` check, since a file could have been placed here instead: #645.

brentleyjones: I like your additional error handling though, so I'm fine with taking your PR if that change is made.

brentleyjones:
The core issue is when two actions try to create directories that need the same parent directories created, like the `_swift_incremental` tree. It won't trigger otherwise, since it currently silently ignores the errors in the worker.

tools/common/file_system.cc
```diff
@@ -108,5 +115,29 @@ bool MakeDirs(const std::string &path, int mode) {
   }
 
   // Create the directory that was requested.
-  return mkdir(path.c_str(), mode) == 0;
+  int mkdir_ret = mkdir(path.c_str(), mode);
+  if (mkdir_ret == 0) {
+    return true;
+  }
+
+  // Save the mkdir errno for better reporting
```
brentleyjones, line 123: nit: 
```suggestion
  // Save the mkdir errno for better reporting.
```

tools/common/file_system.cc
```diff
@@ -108,5 +115,29 @@ bool MakeDirs(const std::string &path, int mode) {
...
+  int mkdir_errno = errno;
+
+  // Deal with a race condition when 2 `MakeDirs` are running at the same time:
+  // one `mkdir` invocation will fail in each recursive call at different
+  // points. Don't recurse here to avoid an infinite loop in failure cases
```
brentleyjones, line 128: nit:
```suggestion
  // points. Don't recurse here to avoid an infinite loop in failure cases.
```

brentleyjones (approved):

brentleyjones:
Thanks @jerrymarino

PR #645, opened 2021-06-10
Fix flakey `MakeDir`

brentleyjones:
Another process could have concurrently created the same directory while MakeDir is recursing, so we need to check if the directory exists before returning `false`.


tools/worker/work_processor.cc
```diff
@@ -129,8 +129,10 @@ void WorkProcessor::ProcessWorkRequest(
       // incremental storage area.
       auto dir_path = Dirname(expected_object_pair.second);
       if (!MakeDirs(dir_path, S_IRWXU)) {
-        std::cerr << "Could not create directory " << dir_path << " (errno "
-                  << errno << ")\n";
+        stderr_stream << "swift_worker: Could not create directory " << dir_path
+                      << " (errno " << errno << ")\n";
+        FinalizeWorkRequest(request, response, EXIT_FAILURE, stderr_stream);
```
brentleyjones, line 134: @jerrymarino For your other PR, we also want this so we don't silently ignore the errors anymore. Now that it only errors when it truly fails to create the directory.

keith (approved):

PR #646, opened 2021-06-10
Delete the `swift_common.swift_runtime_linkopts` function.

segiddins:
This should have been deleted in the previous change that refactored the linker flags. It no longer works because it tries to access a provide field that no longer exists, but nothing broke because nothing calls it anymore. Yay weak typing.

PiperOrigin-RevId: 378673728
(cherry picked from commit 503c9cc8c116599c217ba0c6743998fbfa865f4b)

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -502,36 +502,6 @@ A `File` representing the precompiled module (`.pcm`) file, or `None` if
   the toolchain or target does not support precompiled modules.
 
 
-<a id="#swift_common.swift_runtime_linkopts"></a>
-
-## swift_common.swift_runtime_linkopts
-
-<pre>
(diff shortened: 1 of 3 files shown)


keith (approved):
doc updates looks like

jerrymarino:
Ah edit: I see in the upstream - I think we'll need to get this in: https://github.com/bazelbuild/rules_apple/blob/upstream/apple/internal/swift_support.bzl

allevato:
You'll need to cherrypick https://github.com/bazelbuild/rules_apple/commit/c253e4fa18d6f95ac065021b28662c4ddb37b149 from the upstream branch in rules_apple (it's actually required for the change that preceded this one, which has the same commit message since they were a single change in our monorepo).

jerrymarino:
Oh nice, thanks for clarifying that 👍 - I'm still getting up-to-speed on the new workflow, makes it really easy since they've got the same commit message!

PR #647, opened 2021-06-10
Refer to build_bazel_rules_swift_index_import via a file rather than …

segiddins:
…inline string

For consistency with other repos that do not define their own build files

### DIFF
--- a/third_party/build_bazel_rules_swift_index_import/BUILD.overlay
+++ b/third_party/build_bazel_rules_swift_index_import/BUILD.overlay
@@ -1,0 +1,9 @@
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+native_binary(
+    name = "index_import",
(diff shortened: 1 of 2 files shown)


keith (approved):

third_party/build_bazel_rules_swift_index_import/BUILD.overlay
```diff
@@ -0,0 +1,9 @@
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+native_binary(
+    name = "index_import",
+    src = "index-import",
+    out = "index-import",
+)
```
thii, line 9: Can't this be just `exports_files`? Why do need to do a copy here?

segiddins: to make it an executable that can be used as an executable attribute label, iirc

thii (approved):

PR #649, opened 2021-06-11
Fix index_import label

keith:
Not sure why this works sometimes and not others but we need to
terminate this where the BUILD file lives


segiddins:
My bad!

keith:
np, not sure why it worked here but not for us

PR #653, opened 2021-06-15
Truly silence warnings about using -wmo with -dump-ast

segiddins:
We need an early return to completely avoid consuming the WMO-enabling arg


tools/worker/swift_runner.cc
```diff
@@ -327,7 +327,7 @@ bool SwiftRunner::ProcessArgument(
       } else if (is_dump_ast_ && ArgumentEnablesWMO(arg)) {
         // WMO is invalid for -dump-ast,
         // so omit the argument that enables WMO
-        changed = true;
+        return true;  // return to avoid consuming the arg
```
keith, line 330: maybe `continue` instead?

segiddins: I don't think we're inside a loop context here

PR #655, opened 2021-06-21
Fix breaking rpaths with swift_test on macOS

keith:
If your binary gets an rpath like `@loader_path/../foo`, when we moved
this into the xctest bundle directory that path would become invalid.
Bazel already does rewrite these paths as needed, but since the link
action doesn't write to the xctest bundle it's still wrong. Ideally we
could output there directly but there is a bazel issue with conflicting
prefix paths when you try to do that because of params file usage.


brentleyjones (approved):

keith:
I might be able to fix this with https://github.com/bazelbuild/bazel/commit/b4cc44c978bc0e25b6652b66018b6aad12bff820 now

keith:
replaced by https://github.com/bazelbuild/rules_swift/pull/678

Issue #516, opened 2020-11-04
Investigate using json for worker


keith:
This is supported now and would reduce our usage of proto which can take a while to build on its own. We already have a json parsing library in the worker for output file map manipulation 

https://github.com/bazelbuild/bazel/issues/7998#issuecomment-721796702

arturdryomov:
@keith, do you have ideas about how stable the JSON support gonna be? It would be nice to know or have some insight about whether it will be removed from Bazel itself. I would like to switch from PB to JSON in Detekt rules but not sure if it’s a good idea to require setting a Bazel execution flag (`--experimental_worker_allow_json_protocol`) from clients. TBH it’s tempting since compiling PB from scratch is a bit ridiculous.

keith:
It was just added so I'm optimistic it will stick around

arturdryomov:
@keith, did some digging in buildfoundation/bazel_rules_detekt#107.

keith:
Thanks for the link

PR #657, opened 2021-06-23
Convert worker to use json instead of protobuf

keith:
This was complicated by the difficulty in reading from stdin since it's
populated asynchronously from a file while the worker is running.

Fixes https://github.com/bazelbuild/rules_swift/issues/516


tools/worker/compile_with_worker.cc
```diff
@@ -76,40 +74,34 @@
...
+      line.append(buffer);
+    } while (result == 1024);
+
+    if (line == "") {
+      continue;
     }
 
-    blaze::worker::WorkResponse response;
+    auto request_json = nlohmann::json::parse(line);
+    WorkRequest request(0, request_json["arguments"]);
```
keith, line 101: This 0 is because it doesn't seem that we get requestID, maybe that requires supporting multiplexing, maybe we should still read and default or something

tools/worker/compile_with_worker.cc
```diff
@@ -76,40 +74,35 @@
...
-      std::cerr << "Could not read WorkRequest from stdin. Killing worker "
-                << "process.\n";
-      return 254;
+    std::string line;
+    int result;
+    do {
+      char buffer[1024];
+      lseek(STDIN_FILENO, offset, SEEK_SET);
+      result = read(STDIN_FILENO, buffer, 1024);
+      buffer[result] = '\0';
```
segiddins, line 91: couldnt this potentially be writing out of bounds of the array, if the full 1024 bytes are read?

tools/worker/compile_with_worker.cc
```diff
@@ -76,40 +74,35 @@
...
+      offset += result;
+      line.append(buffer);
+    } while (result == 1024);
```
segiddins, line 94: i'm wondering if it's possible we accidentally read in more than 1 message's worth during a single iteration, and then parsing fails? or does bazel guarantee that no request will be sent until the response is received?

tools/worker/compile_with_worker.cc
keith: good question need to test

tools/worker/compile_with_worker.cc
keith: yea that's a good question, I _think_ it's line by line but haven't stress tested it

.bazelrc
```diff
@@ -1,3 +1,5 @@
+build --experimental_worker_allow_json_protocol
```
arturdryomov, line 1: Isn’t the protocol stable [since 4.1](https://blog.bazel.build/2021/05/21/bazel-4-1.html)? If so you might not need the flag. However, all 4.0 users will need to declare the flag.

tools/worker/compile_with_worker.cc
```diff
@@ -76,40 +74,35 @@
...
-    if (!google::protobuf::util::SerializeDelimitedToZeroCopyStream(
-            response, &file_output_stream)) {
-      std::cerr << "Could not write WorkResponse to stdout. Killing worker "
-                << "process.\n";
-      return 254;
-    }
-    // Flush stdout after writing to ensure that Bazel doesn't hang waiting for
-    // the response due to buffering.
-    file_output_stream.Flush();
+    std::cout << response.to_json().dump() << std::flush;
```
arturdryomov, line 105: Tests? 🙈

.bazelrc
keith: still not in 5.x

brentleyjones:
Next 5.0rc will have those `latest-bazel` checks go green.

keith:
closing for https://github.com/bazelbuild/rules_swift/pull/801

PR #663, opened 2021-08-05
Fix WMO

brentleyjones:
`is_dump_ast_` was defulting to `true`, causing WMO not to be applied.


tools/worker/swift_runner.cc
```diff
@@ -108,6 +108,7 @@ static bool StripPrefix(const std::string &prefix, std::string &str) {
 SwiftRunner::SwiftRunner(const std::vector<std::string> &args,
                          bool force_response_file)
     : force_response_file_(force_response_file) {
+  is_dump_ast_ = false;
```
keith, line 111:
```suggestion
    : force_response_file_(force_response_file), is_dump_ast_(false) {
```

segiddins (approved):
Ooof, that's my bad

PR #669, opened 2021-08-10
Remove the `swift.index_while_building` feature.

thii:
This should not be the responsibility of the build rules, and removing it reduces complexity. There are better approaches:

*   For completely local builds, just pass in the indexing flags as command line swiftcopts and multiple compilation actions can share the same index store location (excluding it from sandboxing, if necessary).
*   For remote builds, use an aspect to construct indexing actions based on the `SwiftCompile` actions and run those in parallel.

PiperOrigin-RevId: 378535849
(cherry picked from commit 7b40da188d7bea5dbdda15f95b3419aef0cf6a71)

 Conflicts:
	swift/internal/compiling.bzl
	swift/internal/derived_files.bzl
	swift/internal/feature_names.bzl


keith:
We hadn't talked about this one for a while but I was leaning towards not taking it until a compelling alternative exists. I think all sophisticated consumers of the rules are using this today, so we can't break it with no other option

thii:
Skipping this for now—there are still users of it outside of Google.

PR #671, opened 2021-08-11
Cherry pick upstream

thii:


thii:
`swift_interop_hint` related changes require the new `aspect_hints` attribute which is not available in Bazel 4.x.

segiddins (approved):

PR #673, opened 2021-08-12
Collect linking context from `swift_{binary,test}` if they have `srcs`.

thii:
This is required for the `autolink_extract` to extract linker flags from
the import statements from those source files.

### DIFF
--- a/swift/internal/swift_binary_test.bzl
+++ b/swift/internal/swift_binary_test.bzl
@@ -15,6 +15,7 @@
 """Implementation of the `swift_binary` and `swift_test` rules."""
 
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
+load(":actions.bzl", "is_action_enabled", "swift_action_names")
 load(":compiling.bzl", "output_groups_from_other_compilation_outputs")
 load(":derived_files.bzl", "derived_files")
 load(":feature_names.bzl", "SWIFT_FEATURE_BUNDLED_XCTESTS")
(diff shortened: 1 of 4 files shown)


keith (approved):

brentleyjones (approved):

swift/internal/swift_binary_test.bzl
```diff
@@ -202,6 +203,25 @@ def _swift_linking_rule_impl(
...
+                label = ctx.label,
+                linking_contexts = [
+                    dep[CcInfo].linking_context
+                    for dep in ctx.attr.deps
+                    if CcInfo in dep
+                ],
+                module_context = module_context,
+                swift_toolchain = swift_toolchain,
+            )
+            additional_linking_contexts.append(linking_context)
```
thii, line 224: @allevato This was needed after the refactoring in https://github.com/bazelbuild/rules_swift/commit/0bbf2d3cbff41bbdfc8f07ff5d84cad45c4ce358. Does it not break with the upstream version? Or `swift_binary` and `swift_test` aren't supposed to have direct `srcs` anymore?

allevato: We don't use `autolink-extract` in our configuration, so that explains why it didn't cause any problems for us. The addition you've made here makes sense (with one question, below).

The object files from the direct `srcs` themselves are handled by passing `cc_compilation_outputs` into `register_link_binary_action` below, so I guess the only thing getting dropped here are the autolink directives for anything imported by those sources.

However, `create_linking_context_from_compilation_outputs` adds the `cc_compilation_outputs` passed to it to the linking context, and that linking context is also passed to `register_link_binary_action` along with the same compilation outputs again, so do you not see the object files being passed to the linker twice? I wonder if the ones from the linking context are being wrapped in `--start-lib/--end-lib` and it's ignoring them after it's already seen the symbols in the ones passed directly from the `cc_compilation_outputs`, which would be passed to the linker as just the .o files without any wrapping, to force them all to link.

I wonder if this linking context should be created unconditionally instead, with `alwayslink = True`, and then don't pass the compilation outputs to `register_link_binary_action` at all to avoid the possible duplication.

thii: Thanks for pointing that out. Indeed it would cause symbol duplication on Apple platforms if this was enabled unconditionally. Submitted a PR based on the suggestion https://github.com/bazelbuild/rules_swift/pull/677.

PR #685, opened 2021-09-13
Add workaround for SR-15123

keith:
This works around an issue in Swift 5.5 / Xcode 13.0 beta 5 where
absolute paths to the sources being compiled are _always_ used and
embedded in the case a vfs overlay is used. This tricks the compiler by
using a vfs overlay for the source files as well.

https://bugs.swift.org/browse/SR-15123

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -56,6 +56,7 @@ load(
     "SWIFT_FEATURE_OPT_USES_OSIZE",
     "SWIFT_FEATURE_OPT_USES_WMO",
     "SWIFT_FEATURE_REWRITE_GENERATED_HEADER",
+    "SWIFT_FEATURE_SOURCES_VFSOVERLAY",
     "SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION",
     "SWIFT_FEATURE_SUPPORTS_BARE_SLASH_REGEX",
     "SWIFT_FEATURE_SUPPORTS_LIBRARY_EVOLUTION",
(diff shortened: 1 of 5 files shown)


keith:
We might not need this pending https://github.com/apple/llvm-project/pull/3246 but it does appear to work (only with Xcode 13 though)

keith:
optimized a bit for revertability here

thii (approved):

acecilia:
I just hit this issue trying to update xcode from `12.5.1` to `13.2.1`. In my case, what I observed is that `#fileID` prints an absolute path that ends up inside the binaries. Before finding this PR I did write a simple reproducible example [here](https://github.com/acecilia/bazel-tests/tree/main/file_id)

Thanks for the multiple attempts to fix it in different repositories @keith 🙏 

How are you currently dealing with this problem? I am not sure how to work around it considering that:
* The bug is still present in the latest Xcode `13.2.1`
* This PR is a draft, did not make it to `master`

Some options on the top of my mind:
* Wait for swift `5.5.X`, which hopefully comes with a fix (this is probably not a good idea, as the fix may take long time to land)
* Stop using vfs overlay
* Maintain my own clone of `rules_swift` and add this PR to it

brentleyjones:
Instead of maintaining a fork, you can apply a patch with bazel: https://brentley.dev/patching-bazel-external-dependencies/

keith:
Yep that's what I would recommend in general, and what we're doing at Lyft. I felt a bit weird about merging this at the time because I wasn't sure if it would have any unexpected side-effects. Realistically we could probably merge it now, but I was hoping that my actual fixes upstream would land soon enough that we wouldn't need too. Unfortunately it looks like they won't be fixed until Swift 5.6, which I assume won't come to Xcode until the spring

acecilia:
Got it, thank you 🙌

brentleyjones:
The linked SR is closed. Is it still a valid bug that should be reopened?

keith:
needs more investigation, i think there's a new one

keith:
as of Xcode 14 we have confirmed this is fixed, and don't see any new reason to use this at the moment. I believe it was fixed earlier, with Swift 5.6, but I didn't test before Swift 5.7. Please file an issue if you find any new case that needs this!

PR #691, opened 2021-09-24
Add `aspect_hints` feature

keith:
This includes all the upstream commits related to this, which is blocked on bazel's default for this feature


brentleyjones (dismissed):

keith:
blocked on 5.x

brentleyjones:
https://github.com/bazelbuild/bazel/commit/47bd65adf6fd968c9eb84f96044050d852750d62

thii:
Many other changes rely on this now. Let's enable the flag and mention about this requirement in the release notes.

brentleyjones:
I'm not sure we want to force that flag on all users of rules_swift. Let's see if @allevato has any additional thoughts.

allevato:
> I'm not sure we want to force that flag on all users of rules_swift. Let's see if @allevato has any additional thoughts.

The only real advice I can offer is to bring up to the Bazel team the difficulty the experimental nature of the feature poses to integrating upstream rules_swift changes (especially as they stack on each other). Unfortunately the decision to make the flag experimental in Bazel was made some time after the feature was implemented and the Swift rules had started adopting it...

luispadron:
Dang i think this is blocking cherry-picking https://github.com/bazelbuild/rules_swift/commit/bfcae19af93924651911e6bdc669608e9453a774 as well.

PR #694, opened 2021-10-04
Retrieve single arch cpu from `cc_toolchain` instead of from `apple_fragment`

thii:
`apple_fragment.single_arch_cpu` was returning the default macos cpu for
the host platform for tools while it should return the cpu of the
execution platform.

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -635,13 +635,44 @@ def _xcode_env(xcode_config, platform):
         apple_common.target_apple_env(xcode_config, platform),
     )
 
+# TODO: Revisit ctx.fragments.apple.single_arch_cpu to see if it returns the
+# correct cpu for the exec platform.
+def _single_arch_cpu(cc_toolchain, platform):
+    """Returns the single effective architecture for the current configuration.
+
+    Args:
+        cc_toolchain: The C++ toolchain from which linking flags and other
+            tools needed by the Swift toolchain (such as `clang`) will be
+            retrieved.
+        platform: The `apple_platform` value describing the target platform
+            being built.
+
+    Returns:
+        A single architecture without the platform name (e.g. `x86_64`).
+    """
+    cpu = cc_toolchain.cpu
+    platform_type = str(platform.platform_type)
+    if platform_type == "ios":
+        if cpu.startswith("ios_sim_"):
+            return cpu[len("ios_sim_"):]
+        else:
+            return cpu[len("ios_"):]
+    if platform_type == "watchos":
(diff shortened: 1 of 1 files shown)


brentleyjones:
Do you feel this is a bug with `apple_fragment.single_arch_cpu`?

brentleyjones (approved):

thii:
It could be, but I think it's safe to use this as it's the one cc_library and objc_library use.

keith:
I think now that this might be fixed by the change like this https://github.com/bazelbuild/rules_apple/pull/1154 we should try that first

brentleyjones:
Now that bazelbuild/rules_apple#1154 is merged, is this still an issue?

thii:
Yes, still. I think this is a Bazel issue, since rules_swift doesn't depend on rules_apple.

brentleyjones:
Yep, we have to use this patch as well until Bazel fixes it's stuff. Doesn't look like anyone is looking at it over there so low chance of a 5.0 fix?

keith:
Can you test this with bazel HEAD in general? A quick look at the bazel source makes it look like the defaults read from https://github.com/bazelbuild/bazel/blob/c873525af9e045c87dfca2ac82ca68e9b76f8820/src/main/java/com/google/devtools/build/lib/rules/apple/AppleConfiguration.java#L82 and https://github.com/bazelbuild/bazel/blob/c873525af9e045c87dfca2ac82ca68e9b76f8820/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java#L201-L203

thii:
Looks like it has been fixed at https://github.com/bazelbuild/bazel/commit/c873525af9e045c87dfca2ac82ca68e9b76f8820.

Do you have any idea which commit fixes this? I think it should be cherry-picked into 5.0.

thii:
Interesting that the baseline of 5.0 doesn't have this issue, but 5.0.0rc2 does.

thii:
Sorry, above two comments were incorrect. I was testing it wrong due to a bug in bazelisk - it is pulling x86_64 binaries for unreleased commit on M1 https://github.com/bazelbuild/bazelisk/issues/260

Update: It has not been fixed at HEAD.

brentleyjones:
I saw we merge this, to fix the issue, and revert when Bazel gets a fix.

thii:
Does no one else have this problem? We're still not able to build with universal tools without this patch.

brentleyjones:
Yes, we need this patch to build universal tools. My comment still stands, I think we should merge it for now.

keith (approved):

thii:
Thanks 🙏
I'm taking this in for now

PR #696, opened 2021-10-05
swift_autoconfiguration: handle case where `CC` isn't set

dflems:
This was added in #683. `CC` may not be set in the env, in which case it will be `None`, which will fail with an error like `TypeError: argument of type 'NoneType' is not iterable`

It was going to fail anyway, but now it will fail with the right error.


dflems:
@keith

keith (approved):
Nice thanks

PR #699, opened 2021-10-14
Use newer split module flag with Swift 5.4

brentleyjones:
Uses the newer `-experimental-skip-non-inlinable-function-bodies-without-types` which was introduced here: https://github.com/apple/swift/pull/34612. This should improve LLDB usage in some cases.

When using WMO, it has two downsides in Swift 5.5, both introduced by https://github.com/apple/swift/pull/38939:
- The swiftmodules will have swiftdeps info embedded in them, which is only needed for incremental compilation
- Interface hashing is enabled, which again is only needed for incremental compilation

This is because the swift compiler only expects `-experimental-skip-non-inlinable-function-bodies-without-types` to be used with the new `-emit-module-separately` incremental build. In the same vein, in order to support split modules with incremental builds, we need to use this flag (though we could choose to use this version _only_ with incremental + split modules, if we decide it's not worth the above regressions).


brentleyjones:
I'm not sure how to make the test differ based on Xcode version used.

PR #701, opened 2021-10-19
Persist pch files in output dir

jerrymarino:
Introduce a feature to persist pch files created internally. This simply
adds the flag pch_output_dir to point at a directory when the feature is
present


jerrymarino:
Swift compiles a pch in for some cases and you can enable this flag to persist them amongst different invocations. For a test you mean asserting the flag is passed w/ the configurator - something like `test/module_cache_settings_tests.bzl`?

keith:
yea something like that is good

keith:
Thanks for the PR!

Issue #487, opened 2020-09-13
Building a small code change is slow


acecilia:
Hey 😃 

I am playing a bit with building an app with bazel. I measured how long it takes in my machine to rebuild the app when I introduce a one-line change in the appDelegate. The methodology is:
1. I build the app clean.
2. I uncomment a `print("hello")` in the `AppDelegate` of the app.
3. I build the app again without cleaning, and measure how long it takes to build this small change.

These are my results:
* Scenario 1, using Xcode: **4.771s**
* Scenario 2, using bazel with `swift_copts = [
        "-whole-module-optimization",  "-swift-version", "5", "-DDEBUG",  "-Onone", "-g"]` and a basically empty `.bazelrc` file: **22.5s**
* Scenario 3, using bazel with `swift_copts = ["-swift-version", "5", "-DDEBUG",  "-Onone", "-g"]` and [`build --strategy=SwiftCompile=worker`](https://github.com/bazelbuild/rules_swift/pull/174) in my `.bazelrc` file: **21.5s**

Profiling bazel for the latest scenario (scenario 3) I see:
* **9s** of `compiling swift module`
* **3s** of `linking`
* **0.8s** of `SwiftStdlibCopy`
* **1.3s** of `bundling`
* **9s** of `processing and signing App`

Looking at these numbers I have some questions, if somebody can help out:
1. Looking at the configuration I mentioned, am I doing something wrong?
2. Are those build time numbers I mention somehow expected? Seems quiet a big difference
3. Is there any other configuration I should setup in order to get faster builds when introducing small code changes?

My goal is to use bazel for active development (not only CI), but this build time difference can become an issue.

Thanks!

acecilia:
Related blog: https://brentley.dev/xcodebuild-vs-bazel-incremental/

brentleyjones:
I would say the biggest issue you were probably running into was the module cache, and the overhead of sandboxing.

acecilia:
@brentleyjones Thanks so much for your writeup and shining some light on this, much appreciated! 🙌

keith:
I wonder if we should flip the default on the module cache 🙃

brentleyjones:
I would be in favor of that. It more aligns with Xcode, and if you have sandboxing enabled you still get a non-shared cache.

jerrymarino:
@acecilia it'd be interesting to see some output from perf runs of your [build](https://docs.bazel.build/versions/master/skylark/performance.html#performance-profiling)

[...]

PR #702, opened 2021-10-26
Change module cache default to be global

keith:
The performance impact of using the global module cache is so
significant for projects with multiple swift_library targets, and the
crashes seem to be so rare, that I think this is the right tradeoff for
the default.

Also discussed on https://github.com/bazelbuild/rules_swift/issues/487


brentleyjones (approved):

keith:
https://github.com/bazelbuild/rules_swift/pull/956#event-8082296945

PR #711, opened 2021-11-05
expose protoc-gen-grpc-swift as a gRPC generator

crockeo:
Up until now rules_swift only exposed `protoc-gen-swiftgrpc`, the 0.x gRPC generated for Swift. This PR adds support for the 1.x gRPC generator, `protoc-gen-grpc-swift`. `protoc-gen-grpc-swift` generates code that depends on `GPRC` and `NIO`, so most of the code in this PR is setting up builds for `GRPC` and `NIO`. Not sure if this needs to live in `rules_swift`, but it's good enough for now 🙂 

Some remaining todos:

- [ ] Audit all of the additional builds to ensure they have their required dependencies (cross-reference w/ upstream builds)
- [ ] Create builds for CNIOWindows and CNIOLinux 
- [x] Fix issue where generated `.grpc.swift` files can't compile because they don't import their accompanying `.pb.swift`
- [x] Only compile the relevant set of support libraries (`SwiftGRPC` for `swiftgrpc`, `GRPC`/`NIO` for `grpc-swift`)


crockeo:
@googlebot I signed it!

swift/internal/swift_grpc_library.bzl
```diff
@@ -249,6 +264,17 @@ def _swift_grpc_library_impl(ctx):
...
 
+    # grpc-swift depends on SwiftProtobuf to determine which modules need to be imported
+    # into the generated .grpc.swift file for the generated .gRPC.swift file
+    #
+    # SwiftProtobuf explicitly disallows the current .proto file's module name
+    # but when generating the .grpc.swift and .pb.swift files to different locations
+    # one MUST import the current .proto file's module name
+    if ctx.attr.protoc_plugin_name == "grpc-swift":
+        proto_module_name = deps[0][SwiftInfo].direct_modules[0].name
+        extra_module_imports.append(proto_module_name)
```
crockeo, line 275: this seems like a hack that i don't want to commit to rules_swift. we should have a better way to determine if we want to exclude the current module or not, but that involves at least two upstream tickets (against [grpc-swift](https://github.com/grpc/grpc-swift) and [swift-protobuf](https://github.com/apple/swift-protobuf)).

brentleyjones:
The proto rules are completely different now, so this would have to be reworked at a minimum.

PR #712, opened 2021-11-11
Add a flag to allow tools to be compiled as universal binaries

thii:
This adds a new boolean flag
`--@build_bazel_rules_swift//swift:universal_tools` that when used will
force all the tools used in this repository to be compiled into
universal binaries. This allows execution running on different macOS
platforms (`arm64` and `x86_64`) to be compatible with each other and
can run natively on both.

### DIFF
--- a/third_party/com_github_apple_swift_protobuf/BUILD.overlay
+++ b/third_party/com_github_apple_swift_protobuf/BUILD.overlay
@@ -1,4 +1,8 @@
 load(
+    "@build_bazel_apple_support//rules:universal_binary.bzl",
+    "universal_binary",
+)
+load(
     "@build_bazel_rules_swift//swift:swift.bzl",
     "swift_binary",
     "swift_library",
@@ -30,4 +34,19 @@ swift_binary(
     ]),
     visibility = ["//visibility:public"],
     deps = [":SwiftProtobufPluginLibrary"],
+)
+
+universal_binary(
+    name = "universal_ProtoCompilerPlugin",
+    binary = ":ProtoCompilerPlugin",
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "ProtoCompilerPlugin_wrapper",
+    actual = select({
+        "@build_bazel_rules_swift//swift:universal_tools_config": ":universal_ProtoCompilerPlugin",
+        "//conditions:default": ":ProtoCompilerPlugin",
+    }),
+    visibility = ["//visibility:public"],
 )
--- a/swift/BUILD
+++ b/swift/BUILD
@@ -1,5 +1,5 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
-load("@bazel_skylib//rules:common_settings.bzl", "bool_setting")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "bool_setting")
 load(
     "//swift/internal:build_settings.bzl",
     "per_module_swiftcopt_flag",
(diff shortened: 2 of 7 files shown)


thii:
Starlark flag is broken in Bazel 4.2.1 somehow - when testing inside this repo, `--//swift:universal_worker` works, but `--@build_bazel_rules_swift//swift:universal_worker` doesn't. Works fine either way with rolling releases.

thii:
Also submitted https://github.com/bazelbuild/apple_support/pull/95 if the rule should better be there.

brentleyjones:
I like the rule living in `apple_support`.

.bazelci/update_workspace_to_deps_heads.sh
```diff
@@ -39,7 +39,7 @@ git_repository(\
 git_repository(\
 \    name = "build_bazel_apple_support",\
 \    remote = "https://github.com/bazelbuild/apple_support.git",\
-\    branch = "master",\
+\    commit = "afc32d589c0db680c855dbab383cf3366b1ba041",\
```
thii, line 42: to be reverted once https://github.com/bazelbuild/apple_support/pull/95 is merged

third_party/com_github_apple_swift_protobuf/BUILD.overlay
```diff
@@ -1,3 +1,7 @@
+load(
+    "@build_bazel_apple_support//lib:universal_binary.bzl",
```
BalestraPatrick, line 2: Since the rule was moved to `rules` in https://github.com/bazelbuild/apple_support/commit/a448644dce247e59268864fd66b6578195412b59, maybe we can already bump to that version and make the breaking change in this PR?

brentleyjones (commented):
I somehow wrote some comments 2 months ago and never submitted them. Better late than never?

.bazelci/presubmit.yml
```diff
@@ -33,8 +33,7 @@ x_defaults:
 tasks:
   macos_latest:
     name: "Latest Bazel"
-    # TODO(#722): Change back to `latest` after 5.0 is released
-    bazel: last_rc
```
keith, line 37: was this meant to be in a separate change?

.bazelci/presubmit.yml
```diff
@@ -33,8 +33,7 @@ x_defaults:
...
+    bazel: latest
```
brentleyjones, line 36: Seems these snuck in?

swift/repositories.bzl
```diff
@@ -51,9 +51,10 @@ def swift_rules_dependencies():
         http_archive,
         name = "build_bazel_apple_support",
         urls = [
-            "https://github.com/bazelbuild/apple_support/releases/download/0.11.0/apple_support.0.11.0.tar.gz",
+            "https://github.com/bazelbuild/apple_support/archive/a448644dce247e59268864fd66b6578195412b59.zip",
```
keith, line 54: I'll cut a release of this that i can bump to before we merge instead

keith (commented):
i think we should strive to enable this by default, but I don't think we can right now and support old-ish xcodes. Also with protoc that might be annoying, but if we can drop that for json workers it should be light enough to in the future.

.bazelci/presubmit.yml
thii: oops, rebased!

.bazelci/presubmit.yml
thii: it was from my local master, just rebased now

brentleyjones:
Since we probably have to support older Xcodes, maybe opt out for now?

keith:
i just worry about folks being even more annoyed about protoc taking forever to build if they're not using a cache

keith (approved):

PR #713, opened 2021-11-15
Ensure quote_includes are set for swift_library's CcInfo

steeve:
This mimics the behavior of `cc_library`.

When consuming `CcInfo` from a `swift_library`, if the library is in an external
repo, the `quote_includes` paths are not set.
In the case of a cc_library, both the workspace root and the `bindir` +
`workspace_root` are set.

If the workspace_root isn't set (in case of a non external repo), include the
current sandbox's path `.`.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -84,7 +84,7 @@ A new attribute dictionary that can be added to the attributes of a
 
 <pre>
 swift_common.compile(<a href="#swift_common.compile-actions">actions</a>, <a href="#swift_common.compile-feature_configuration">feature_configuration</a>, <a href="#swift_common.compile-module_name">module_name</a>, <a href="#swift_common.compile-srcs">srcs</a>, <a href="#swift_common.compile-swift_toolchain">swift_toolchain</a>,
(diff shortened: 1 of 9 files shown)


thii:
I think `cc_library` (and `objc_library`)'s behavior was a regression since Bazel 3.0, since it adds all external dependencies' workspace paths into `includes` even though not all of them are required. This can slow down the build a lot of you have a lot of external repositories. I thought I've filed a bug for it before but couldn't find it now, but I still have a reproduce example:

https://github.com/thii/objc-external-repos-quote-includes/blob/9a9545eaf6da650c6e46d82d2bc986e3e329298f/objc.diff

steeve:
Is it a regression though? It makes sense to do this since, if not, include paths will be different wether the code is compiled from the main workspace or via an external one.
It would also need to be included as `external/repo1/path/to/the/pkg/include.h`.

thii:
Why I think it was a regression is because the change degrades the build times (but that was a very long time ago and I don't remember the details now).

Is this something you can't do in your target declarations?

PR #715, opened 2021-11-16
Update rules_proto

thii:
Newer rules_proto updates `protobuf` which provides prebuilt binaries of
`protoc` for macOS arm64 (although on Maven only for now).


brentleyjones (approved):

thii:
I was wrong—the arm64 binaries on Maven are just copies of x86_64 binaries with different names.

brentleyjones (approved):

PR #716, opened 2021-11-18
Use `progress_message` placeholders for more performant rules

thii:
From a tip in a BazelCon talk today.

### DIFF
--- a/swift/internal/autolinking.bzl
+++ b/swift/internal/autolinking.bzl
@@ -83,8 +83,6 @@ def register_autolink_extract_action(
         feature_configuration = feature_configuration,
         outputs = [autolink_file],
         prerequisites = prerequisites,
-        progress_message = (
-            "Extracting autolink data for Swift module {}".format(module_name)
-        ),
+        progress_message = "Extracting autolink data for Swift module %{label}",
         swift_toolchain = swift_toolchain,
     )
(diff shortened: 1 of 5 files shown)


brentleyjones:
This only works on 5.0+, correct?

allevato:
FWIW, this changes the message so that it prints the target label instead of the Swift module name; whether that's an improvement or not is debatable.

I also doubt that anyone has Swift builds so large that you'd see measurable memory improvements out of this; the real users who benefit are rules like C++ which create an action per _source file_ (vs. per _target_).

It would be nicer if they gave Starlark an API as rich as the native Bazel one, which would let us separate the constant part of the format string from the placeholder substitution, but still use the module name instead of one of the blessed label/input/output values. I would wait for something like that before jumping on this.

thii:
> This only works on 5.0+, correct?

Yes this was added recently, only available since 5.0.

> FWIW, this changes the message so that it prints the target label instead of the Swift module name; whether that's an improvement or not is debatable.

I was aware about this limitation. I personally think that printing the target label would be more informative than printing just the Swift module name, especially when it comes from an external repository:

```
    Compiling Swift module @com_github_apple_swift_protobuf//:SwiftProtobuf; 2s local
```

brentleyjones:
>I personally think that printing the target label would be more informative than printing just the Swift module name, especially when it comes from an external repository

💯

allevato:
> I personally think that printing the target label would be more informative than printing just the Swift module name, especially when it comes from an external repository

That's reasonable!

brentleyjones (approved):
Now that we are targeting 5.0, this is safe to go in.

PR #723, opened 2021-11-23
Revert "Remove top level functions for now"

brentleyjones:
This reverts commit b1cc2123c44d9763e5e7fb902181cff0dfa87fc8.

### DIFF
--- a/swift/internal/module_maps.bzl
+++ b/swift/internal/module_maps.bzl
@@ -14,19 +14,9 @@
 
 """Logic for generating Clang module map files."""
 
-# TODO: Once bazel supports nested functions unify it with upstream
-def _add_headers(*, headers, kind, content, relative_to_dir, back_to_root_path):
-    # Each header is added to the `Args` object as a tuple along with
-    # `relative_to_dir` and `back_to_root_path`. This gives the mapping
-    # function the information it needs to relativize the header paths even
-    # when they're expanded from a tree artifact (and thus not known at
-    # analysis time).
-    content.add_all(
-        [(file, relative_to_dir, back_to_root_path) for file in headers],
-        format_each = '    {} "%s"'.format(kind),
-        map_each = _header_info_mapper,
-    )
-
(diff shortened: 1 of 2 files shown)


brentleyjones:
Seems like buildifier doesn't handle the inline functions correctly.

swift/internal/module_maps.bzl
```diff
@@ -91,18 +81,33 @@ def write_module_map(
...
+        return [
+            _header_path(
+                header_file = file,
+                relative_to_dir = relative_to_dir,
+                back_to_root_path = back_to_root_path,
+            )
+            for file in directory_expander.expand(file_or_dir)
+        ]
+
+    def _add_headers(*, headers, kind):
```
thii, line 94: Not sure if you're already aware of but the latest of this function in the upstream branch has then been changed a bit (but we can update it in a follow-up PR).

https://github.com/bazelbuild/rules_swift/blob/c8b58b504842b6c16c2791e78db1137204f74a6d/swift/internal/module_maps.bzl#L103

thii (approved):

swift/internal/module_maps.bzl
brentleyjones: Was it changed in another commit that we haven't cherry-picked yet? I would rather grab it then.

thii: https://github.com/bazelbuild/rules_swift/commit/868a76bed25439cf423871cc0803f58bdbf20d89 (already cherry-picked)

I manually fixed that with https://github.com/bazelbuild/rules_swift/commit/589e18fba0c21a702e1fe0e4b4bcc092c9672733

brentleyjones: I'm matching https://github.com/bazelbuild/rules_swift/commit/868a76bed25439cf423871cc0803f58bdbf20d89 so far. Any other changes, like https://github.com/bazelbuild/rules_swift/commit/ef6662cec1fdb7731f2c11004a0831fc8b4faa33, I'll get when we cherry-pick those.

PR #726, opened 2021-11-24
Remove uses of `incompatible_use_toolchain_transition` now that it is enabled by default.

brentleyjones:
This is a step towards removing it entirely.

PiperOrigin-RevId: 406230375
(cherry picked from commit d1ba795ff479881bf14ba70294693dcabace4fd4)

### DIFF
--- a/swift/internal/swift_toolchain.bzl
+++ b/swift/internal/swift_toolchain.bzl
@@ -511,6 +511,5 @@ The version of XCTest that the toolchain packages.
     doc = "Represents a Swift compiler toolchain.",
     fragments = [] if bazel_features.cc.swift_fragment_removed else ["swift"],
     toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
-    incompatible_use_toolchain_transition = True,
     implementation = _swift_toolchain_impl,
 )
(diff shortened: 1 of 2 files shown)


thii:
I think removing this was postponed (at least externally).

brentleyjones:
The flag has been flipped, and we only support 5.0 and above now, so I thought it's on the table again? This PR can just sit here if it's not the time yet.

thii (approved):

thii:
Ahh, right. Let's get this in.

brentleyjones:
We could wait until the flag doesn't exist anymore, to prevent issues for users that somehow need the flag in the other state. Maybe that was the end decision of delaying this externally.

keith (approved):

brentleyjones:
@Keith Thoughts on merging now, or waiting until the flag it dropped entirely (i.e. 6.0)?

keith:
it's unnecessary with 5.x right? if so i agree we should merge now since we already broke 4.x compat

brentleyjones:
A user of Bazel 5.0 might set `--incompatible_override_toolchain_transition=false`, changing away from the default, and then rules_swift would stop working for them.

keith:
hm yea interesting case. i wonder how prevalent that will be. no strong pref since this change likely won't help from a conflicts perspective, so there's no rush to merge it IMO (until last_green breaks compat maybe)

brentleyjones:
>until last_green breaks compat maybe

Yeah, we will have to make a call then. Probably worth merging at that time over forking 5.0 from 6.0.

segiddins (approved):

brentleyjones:
`--incompatible_override_toolchain_transition` is now a no-op in 6.0, but still available in 5.x. We can merge this once we break 5.x compatibility.

mattrobmattrob (approved):

PR #736, opened 2022-01-11
Fix exit code for swiftc crashes

keith:
In the case that swiftc crashed, the worker would return that it exited
0. This was because calling WEXITSTATUS when WIFEXITED is not true is
invalid and happened to return 0 in that case. We now handle the case
where swiftc exits from a signal, and return the signal it exited with
instead. There might be some other cases here that we have to handle in
the future, so for now those are failures so we don't return an invalid
value.

Also see https://github.com/bazelbuild/bazel/issues/14543

### DIFF
--- a/tools/common/process.cc
+++ b/tools/common/process.cc
@@ -173,11 +173,16 @@ int RunSubProcess(const std::vector<std::string> &args,
       return wait_status;
     }
 
-    int exit_status = WEXITSTATUS(status);
-    if (exit_status != 0) {
-      return exit_status;
+    if (WIFEXITED(status)) {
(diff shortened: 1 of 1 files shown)


keith:
For future reference crashes can be tested by adding this to a swift target:

```
copts = [
    "-Xfrontend",
    "-debug-crash-immediately",
],
```

keith:
Also note in today's worker implementation the linked bazel issue is only a problem for non-incremental builds, as incremental builds fail earlier when the expected outputs aren't able to be copied to the temporary directory. That doesn't affect this fix, but was part of my investigation

brentleyjones (approved):

PR #741, opened 2022-01-21
Add SWIFT_AVOID_WARNING_USING_OLD_DRIVER to environment

keith:
Since
https://github.com/apple/swift/commit/e24eef4b35290135634c11a9a83e652de1e4919d
Swift now emits a warning about the old driver being deprecated. If you
enable the old driver flag we can assume you are doing that
intentionally for now and know that isn't a forever solution.

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -534,6 +534,8 @@ def _all_tool_configs(
         env = dict(env)
         env["TOOLCHAINS"] = custom_toolchain
 
+    env["SWIFT_AVOID_WARNING_USING_OLD_DRIVER"] = "1"
+
     tool_config = swift_toolchain_config.driver_tool_config(
         driver_mode = "swiftc",
         env = env,


keith:
alternatively we could force users to set this, but I think it's a pretty low value warning if you're already opting into the disable feature intentionally

brentleyjones:
Are users able to do this themselves with `--action_env` or otherwise? While I'm not super opposed to this, it just feels a little odd to automatically silence this for them.

keith:
Yea definitely ways to do it yourself, it just doesn't seem like a helpful warning, no strong pref

thii:
It's a deprecation warning, without it users wouldn't notice about the deprecation of the old driver after upgrading to Swift 5.6/Xcode 13.3.

keith:
I guess IMO it was implied that it was deprecated because that has been the general positioning in the rollout of the swift-driver in general, and since they flipped the default. 

Currently I don't think there's any other way to do this actually, because of https://github.com/bazelbuild/rules_swift/pull/502

thii (approved):

PR #751, opened 2022-01-27
Patch rules_proto to make `protoc` universal everywhere

thii:
The `protoc` executable is also used by native rules (e.g.
`proto_library`), and is referred to as `@com_google_protobuf//:protoc`.
The `protoc_wrapper` we previously used only worked for rules in
rules_swift but not elsewhere - the universal target has to have the
label `@com_google_protobuf//:protoc`.


thii:
Also submitted https://github.com/bazelbuild/rules_proto/pull/110 but a downside of that PR is that we would have to pass two Starlark flags `--@build_bazel_rules_swift//swift:universal_tools` and `--@rules_proto//proto:universal_tools` to make tools universal.

keith:
FWIW i think we will drop this soon now that 5.x++ should have json worker support without the flag

thii:
I think that is a different one? `@com_google_protobuf//:protobuf` vs `@com_google_protobuf//:protoc`

`@com_google_protobuf//:protoc` is a dependency of `swift_proto_library` and `swift_grpc_library`. Even if you switch to json worker, if you use any of these rules, you would still have to build protobuf and protoc.

keith:
I guess what I'm thinking is in that case you're less likely to have host tools that depend on this, and if you do you'll have to wrap it in the universal wrapper at a higher level anyways, so what we do here won't matter?

thii:
We use `swift_proto_library` and since protoc (and some others) is an implicit tool dependency, without patching rules_swift I don't see there is any other way to provide your own protoc.

https://github.com/bazelbuild/rules_swift/blob/cec0f85cd847ae3133783f45247b253b56f304af/swift/internal/swift_protoc_gen_aspect.bzl#L567-L576

keith (approved):
ah yea that use case makes sense. i guess this could break folks who are using a different version of proto if this patch fails to apply, but maybe that's unlikely

brentleyjones:
That is an interesting concern. Makes me feel like something should be upstreamed instead.

thii:
When https://github.com/protocolbuffers/protobuf starts releasing an macOS arm binary (I think it will from the next non-patch release), we can request rules_proto to make universal binary the default.

PR #756, opened 2022-02-01
Update checksum of all zip archives

thii:
GitHub renegerated these files which broke the builds.


mlequeux:
👍 Faster than I was :)

mlequeux (approved):

thii:
These are now back to the old checksums again, so this is not needed anymore.

brentleyjones:
And they said they would roll out the feature that caused this issue, without the checksum changing, in the future.

keith:
Should we swap these to tars since we use those for most things? Smaller blast radius

thii:
We use zips internally because I hadn't heard zips broken in the past :( . I think size doesn't that matter for most repos once they're cached in the repository cache; also zips usually get extracted faster.

I think either one is fine, but is it possible to have them mirrored in the Bazel mirror?

keith:
I imagine the perf doesn't have much impact for most small rules repos? I guess in general I thought most rules repos recommended tars for download in their READMEs though. To mirror I think you just have to open issues on the repos to have some googler push them (and make sure the URL is there)

PR #763, opened 2022-02-12
Add feature for partially explicit modules

keith:
As of Swift 5.6 and this PR https://github.com/apple/swift/pull/39887
the new json format used for entirely explicit module builds can be used
without entirely disabling implicit modules. This provides an
alternative to VFS overlays for discovering dependencies without `-I`
search paths. This is theoretically about the same but I expect this
file format to have better support in general than VFS overlays for
things like the new incremental compilation support, which currently
doesn't support VFS overlays, and this shouldn't have any downsides vs
VFS files.

This format does support some more keys than just the 2 we pass, as far
as I can tell they are unused today, so I'm opting not to pass them. We
may have to revisit this in the future.

This kind of cherry picks 1666670edadfe6e440b23d7142b0e21bf1043402 but I kept the VFS feature for now for legacy users.

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -166,6 +166,10 @@ SWIFT_FEATURE_REWRITE_GENERATED_HEADER = "swift.rewrite_generated_header"
 # them.
 SWIFT_FEATURE_USE_C_MODULES = "swift.use_c_modules"
 
+# If enabled, Swift modules for dependencies will be passed to the compiler
+# using a JSON file instead of `-I` search paths.
+SWIFT_FEATURE_USE_EXPLICIT_SWIFT_MODULE_MAP = "swift.use_explicit_swift_module_map"
+
 # If enabled, Swift compilation actions will use the same global Clang module
 # cache used by Objective-C compilation actions. This is disabled by default
 # because under some circumstances Clang module cache corruption can cause the
--- a/swift/internal/BUILD
+++ b/swift/internal/BUILD
@@ -45,6 +45,7 @@ bzl_library(
         ":debugging",
         ":derived_files",
         ":developer_dirs",
+        ":explicit_module_map_file",
         ":feature_names",
         ":features",
         ":providers",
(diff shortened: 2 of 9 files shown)


swift/internal/derived_files.bzl
```diff
@@ -316,6 +316,23 @@ def _vfsoverlay(actions, target_name):
     """
     return actions.declare_file("{}.vfsoverlay.yaml".format(target_name))
 
+def _explicit_module_map(actions, target_name):
+    """Declares a file for the explicit module map confoiguration for a
```
jszumski, line 320:
```suggestion
    """Declares a file for the explicit module map configuration for a
```

swift/internal/feature_names.bzl
```diff
@@ -202,6 +202,11 @@ SWIFT_FEATURE_USE_RESPONSE_FILES = "swift.use_response_files"
 # when access to those paths involves traversing a networked file system.
 SWIFT_FEATURE_VFSOVERLAY = "swift.vfsoverlay"
 
+# If enabled, Swift compilation actions will use a explicit module map file for
+# discovering dependencies. This is faster than using search paths when you
+# have many dependencies. This doesn't apply to system dependencies.
```
jszumski, line 207: This doesn't need to go in the comment, but what is the tipping point for "many"?

keith: Hard to say, for us at this point with >2k modules (not all in every dependency tree) this improves build performance by >50%, so if you have more than maybe 20 it's worth a try?

brentleyjones: Is there any downside to just using this always when on Swift 5.6+?

keith: not that I know of yet, I don't recall why we never defaulted the VFS overlay to on 🤔 , but I also haven't tested incremental with this

brentleyjones:
https://github.com/bazelbuild/rules_swift/commit/1666670edadfe6e440b23d7142b0e21bf1043402

keith:
ok I pushed a new version here where I manually pulled changes from the upstream commit, but also left the VFS feature around for existing users. we can probably remove it as upstream did once we're confident this works without downsides and we don't care about older swift versions

PR #768, opened 2022-02-24
Use the target triple from the C++ toolchain instead of deriving it from values in the Apple configuration fragment.

thii:
When using `--host_cpu=darwin_x86_64` on an ARM (Apple Silicon) host, the value returned by the Apple configuration fragment is wrong, and we end up compiling Swift for `arm64`. Since the C++ toolchain already has the correct triple verbatim as we want it in the `target_gnu_system_name` field, just use that, and add some generally useful utilities for manipulating the triples and their components.

Note to open-source rules maintainers: This change requires/assumes that your C++ toolchain configuration returns a complete target triple that includes minimum OS version and target environment; for example, `x86_64-apple-ios13.0-simulator`.

PiperOrigin-RevId: 429897884


thii:
Requires https://github.com/bazelbuild/bazel/pull/13466

keith:
ah i missed this and took a stab too, https://github.com/bazelbuild/rules_swift/pull/858 we can take yours though, looks like all the linked stuff is at least in 5.2?

keith:
ok merged that one

PR #772, opened 2022-03-03
Add support for symbol graph extraction

keith:
- d41cdb923c8dff27570709add2838c3e4c8e40ff
- 47a02f143a93f3f2ef8e61c69a634ad042c652bf
- 39dfb233289c78a73c37d27b7275cb17a06ecebc
- 3e90cda20649d1f729524ffecf787d63c5a65acf

### DIFF
--- a/swift/repositories.bzl
+++ b/swift/repositories.bzl
@@ -89,6 +89,15 @@ def swift_rules_dependencies(include_bzlmod_ready_dependencies = True):
             url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.3.0/bazel_features-v1.3.0.tar.gz",
         )
 
+        _maybe(
+            http_archive,
+            name = "com_github_apple_swift_argument_parser",
+            urls = ["https://github.com/apple/swift-argument-parser/archive/refs/tags/1.3.0.tar.gz"],
+            sha256 = "e5010ff37b542807346927ba68b7f06365a53cf49d36a6df13cef50d86018204",
+            strip_prefix = "swift-argument-parser-1.3.0",
+            build_file = "@build_bazel_rules_swift//third_party:com_github_apple_swift_argument_parser/BUILD.overlay",
+        )
+
     _maybe(
         http_archive,
         name = "com_github_apple_swift_protobuf",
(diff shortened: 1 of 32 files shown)


keith:
blocked on infra changes that are blocked on aspect_hints

luispadron:
@keith is this still blocked? Im trying to cherry-pick 6cae838e1fcbcc344fa2a29dc272261e719cba6e which has some dependencies on this for `swift_test`

keith:
yea unclear. it's probably quicker if you're worried about it to handle the merge conflicts with cherry picking that commit as I did for the other ones related to compiler plugins

brentleyjones:
I'm updating this PR

brentleyjones:
@jpsim This adds a different way of extracting symbol graphs. It also changes the name of the directory for `derived_files.symbol_graph_directory` (to match what was used upstream), and I'm not sure if that could cause a clash if your feature is also enabled while using the new rule/aspect. So, two questions:

1. Does this new stuff replace the need for your existing stuff?
2. If not, is there anything that should be changed here, to ensure that they work well together (including potential doc changes)?

swift/repositories.bzl
```diff
@@ -107,6 +116,15 @@ def swift_rules_dependencies(include_bzlmod_ready_dependencies = True):
...
 
+    _maybe(
+        http_archive,
+        name = "com_github_apple_swift_docc_symbolkit",
+        urls = ["https://github.com/apple/swift-docc-symbolkit/archive/refs/tags/swift-5.10-RELEASE.tar.gz"],
+        sha256 = "de1d4b6940468ddb53b89df7aa1a81323b9712775b0e33e8254fa0f6f7469a97",
+        strip_prefix = "swift-docc-symbolkit-swift-5.10-RELEASE",
+        build_file = "@build_bazel_rules_swift//third_party:com_github_apple_swift_docc_symbolkit/BUILD.overlay",
+    )
+
```
brentleyjones, line 127: I don't like adding another non-bzlmod dependency. I assume we should add this to BCR like `com_github_apple_swift_argument_parser`?

brentleyjones:
Should non-Xcode toolchains support `swift_action_names.SYMBOL_GRAPH_EXTRACT`?

brentleyjones:
I got confirmation that the Linux toolchain isn't being upstreamed anymore. So there are parts we need to do here for that.

luispadron:
This symbol graph rule also seems to better support extending the symbol graph (e.g. with minimum access level attr). Theres more flags related to symbol graph generation we could add too, such as the one required to support: https://github.com/bazelbuild/rules_apple/issues/2445. 

I have a work-in-progress PR that adds `-emit-extension-block-symbols` via a `swift.emit_symbol_graph_extension_blocks` feature but using an attr in this new rule makes more sense. Do we think the change to the existing symbol graph code is worth making or should we wait for this new rule? Do we think this will land soon

brentleyjones:
I would like to get @jpsim's thoughts on this.

jpsim:
I'm no longer involved with any projects that would use this or the previous symbol extraction for DocC. If this is what's upstreamed and sounds like more useful than we have today, then I'm all for cherry-picking it in.

luispadron (approved):
CI should pass on a rebase now

swift/repositories.bzl
brentleyjones: Not doing it for now.

PR #778, opened 2022-03-17
Fix apple_static_framework_import duplication

keith:
Previously we passed every dependent framework binary on the command
line to the link invocation. This resulted in duplicate symbols in the
case that multiple dependencies depended on the same
apple_static_framework_import. This now treats static frameworks just
like dynamic frameworks and uses `-F` + `-framework` to discover these.
Allowing the linker to dedup them.


keith:
Theoretically we could de-dup these ourselves by building up a set of them in this loop, and only adding them on the command line a single time, that might be preferred in the case you had many `-F` flags, but might not be worth it for this type of use case.

Issue #775, opened 2022-03-08
Xcode 13.3 Beta 3 swift_binary not building


regularberry:
I've come across an issue with `swift_binary` and `macos_command_line_application` not building with Xcode 13.3 Beta 3 command line tools (specifically the Mac OS 12.3 SDK that ships with it).

At first I thought it was an issue with creating a custom rule and passing in a command line tool as the executable, but from creating a simple test project I've come across something else, which may or may not be related.

I've attached a very simple example project which replicates the issue.

You can build a `swift_binary` using Xcode 13.2.1 with no additional flags, but to build it in Xcode 13.3.0 you'll need `--swiftcopt=-whole-module-optimization` turned on.

The test project provides 3 configs:
* `bazelisk build SimpleCommandLine --config=success` points at Xcode 13.2.1 and succeeds.
* After a `bazel clean --expunge`, `bazelisk build SimpleCommandLine --config=fail` which points at Xcode 13.3.0 will fail
* Finally, `bazelisk build SimpleCommandLine --config=success-with-13.3` will succeed because it turns whole module optimization on.

Please correct me if I'm just not understanding something, I'm still a Bazel novice. 

Thank you for your time!

[SimpleCommandLine.zip](https://github.com/bazelbuild/rules_swift/files/8207686/SimpleCommandLine.zip)

This is using Bazel 5.0.0 and rules_apple 0.33.0.

The error given is:
```
Compiling Swift module //SimpleCommandLine:SimpleCommandLine failed: (Exit 1): worker failed: error executing command 
  (cd /private/var/tmp/_bazel_Z006LFH/efd460bccebb4d2e1f7d3ce9a81ace5b/execroot/__main__ && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=12.3 \
    XCODE_VERSION_OVERRIDE=13.3.0.13E5104i \
  bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/build_bazel_rules_swift/tools/worker/worker swiftc @bazel-out/darwin-dbg/bin/SimpleCommandLine/SimpleCommandLine_SimpleCommandLine.swiftmodule-0.params)
# Configuration: d03c48ef8c2a47edb079909273e751b2045c82b2c50d44193c24ab65b3afcb53
# Execution platform: @local_config_platform//:host
[...]

PR #782, opened 2022-03-18
Add private feature noting emit-module-separately behavior

keith:
With Swift 5.6 the swift driver now emits modules in a separate step,
which skips the partial module merging step that happened before. This
means partial swiftmodule files are not emitted. This reflects that in
our logic for determining this. It's also enabled by default for WMO
builds, but that shouldn't matter for now either since we already
assumed that.

Technically this isn't a perfect solution, this ignores the fact that
users could pass -no-emit-modules-separately, or the WMO version of
these flags but this doesn't seem to be an issue.

https://github.com/apple/swift-driver/commit/e0c3e5c740e68202e0a7777d3f6af795d9150296

Fixes https://github.com/bazelbuild/rules_swift/issues/775

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -64,6 +64,7 @@ load(
     "SWIFT_FEATURE_USE_OLD_DRIVER",
     "SWIFT_FEATURE_USE_PCH_OUTPUT_DIR",
     "SWIFT_FEATURE_VFSOVERLAY",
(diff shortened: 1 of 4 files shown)


keith:
This was another potential fix for https://github.com/bazelbuild/rules_swift/pull/781/ I merged that one instead so I'm just submitting this in case something comes up with that one where we need an alternative.

PR #789, opened 2022-03-29
Bump Linux and Swift version

keith:
I had to disable grpc tests on linux here because the version of grpc-swift we're on is very old and no longer compiles on Linux. Updating it is quite difficult since it has changed to a large dependency tree on nio

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -24,12 +24,14 @@ x_defaults:
     build_targets:
       - "//examples/..."
       - "-//examples/apple/..."
+      - "-//examples/xplatform/grpc/..." # TODO: Fix grpc on Linux
     test_flags: *linux_flags
     test_targets:
       - "//doc/..."
       - "//examples/..."
       - "//test/..."
       - "-//examples/apple/..."
+      - "-//examples/xplatform/grpc/..." # TODO: Fix grpc on Linux
 
 tasks:
   macos_latest:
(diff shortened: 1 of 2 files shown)


.bazelci/presubmit.yml
```diff
@@ -50,31 +51,31 @@ tasks:
...
-  ubuntu1804_latest:
+  ubuntu2004_latest:
     name: "Latest Bazel"
     bazel: latest
     shell_commands:
-      - "echo --- Downloading and extracting Swift 5.4.2 to $SWIFT_HOME"
+      - "echo --- Downloading and extracting Swift $SWIFT_VERSION to $SWIFT_HOME"
       - "mkdir $SWIFT_HOME"
-      - "curl https://mirror.bazel.build/swift.org/builds/swift-5.4.2-release/ubuntu1804/swift-5.4.2-RELEASE/swift-5.4.2-RELEASE-ubuntu18.04.tar.gz | tar xvz --strip-components=1 -C $SWIFT_HOME"
+      - "curl https://download.swift.org/swift-${SWIFT_VERSION}-release/ubuntu2004/swift-${SWIFT_VERSION}-RELEASE/swift-${SWIFT_VERSION}-RELEASE-ubuntu20.04.tar.gz | tar xvz --strip-components=1 -C $SWIFT_HOME"
```
brentleyjones, line 60: Do we want to make `ubuntu2004` a variable as well?

keith: eh slightly more annoying because of having to have 2004 and 20.04, probably fine to just find + replace for now

brentleyjones (approved):

PR #793, opened 2022-04-06
Add missing `bzl_library` declarations.

cgrindel:
Generating stardoc that depends upon `rules_swift` fails without these declarations.


keith:
We generate some docs on CI, I guess we just don't hit these in that codepath?

keith:
Can we add whatever you're doing that hit this issue to CI here as well? we can merge this to unblock tho

keith (approved):

cgrindel:
Will do. I will try to put something together this weekend.

PR #795, opened 2022-04-09
Add test for missing `bzl_library` declarations

cgrindel:
Related to #793.

- Added `bzl_test` macro to define a build target that ensures that a `bzl_library` target is valid (e.g. loaded dependencies are specified in `deps`).


keith (approved):
thanks!

PR #797, opened 2022-04-12
common: remove `file_system` library

compnerd:
Remove the custom `file_system` library in favour of `std::filesystem`.
This allows us to avoid having to re-implement the functionality on all
the platforms.  This will also enable us to support Windows eventually
without having to do all the file system handling (which requires a
significant amount of UTF-8 to UCS-2 conversions) and trying to emulate
the semantics of the functions.


tools/worker/BUILD
```diff
@@ -35,10 +36,17 @@ cc_library(
         "work_processor.h",
     ],
     hdrs = ["compile_with_worker.h"],
+    copts = selects.with_or({
+        ("@bazel_tools//tools/cpp:clang-cl", "@bazel_tools//tools/cpp:msvc"): [
```
keith, line 40: should these be conditionals on windows instead?

tools/common/file_system.cc
```diff
@@ -1,141 +0,0 @@
...
-#error Only macOS and Unix are supported at this time.
-#endif
-}
-
-bool CopyFile(const std::string &src, const std::string &dest) {
-#ifdef __APPLE__
-  // The `copyfile` function with `COPYFILE_ALL` mode preserves permissions and
-  // modification time.
-  return copyfile(src.c_str(), dest.c_str(), nullptr,
-                  COPYFILE_ALL | COPYFILE_CLONE) == 0;
```
keith, line 63: So with `std::filesystem::copyfile` are we losing this optimization? It doesn't look like there was a performance citation in the original change https://github.com/bazelbuild/rules_swift/pull/232 but theoretically it was saving time + disk space

tools/worker/BUILD
compnerd: So, one _can_ cross-compile to Windows without `clang-cl` and `msvc`.  However, it does seem that the constraint isnt being matched :-(.  Ultimately, it is the compiler you are using rather than targeting Windows.

tools/common/file_system.cc
compnerd: No, we do not: https://github.com/llvm/llvm-project/blob/main/libcxx/src/filesystem/operations.cpp#L836-L855

keith: Hrm I don't think that's the same without the `COPYFILE_CLONE` flag?

compnerd: Ah, managed to find a man page on it, yes, it is definitely different.  I suppose I can create a wrapper function that uses `copyfile` manually on macOS :/

keith (approved):

PR #798, opened 2022-04-14
common: remove `path_utils` library

compnerd:
Remove the custom `path_utils` library in favour of `std::filesystem`.
This allows us to avoid having to re-implement the functionality on all
the platforms. This will also enable us to support Windows eventually
without having to do all the file system handling (which requires a
significant amount of UTF-8 to UCS-2 conversions) and trying to emulate
the semantics of the functions.


keith (approved):

keith:
small buildifier issue

PR #802, opened 2022-04-15
common: port `temp_file` to C++17 and Windows

compnerd:
Use C++17 more heavily here to avoid the use of `fts.h` making the code
easier to port.  Reduce the string manipulation by using
`std::filesystem::path` and use that to additionally locate the
temporary file location.  Emulate the POSIX semantics for `mkdtemp` and
`mkstemp` as much as possible to provide equivalent behaviour on
Windows.

### DIFF
--- a/tools/common/temp_file.h
+++ b/tools/common/temp_file.h
@@ -15,46 +15,65 @@
 #ifndef BUILD_BAZEL_RULES_SWIFT_TOOLS_COMMON_TEMP_FILE_H
 #define BUILD_BAZEL_RULES_SWIFT_TOOLS_COMMON_TEMP_FILE_H
 
-#include <fts.h>
-#include <string.h>
-#include <unistd.h>
-
+#include <cassert>
 #include <cerrno>
+#include <cstdio>
 #include <cstdlib>
+#include <cstring>
+#include <filesystem>
 #include <iostream>
 #include <memory>
 #include <string>
 
+#if defined(_WIN32)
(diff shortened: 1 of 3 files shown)


tools/common/temp_file.h
```diff
@@ -15,46 +15,65 @@
...
+  TempFile(const TempFile &) = delete;
+  TempFile &operator=(const TempFile &) = delete;
+  TempFile(TempFile &&) = default;
+  TempFile &operator=(TempFile &&) = default;
+
   // Create a new temporary file using the given path template string (the same
   // form used by `mkstemp`). The file will automatically be deleted when the
   // object goes out of scope.
   static std::unique_ptr<TempFile> Create(const std::string &path_template) {
-    const char *tmpDir = getenv("TMPDIR");
```
keith, line 35: I _think_ dropping this is ok, but we might end up needing to restore it later.

keith (approved):

tools/common/temp_file.h
compnerd: I _think_ that the C++ runtime should evaluate that on Unix platforms and thus should be covered already.

keith: Looks like it! https://github.com/llvm/llvm-project/blob/94d96c2a6201e0927e835b5c0e5b4cb8f95eeadf/libcxx/src/filesystem/operations.cpp#L1546-L1553

PR #803, opened 2022-04-15
common: add subprocess support for Windows

compnerd:
Add a Windows port for the subprocess handling.  This is fairly brittle,
does not perform job control (which means runaway processes potentially
on interrupt), and does not perform the proper handling for Unicode.
However, the logic is complex enough that adding in an additional
encoding conversion for the initial version may not be the wisest idea.
Subsequent passes could clean up the implementation.

### DIFF
--- a/tools/common/process.cc
+++ b/tools/common/process.cc
@@ -12,6 +12,213 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "process.h"
+
+#include <algorithm>
+#include <cassert>
+#include <iostream>
+#include <iterator>
+#include <sstream>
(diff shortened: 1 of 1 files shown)


tools/common/process.cc
```diff
@@ -12,6 +12,213 @@
...
+#include <cassert>
+#include <iostream>
+#include <iterator>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#if defined(_WIN32)
+
+#define WIN32_LEAN_AND_MEAN
```
keith, line 27: amazing

keith (approved):

PR #805, opened 2022-04-20
build: make the build constraints for MSVC work

compnerd:
This adds a working form of the constraints to get the correct flags by
default when building with cl (MSVC or clang) for Windows.

### DIFF
--- a/tools/worker/BUILD
+++ b/tools/worker/BUILD
@@ -37,7 +37,7 @@ cc_library(
     ],
     hdrs = ["compile_with_worker.h"],
     copts = selects.with_or({
-        ("@bazel_tools//tools/cpp:clang-cl", "@bazel_tools//tools/cpp:msvc"): [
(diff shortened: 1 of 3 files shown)


tools/worker/BUILD
```diff
@@ -20,6 +20,20 @@ config_setting(
...
+        "@bazel_tools//tools/cpp:compiler": "msvc-cl",
+    },
+)
+
+config_setting(
+    name = "clang-cl",
+    flag_values = {
+        "@bazel_tools//tools/cpp:compiler": "clang-cl",
+    },
+)
```
keith, line 35: instead of duplicating these what do you think about including them in tools/BUILD and making their visibility limited to subpackages of tools if needed?

compnerd: That sounds good to me, I wasn't particularly a fan of having them be duplicated.

keith (approved):

PR #809, opened 2022-04-21
swift: use executable suffixes for tool invocation

compnerd:
When building on Windows, we should suffix the executable tool with the
executable suffix.  This ensures that the worker can actually launch the
process.  If the user specified the swift_executable, ensure that we do
not double suffix the tool.


keith (approved):

keith:
minor buildifier

PR #814, opened 2022-04-22
swift_toolchain: update the toolchain rules for Windows

compnerd:
Introduce a `SDKROOT` attr which is configured initially via the
environment.  This can then be passed along to the compiler
appropriately (`-sdk SDKROOT`).  This attribute also provides a
number of subsequent other values:
- `../../../Info.plist` contains version information for XCTest
- `../../Library/XCTest-$(VERSION)/usr/lib/swift/{os}` XCTest modules
- `../../Library/XCTest-$(VERSION)/usr/lib/swift/{os}/{arch}` XCTest libraries
- `../../Library/XCTest-$(VERSION)/usr/bin{32,64,32a,64a}` XCTest runtime

It can also be used for additional information:
- `SDKSettings.plist` contains additional flags that we should pass to
  the compiler

Add the new `xctest_version` attribute to allow the toolchain to specify
the XCTest version information which is required when building with the
Windows SDK.

Introduce features for `swift-autolink-extract` as these are platform
level configuration rather than build host and are not needed when
targeting PE/COFF or MachO.


swift/internal/swift_autoconfiguration.bzl
```diff
@@ -289,44 +295,83 @@ xcode_swift_toolchain(
...
   "swift_toolchain",
 )
+load(
+  "@build_bazel_rules_swift//swift:swift.bzl",
+  "swift_package_configuration",
+)
 
 package(default_visibility = ["//visibility:public"])
 
+swift_package_configuration(
```
keith, line 350: Hrm I think instead of needing to use this rule, you should be able to set `disabled_features` on the `swift_toolchain` below which then gets used here:

https://github.com/bazelbuild/rules_swift/blob/8d4b096b90e47095755e47c27e749ae9b9f83e81/swift/internal/swift_toolchain.bzl#L254-L256

This rule hasn't been used much externally so I'd be a bit worried about how it composes to set something global like this.

swift/internal/compiling.bzl
```diff
@@ -153,6 +161,61 @@ def compile_action_configs(
         ),
     ]
 
+    if sdkroot:
```
keith, line 164: I think the way these settings all compose is a bit surprising. maybe at least the conditionals below should be de-dented?

swift/internal/compiling.bzl
```diff
@@ -153,6 +161,61 @@ def compile_action_configs(
...
+            swift_toolchain_config.action_config(
+                actions = [
+                    swift_action_names.COMPILE,
+                    swift_action_names.DERIVE_FILES,
+                    swift_action_names.PRECOMPILE_C_MODULE,
+                    swift_action_names.DUMP_AST,
+                ],
+                configurators = [
+                    swift_toolchain_config.add_arg(
+                        "-sdk",
```
keith, line 175: The difference between `-sdk` being passed here vs `xcode_swift_toolchain` is a bit strange. but maybe that seems like it has to work that way because of bazel's apple implementation details?

swift/internal/compiling.bzl
```diff
@@ -153,6 +161,61 @@ def compile_action_configs(
...
+                        swift_toolchain_config.add_arg(
+                            paths.join(sdkroot, "..", "..", "Library", "XCTest-{}".format(xctest_version), "usr", "lib", "swift", os),
+                            format = "-I%s",
+                        ),
+                    ],
+                ),
+            )
+
+            # Compatibility with older builds of the Swift SDKs
+            if arch:
```
keith, line 201: if this is just for older versions can we choose not to support it? how old are we talking? even with apple + linux today i don't think we realistically aim to support old verisons just because of the low use + maintenance burden

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -289,44 +295,83 @@ xcode_swift_toolchain(
...
+def _get_python_bin(repository_ctx):
+    if "PYTHON_BIN_PATH" in repository_ctx.os.environ:
+        return repository_ctx.os.environ.get("PYTHON_BIN_PATH").strip()
+    out = repository_ctx.which("python3.exe")
+    if out:
+        return out
+    out = repository_ctx.which("python.exe")
+    if out:
+        return out
+    return None
```
keith, line 307: should this case `fail()` instead?

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -289,44 +295,83 @@ xcode_swift_toolchain(
         ),
     )
 
+def _get_python_bin(repository_ctx):
+    if "PYTHON_BIN_PATH" in repository_ctx.os.environ:
+        return repository_ctx.os.environ.get("PYTHON_BIN_PATH").strip()
+    out = repository_ctx.which("python3.exe")
+    if out:
+        return out
+    out = repository_ctx.which("python.exe")
```
keith, line 304: it looks fine for this case but realistically we don't support py2 here anymore

swift/internal/swift_toolchain.bzl
```diff
@@ -126,6 +149,8 @@ def _all_action_configs(additional_swiftc_copts):
 def _swift_linkopts_cc_info(
```
keith, line 149: considering there's almost no overlap in these 2 functions, what do you think about hoisting out the windows support and doing a separate `_swift_windows_linkopts_cc_info` instead? that would make the logic a bit nicer, and reduce the risk of conflicts from upstream

swift/internal/feature_names.bzl
```diff
@@ -195,6 +195,14 @@ SWIFT_FEATURE_USE_OLD_DRIVER = "swift.use_old_driver"
 # using a sufficiently recent version of Swift (4.2 or higher).
 SWIFT_FEATURE_USE_RESPONSE_FILES = "swift.use_response_files"
 
+# If enabled, Swift linking actions will use `swift-autolink-extract` to extract
+# the linker arguments.  This is required for ELF targets.
```
keith, line 199: Can you add a bit to these 2 comments about how they shouldn't be used by consumers?

swift/internal/compiling.bzl
compnerd: By old, we are talking about current stable release :)

swift/internal/swift_autoconfiguration.bzl
compnerd: I don't have a strong opinion on this - as a best effort, seems nice to just use it if available.  An other option would be to grow a dependency on a library (e.g. https://github.com/animetrics/PlistCpp).

swift/internal/swift_autoconfiguration.bzl
compnerd: I suppose so.  Makes me wonder if we should just add a C++ tool to deal with this ...

swift/internal/swift_toolchain.bzl
compnerd: Ah good point, they are completely divergent now.  This wasn't the case when I started.

swift/internal/compiling.bzl
compnerd: Well, they rely on sdkroot still.  So de-denting them and repeating the check is no better, and reads worse IMO.

swift/internal/compiling.bzl
compnerd: Hmm, which difference is that?  And, yeah, I couldn't figure out how to make it identical.

swift/internal/feature_names.bzl
compnerd: Ah, I see what you mean.  Sure.

swift/internal/swift_autoconfiguration.bzl
compnerd: I tried to do that but failed initially.  I can give it another shot.

swift/internal/swift_autoconfiguration.bzl
keith: I think this option is fine for now, and leaving py2 is fine too assuming it works in both anyways, but if things get more complex we might want to drop py2

swift/internal/swift_autoconfiguration.bzl
keith: IIRC swift requires py so I assume it's safe in general? I don't love the idea of another dep for this. Ideally that file would be json I guess at least

swift/internal/compiling.bzl
keith: Ha, ok

swift/internal/compiling.bzl
keith: I was thinking about how for macOS the `-sdk` flag is configured there, not in this more standard place

swift/internal/swift_autoconfiguration.bzl
compnerd: Yes and no.  Yes, you need python for lldb.  No, that does not guarantee python.  In fact, the Docker containers for Windows with Swift show you this - they only have enough python to run lldb, not enough python to run this command :).  There is no dependency on the python executable for swift, only the library.

swift/internal/swift_autoconfiguration.bzl
compnerd:
```
ERROR: C:/sourcecache/_bazel_user/lqtpnjni/external/build_bazel_rules_swift_local_config/BUILD:9:16: @build_bazel_rules_swift_local_config//:toolchain: no such attribute 'disabled_features' in 'swift_toolchain' rule
ERROR: C:/sourcecache/sourcecache/rules_swift/examples/xplatform/proto/BUILD:23:13: every rule of type swift_binary implicitly depends upon the target '@build_bazel_rules_swift_local_config//:toolchain', but this target could not be found because of: Target '@build_bazel_rules_swift_local_config//:toolchain' contains an error and its package is in error
ERROR: Analysis of target '//examples/xplatform/proto:proto' failed; build aborted: Analysis failed
```

compnerd: This is potentially incorrectly done, but should move the disabling to an allow list applied to `"//..."` as it seems that the packages need to be specified?  Or would the empty packages set get applied to all packages?  I was unable to determine that.

keith: I think this is just kinda the wrong tool for the job. I realized one actual problem with this which is if you have external dependencies I don't think this would apply to them 🤔 since those aren't part of `//...` but in some other namespace like `@foo//...`

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -289,44 +295,90 @@ xcode_swift_toolchain(
...
+        _get_python_bin(repository_ctx),
+        "-c",
+        "import os, plistlib; " +
+        "print(plistlib.loads(open(os.path.join(r'{}', '..', '..', '..', 'Info.plist'), 'rb').read(), fmt=plistlib.FMT_XML)['DefaultProperties']['XCTEST_VERSION'])".format(repository_ctx.os.environ["SDKROOT"]),
+    ])
+
+    env = {
+        "Path": repository_ctx.os.environ["Path"] if "Path" in repository_ctx.os.environ else repository_ctx.os.environ["PATH"],
+        "ProgramData": repository_ctx.os.environ["ProgramData"],
+    }
```
keith, line 339: I think you'll also want to add these here: https://github.com/bazelbuild/rules_swift/blob/ada7d31ba7b461d916ee923ba4ebe3e043f65079/swift/internal/swift_autoconfiguration.bzl#L347 to make sure that changes to them cause the repository_rule to be re-run. I'm not sure about the case insensitivity piece of `Path` vs `PATH`, but it probably doesn't hurt to have both. I'm surprised to not see these in bazel itself https://github.com/bazelbuild/bazel/blob/73f1ecbc1cb00e16ceda4b582f4d57268f8701cd/tools/cpp/cc_configure.bzl#L127-L174

swift/internal/swift_autoconfiguration.bzl
compnerd: I don't mind changing this, I would prefer that we don't do this - but I don't see how to make this work otherwise.

swift/internal/swift_toolchain.bzl
```diff
@@ -247,7 +355,7 @@ def _swift_toolchain_impl(ctx):
             root_dir = toolchain_root,
             swift_worker = ctx.executable._worker,
             test_configuration = struct(
-                env = {},
+                env = env,
```
keith, line 358: Since we weren't passing anything here before, was literally nothing being propagated to the tests? Or is there some default case around `PATH` or something?

swift/internal/swift_toolchain.bzl
```diff
@@ -207,16 +290,41 @@ def _swift_toolchain_impl(ctx):
...
+            bindir = "bin32"
+        elif ctx.attr.arch == "arm64":
+            bindir = "bin64a"
+        else:
+            fail("unsupported arch `{}`".format(ctx.attr.arch))
+
+        xctest = paths.normalize(paths.join(ctx.attr.sdkroot, "..", "..", "Library", "XCTest-{}".format(ctx.attr.xctest_version), "usr", bindir))
+        env = dicts.add(
+            ctx.attr.env,
+            {"Path": xctest + ";" + ctx.attr.env["Path"]},
```
keith, line 323: For reference there is this https://docs.bazel.build/versions/5.1.0/skylark/lib/configuration.html#host_path_separator but in this case since you already have the windows `if` it's not a huge deal either way

swift/internal/swift_autoconfiguration.bzl
compnerd: I believe that this should be correct now.

swift/internal/swift_toolchain.bzl
compnerd: IIRC, the environment was entirely empty - including some of the testing environment.  Whoops!

swift/internal/swift_autoconfiguration.bzl
compnerd: The lookup needs to be case-sensitive for starlark, `GetEnvironmentVariable` is case insensitive.  This was annoying to figure out - if there is a way to do this nicely, please let me know.

swift/internal/swift_toolchain.bzl
keith: Ah but I guess this is a no-op on linux still since you'd have to actually set this new var to get anything?

compnerd: Right - the default value of `env` is `{}` so this spells it differently now to support Windows without changing the non-Windows path.

swift/internal/swift_autoconfiguration.bzl
compnerd: Added the variables to `environ` to track.

keith (approved):
awesome! thanks for the hard work!

PR #815, opened 2022-04-25
Add `compilation_context` to `SwiftInfo` modules

brentleyjones:
This adds a new struct to `SwiftInfo` modules that exposes the information used during compilation of a Swift target. The main use case is aspects, in particular ones for IDE integration. The attributes exposed mirror some of those found in `CcInfo.compliation_context` and `ObjcProvider`.

The initial attributes are ones that I personally need right now, but we can add more later as they are identified.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -208,6 +208,39 @@ A `struct` containing the `compilation_context`, `module_map`, and
   `precompiled_module` fields provided as arguments.
 
 
+<a id="#swift_common.create_compilation_context"></a>
+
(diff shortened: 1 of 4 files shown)


brentleyjones:
This should be good to go now.

keith (approved):
seems fine. why don't things need this previously?

brentleyjones:
Just depends on how you want to do your IDE integration. The `direct_sources` exist on the other providers for IDE integration and make it so you don't have to special case for rules, which is nice.

PR #822, opened 2022-05-03
swift: correct path separator

compnerd:
When using the rules on Windows, we should use `;` rather than `:` for
the path separator.  This uses the bazel configuration to drive the
value making the selection agnostic to the platform.

### DIFF
--- a/swift/internal/swift_grpc_library.bzl
+++ b/swift/internal/swift_grpc_library.bzl
@@ -49,7 +49,8 @@ def _register_grpcswift_generate_action(
         protoc_executable,
         protoc_plugin_executable,
         flavor,
-        extra_module_imports):
+        extra_module_imports,
+        host_path_separator):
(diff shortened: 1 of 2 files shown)


swift/internal/swift_grpc_library.bzl
```diff
@@ -49,7 +49,8 @@ def _register_grpcswift_generate_action(
         protoc_executable,
         protoc_plugin_executable,
         flavor,
-        extra_module_imports):
+        extra_module_imports,
+        host_path_separator):
```
segiddins, line 53: buildifier wants docs for the `host_path_separator` param

segiddins (approved):

swift/internal/swift_grpc_library.bzl
compnerd: Ah, right; thanks!  Fixed.

keith (approved):

PR #824, opened 2022-05-03
swift: introduce a feature to control debug info on Windows

compnerd:
Windows poorly supports DWARF (or rather DWARF poorly supports Windows)
and the swift driver poorly supports CodeView.  However, both are useful
in different scenarios.  Add a feature to use codeview, and enable it by
default.  The default is important as DWARF cannot supported by link
which is picked up as the default linker.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -30,6 +30,7 @@ load(
     ":feature_names.bzl",
     "SWIFT_FEATURE_BITCODE_EMBEDDED",
     "SWIFT_FEATURE_CACHEABLE_SWIFTMODULES",
+    "SWIFT_FEATURE_CODEVIEW_DEBUG_INFO",
     "SWIFT_FEATURE_COVERAGE",
     "SWIFT_FEATURE_COVERAGE_PREFIX_MAP",
     "SWIFT_FEATURE_DBG",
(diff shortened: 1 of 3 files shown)


swift/internal/compiling.bzl
```diff
@@ -498,7 +499,24 @@ def compile_action_configs(
...
+                [SWIFT_FEATURE_FULL_DEBUG_INFO],
+            ],
+        ),
+        swift_toolchain_config.action_config(
+            actions = [
+                swift_action_names.COMPILE,
+                swift_action_names.DERIVE_FILES,
+            ],
+            configurators = [
+                swift_toolchain_config.add_arg("-g"),
```
segiddins, line 513: will this add a duplicate `-g` flag? is that an issue?

compnerd: No, it should not add a second `-g` as the default debug opt is made incompatible with this and so will not be applied.

swift/internal/compiling.bzl
```diff
@@ -509,7 +527,10 @@ def compile_action_configs(
                 swift_toolchain_config.add_arg("-gline-tables-only"),
             ],
             features = [SWIFT_FEATURE_FASTBUILD],
-            not_features = [SWIFT_FEATURE_FULL_DEBUG_INFO],
+            not_features = [
+                [SWIFT_FEATURE_FULL_DEBUG_INFO],
+                [SWIFT_FEATURE_CODEVIEW_DEBUG_INFO],
+            ],
```
keith, line 533: Wouldn't this just disable this feature for fastbuild, but then also not apply the other one since it's limited to dbg?

swift/internal/compiling.bzl
```diff
@@ -498,7 +499,24 @@ def compile_action_configs(
                 swift_action_names.DERIVE_FILES,
             ],
             configurators = [swift_toolchain_config.add_arg("-g")],
-            features = [[SWIFT_FEATURE_DBG], [SWIFT_FEATURE_FULL_DEBUG_INFO]],
+            features = [
+                [SWIFT_FEATURE_DBG],
+                [SWIFT_FEATURE_FULL_DEBUG_INFO],
+            ],
```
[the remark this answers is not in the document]
compnerd: Ah right, I had originally added in `SWIFT_FEATURE_CODEVIEW_DEBUG_INFO`, which didn't work out, and removal didn't undo the formatting change.

swift/internal/compiling.bzl
compnerd: This should enable it for fastbuild without full debug info.  However, if you use codeview, it doesn't matter what you do, `-glinetables` should not be enabled with that.

keith: here's my understanding of what flags should apply, can you fill in the question marks and correct any mistakes

|targeting windows|cfg|codeview enabled?|desired flags|
|-|-|-|-|
|yes|dbg|yes|`-g -debug-info-format=codeview`|
|yes|dbg|no|`-g`|
|yes|fastbuild|yes|??|
|yes|fastbuild|no|??|
|no|dbg|yes (ignored)|`-g`|
|no|dbg|no|`-g`|
|no|fastbuild|yes (ignored)|`-gline-tables-only`|
|no|fastbuild|no|`-gline-tables-only`|

keith: note the 2 "ignored" lines don't actually work with this change, and might be something we should fix at some point with a "supports codeview" feature that we only add for windows 🤔

compnerd:
Made fastbuild and dbg the same as the C++ rules.

keith (approved):

PR #825, opened 2022-05-04
tools: port mkdir_and_run to work on Windows

compnerd:
Prefer a cmd script over bash on Windows.  This allows this command to
be used across platforms.


keith (approved):
seems fine, buildifier issue

PR #831, opened 2022-05-24
build: workaround LLVM bug on Windows

compnerd:
The outliner does not correctly handle landingpads and will attempt to
outline the EH funclet which generates incorrect IR (and is also
pointless).  Workaround this for the time being.

### DIFF
--- a/tools/common/BUILD
+++ b/tools/common/BUILD
@@ -8,6 +8,13 @@ cc_library(
     name = "bazel_substitutions",
     srcs = ["bazel_substitutions.cc"],
     hdrs = ["bazel_substitutions.h"],
+    copts = select({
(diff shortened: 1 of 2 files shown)


keith:
Is there any issue link or test we can add so we know when it's fixed?

compnerd:
I _think_ that I had opened an issue for this?  Let me check.

keith (approved):

hiraditya:
The repro in https://github.com/llvm/llvm-project/issues/40056 works fine now. Do we have another testcase to work with?

compnerd:
I don't have a simple reproducer, but building apple/SourceKit-LSP in release mode should trigger it (I think it might even show up in just apple/indexstore-db in release mode).

PR #832, opened 2022-05-24
CI: complete the Windows CI coverage

compnerd:
Extend the Windows CI to include the tests as well.  This will mostly
complete the Windows support (the remaining pieces remain in supporting
the Windows ARM64 build host).

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -40,12 +40,24 @@ x_defaults:
       - "-//test:output_file_map_default"
   windows_common: &windows_common
     platform: windows
+    environment:
+      SWIFT_VERSION: 6.1.0
+      PATH: "%LOCALAPPDATA%\\Programs\\Swift\\Toolchains\\%SWIFT_VERSION%+Asserts\\usr\\bin;%LOCALAPPDATA%\\Programs\\Swift\\Runtimes\\%SWIFT_VERSION%\\usr\\bin;%Path%"
+      SDKROOT: "%LOCALAPPDATA%\\Programs\\Swift\\Platforms\\%SWIFT_VERSION%\\Windows.platform\\Developer\\SDKs\\Windows.sdk"
     build_flags:
       # Override 'sandboxed' strategy set in .bazelrc because it's not
       # available on Windows
       - "--strategy=SwiftCompile="
+      - "--repo_env=SDKROOT=%LOCALAPPDATA%\\Programs\\Swift\\Platforms\\%SWIFT_VERSION%\\Windows.platform\\Developer\\SDKs\\Windows.sdk"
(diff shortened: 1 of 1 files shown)


keith (approved):

compnerd:
@keith okay, the install issue is resolved, but now the problem is that `Path` and `SDKROOT` are not in the environment, and I don't see how to get that added :(

keith:
are the values stable enough to pass directly? like `--repo_env=ProgramData=whatever`

compnerd:
I supposes that they should be - the installer is installing into a user specific location - `%LocalAppData%\Program\Swift\Toolchains\%SwiftVersion%+Asserts\usr\bin`, the runtime is located at `%LocalAppData%\Programs\Swift\Runtimes\%SwiftVeresion%\usr\bin` and SDKROOT can be assumed to be for Windows at `%LocalAppData%\Program\Swift\Platforms\Windows.platform\Developere\SDKs\Windows.sdk`.

compnerd:
@keith I'm not sure how to make the appropriate changes for this, so a bit of guidance would be appreciated.

keith:
I was thinking that if you set that in the `environment` key like you're doing for the others, and then passed `--repo_env=ProgramData` it would cover it?

compnerd:
I'm not too fond of the approach, but it does get further:

```
Auto-Configuration Error: Setting up VC environment variables failed, WINDOWSSDKDIR is not set by the following command:
--
  | "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\VCVARSALL.BAT" amd64  -vcvars_ver=14.37.32822
```

keith:
is that error one you're thinking the CI setup should have solved?

compnerd:
Well, that seems like something that general Windows support should have figured out. I'm not sure what is going on, but it seems that we are missing something in the image possibly? Or we are not invoking things properly.

keith:
this makes it sound like maybe the windows setup doesn't have the SDK? https://github.com/bazelbuild/bazel/issues/13261 

is that what you're thinking?

should that be solved by https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/setup-windows.ps1#L137 ?

compnerd:
Yeah, exactly, it seems like that is what is needed.

keith:
is it possible it's setup correctly but the env var isn't set? i assume that link is the current runner

brentleyjones:
@compnerd Wondering where we stand with this. We are going to be making some changes to the worker soon, and I want to make sure I don't regress Windows support.

compnerd:
@brentleyjones I think that @keith is right and that the Windows SDK is missing. Once that is installed, I think that this should just work out of the box now. The admin requirements are removed upstream and we now install to the user's directory. We no longer need to modify system files so that is also not an issue. As long as we have the `SeCreateSymbolicLinkPrivilege` privilege (or are running as admin), we should be able to support this. I'll try to see if I can get this rebased soon so we can finally get it merged.

keith:
this is now being tested @ HEAD. so at least this part isn't needed anymore

PR #837, opened 2022-05-31
Cleanup swiftdeps on corrupt incremental module

### LINKED ISSUE #819: Swift incremental can get into a weird state where it constantly fails to compile
In one of my workflows I have rapidly cancelling builds. It seems that this can get the Swift incremental storage into a weird state where subsequent builds fail to compile:

```
Failed to copy bazel-out/darwin_x86_64-dbg-ST-1c975e2849a6/bin/external/com_github_tadija_aexml/AEXML.swiftdoc for incremental builds, maybe it wasn't produced?
```

And in the incremental storage area there are no `swiftdoc`, `swiftmodule`, or `swiftsourceinfo` files.

The only way I can fix it is with `bazel clean`. Sadly with my workflow, that isn't an easy thing to do.

brentleyjones:
Fixes #819.

When a SwiftCompile action is cancelled, it can leave the incremental storage area in a corrupt state, in which all later compiles return exit code 0, but don't actually produce an output (#819).

34ce7835bda5e787b633f2ce0cc9687e90f7f952 identified a good way to tell that the storage area was in a corrupt state, but it didn't fully fix the problem. We also need to delete the existing `swiftdeps` files, which this change does.

### DIFF
--- a/tools/worker/output_file_map.cc
+++ b/tools/worker/output_file_map.cc
@@ -73,6 +73,7 @@ void OutputFileMap::UpdateForIncremental(const std::string &path,
   nlohmann::json new_output_file_map;
   std::map<std::string, std::string> incremental_outputs;
   std::map<std::string, std::string> incremental_inputs;
+  std::vector<std::string> incremental_cleanup_outputs;
 
   // The empty string key is used to represent outputs that are for the whole
   // module, rather than for a particular source file.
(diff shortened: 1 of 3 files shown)


brentleyjones:
I was able to create a reliable repro with `--jobs=1` and ctrl-c. This change always worked.

PR #838, opened 2022-06-01
Add `swift.emit_symbol_graph` feature

jpsim:
That emits the symbol graph for the Swift target to `bazel-bin`.

Symbol graphs can then be fed to tools such as [DocC](https://github.com/apple/swift-docc) or [jazzy](https://github.com/realm/jazzy) to generate documentation for a target.

For example, for [Envoy Mobile](https://github.com/envoyproxy/envoy-mobile):

```
$ ./bazelw build //library/swift:ios_lib --config=ios
$ "$(xcrun --find docc)" convert \
  --index \
  --fallback-display-name Envoy \
  --fallback-bundle-identifier io.envoyproxy.EnvoyMobile \
  --fallback-bundle-version 0.4.6 \
  --output-dir Envoy.doccarchive \
  --transform-for-static-hosting \
  --additional-symbol-graph-dir bazel-bin/library/swift/ios_lib.symbolgraph
```

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -707,8 +707,6 @@ def compile_action_configs(
         swift_toolchain_config.action_config(
             actions = [
                 swift_action_names.COMPILE,
-                swift_action_names.DERIVE_FILES,
-                swift_action_names.DUMP_AST,
             ],
             configurators = [_emit_symbol_graph_configurator],
             features = [
@@ -1625,11 +1623,6 @@ def _emit_symbol_graph_configurator(prerequisites, args):
 
       This is a directory to persist symbol graph files that can be used by
       tools such as DocC or jazzy to generate documentation.
-
-      Note: that like the global index store and module cache, we expect clang
-      to namespace these correctly per arch / os version / etc by the hash in
-      the path. However, it is also put into the bin_dir for an added layer of
-      safety.
     """
     args.add(
         "-Xfrontend",
@@ -2882,6 +2875,11 @@ def output_groups_from_other_compilation_outputs(*, other_compilation_outputs):
     if other_compilation_outputs.indexstore:
         output_groups["swift_index_store"] = depset([
             other_compilation_outputs.indexstore,
+        ])
+
+    if other_compilation_outputs.symbol_graph:
+        output_groups["swift_symbol_graph"] = depset([
+            other_compilation_outputs.symbol_graph,
         ])
 
     return output_groups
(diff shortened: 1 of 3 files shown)


Augustyniak:
unrelated observation: looking at the documentation page you generated I realized that our mocks are `public` 🤔

jpsim:
Yes the Envoy Mobile mocks are in the same framework for ease of distribution, but that’s something we should fix when we redesign the API.

swift/internal/compiling.bzl
```diff
@@ -1608,6 +1618,21 @@ def _pch_output_dir_configurator(prerequisites, args):
...
+      This is a directory to persist symbol graph files that can be used by
+      tools such as DocC or jazzy to generate documentation.
+    """
+    args.add(
+        "-Xfrontend",
+        "-emit-symbol-graph",
+    )
+    args.add(
+        "-emit-symbol-graph-dir",
+        prerequisites.symbol_graph_directory.path,
```
BalestraPatrick, line 1633: I assume the build time penalty to generate this is pretty irrelevant, right?

BalestraPatrick (approved):

swift/internal/compiling.bzl
brentleyjones: Regardless it's behind a feature flag, similar to indexing while building.

BalestraPatrick: Yeah just out of curiosity. If it has a big impact, it might be worth documenting it.

jpsim: Not sure if my methodology is correct, but I see a 2% hit to overall build times with the feature enabled:

```
$ hyperfine \
  --warmup 5 \
  --prepare "bazel clean && sleep 0.5" \
  "bazel build --features=swift.emit_symbol_graph --output_groups=+swift_symbol_graph //examples/apple/objc_interop:Printer" \
  "bazel build //examples/apple/objc_interop:Printer"
Benchmark 1: bazel build --features=swift.emit_symbol_graph --output_groups=+swift_symbol_graph //examples/apple/objc_interop:Printer
  Time (mean ± σ):     12.694 s ±  0.390 s    [User: 0.063 s, System: 0.055 s]
  Range (min … max):   12.379 s … 13.643 s    10 runs
 
Benchmark 2: bazel build //examples/apple/objc_interop:Printer
  Time (mean ± σ):     12.504 s ±  0.258 s    [User: 0.062 s, System: 0.051 s]
  Range (min … max):   12.105 s … 12.907 s    10 runs
 
Summary
  'bazel build //examples/apple/objc_interop:Printer' ran
    1.02 ± 0.04 times faster than 'bazel build --features=swift.emit_symbol_graph --output_groups=+swift_symbol_graph //examples/apple/objc_interop:Printer'
```

Since it's opt-in I don't see this needing a perf warning.

jpsim:
Thanks for the reviews, everyone! I may revisit @thii's comment (https://github.com/bazelbuild/rules_swift/pull/838#issuecomment-1144322809) after we use this some more but for now at least I can generate docs for my project.

PR #840, opened 2022-06-08
Conditionally link developer framework paths

maxwellE:
These framework paths should only be linked when the associated
target has set `testonly = True`


swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -197,6 +198,7 @@ def _swift_linkopts_providers(
         toolchain_label: The label of the Swift toolchain that will act as the
             owner of the linker input propagating the flags.
         xcode_config: The Xcode configuration.
+        is_test: The value of `testonly` for the target
```
brentleyjones, line 201: Can you move this to be above `platform` to keep it alphabetical?

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -658,9 +664,10 @@ def _xcode_swift_toolchain_impl(ctx):
     swift_linkopts_providers = _swift_linkopts_providers(
         apple_fragment,
         apple_toolchain,
-        platform,
+        ctx.attr.testonly,
         ctx.label,
         xcode_config,
+        platform,
```
brentleyjones, line 670: I think this is the wrong place now.

keith:
The CI failures expose the potential issue with this. You have to mark all libs that are used in tests as testonly, I think that is a reasonable requirement, but IIRC google had trouble with enforcing that. I wonder how many users aren't doing that today 🤔

brentleyjones:
Yeah, it's a breaking change, but I agree that it should be this way, and it makes rules_xcodeproj integration easier as well, as _it's_ making this assumption already.

keith:
yea seems fine, we can see if folks hit it

brentleyjones:
This is for this @maxwellE! Can you add test for this as well?

swift/internal/linking.bzl
```diff
@@ -22,6 +22,7 @@ load(
     "ensure_swiftmodule_is_embedded",
     "should_embed_swiftmodule_for_debugging",
 )
+load(":compiling.bzl", "developer_framework_paths", "swift_developer_lib_dir")
```
brentleyjones, line 25: These can be in their own file, to prevent `linking.bzl` from depending on `compiling.bzl`.

swift/internal/compiling.bzl
```diff
@@ -1133,6 +1147,167 @@ def _output_or_file_map(output_file_map, outputs, args):
...
+        (
+            platform_type == apple_common.platform_type.watchos and
+            not is_xcode_at_least_version(xcode_config, "12.5")
+        )
+    ):
+        return None
+
+    return paths.join(apple_toolchain.sdk_dir(), "Developer/Library/Frameworks")
+
+def is_xcode_at_least_version(xcode_config, desired_version):
```
brentleyjones, line 1233: Can we leave this where it was, but make it public there instead of moving it here? Seems better matched for the xcode_toolchain.

swift/internal/swift_binary_test.bzl
```diff
@@ -15,14 +15,20 @@
...
 load(":feature_names.bzl", "SWIFT_FEATURE_BUNDLED_XCTESTS")
 load(":linking.bzl", "register_link_binary_action")
 load(":providers.bzl", "SwiftToolchainInfo")
 load(":swift_clang_module_aspect.bzl", "swift_clang_module_aspect")
 load(":swift_common.bzl", "swift_common")
-load(":utils.bzl", "expand_locations")
+load(
+    ":utils.bzl",
+    "expand_locations",
+)
```
brentleyjones, line 31: I don't think this needs to be changed?

swift/internal/compiling.bzl
maxwellE: This introduced a cycle, ended up moving [here](https://github.com/maxwellE/rules_swift/blob/5ad13e5575f9c0a4ed6620eea8e5a4890681d9c0/swift/internal/utils.bzl#L348)

brentleyjones: Nice! Can you place it alphabetically in that file?

swift/internal/compiling.bzl
```diff
@@ -1133,6 +1155,58 @@ def _output_or_file_map(output_file_map, outputs, args):
...
+            apple_common.apple_toolchain(),
+            apple_fragment,
+            xcode_config,
+        )
+        args.add(swift_developer_lib_dir_path, format = "-I%s")
+        args.add(swift_developer_lib_dir_path, format = "-L%s")
+
+# The platform developer framework directory contains XCTest.swiftmodule
+# with Swift extensions to XCTest, so it needs to be added to the search
+# path on platforms where it exists.
```
brentleyjones, line 1175: This comment should go with with the above function, not here.

swift/internal/compiling.bzl
```diff
@@ -1133,6 +1155,58 @@ def _output_or_file_map(output_file_map, outputs, args):
...
+        xcode_config,
+    )
+    if platform_developer_framework:
+        swift_developer_lib_dir_path = swift_developer_lib_dir(
+            apple_common.apple_toolchain(),
+            apple_fragment,
+            xcode_config,
+        )
+        args.add(swift_developer_lib_dir_path, format = "-I%s")
+        args.add(swift_developer_lib_dir_path, format = "-L%s")
```
brentleyjones, line 1171: I think this shouldn't be here (compiling doesn't use `-L`)? Also, the name of this function is wrong, it's not linkopts, it will just be the swift imports, so maybe name it `_add_developer_swift_imports`?

swift/internal/developer_frameworks.bzl
```diff
@@ -0,0 +1,119 @@
...
+    return paths.join(
+        paths.dirname(paths.dirname(platform_framework_dir)),
+        "usr",
+        "lib",
+    )
+
+def developer_framework_paths(
+        apple_fragment,
+        xcode_config):
+    """Returns the developer framework paths for the given apple fragment and xcode configuration.
```
brentleyjones, line 68: Please wrap this, and all files to 80 characters. In this case, since it needs to be one line, you can do it like this:
```suggestion
    """Returns the developer framework paths for the given apple fragment and \
    xcode configuration.
```

thii (approved):

swift/internal/linking.bzl
```diff
@@ -142,6 +153,22 @@ def create_linking_context_from_compilation_outputs(
...
         name = label.name
 
+    if is_test:
+        swift_developer_lib_dir_path = swift_developer_lib_dir(
+            apple_common.apple_toolchain(),
+            apple_fragment,
+            xcode_config,
+        )
+        developer_paths_linkopts = [
+            "-I%s" % swift_developer_lib_dir_path,
```
brentleyjones, line 163: Is this needed? I think this is a compile only flag?

brentleyjones (commented):
Thank you for this @maxwellE! One last question below.

brentleyjones (approved):
Thanks for this @maxwellE!

Can you

PR #841, opened 2022-06-10
Update remote swift debugging doc

keith:
As far as I can tell at this point the local debugging options enabled
module is no longer required at least with Swift 5.6+, as long as you
correctly set the framework search paths to include the directories
containing XCTest.

### DIFF
--- a/doc/debuggable_remote_swift.md
+++ b/doc/debuggable_remote_swift.md
@@ -1,15 +1,15 @@
-# Debuggable Remotely Built Swift
+# Debugging Remotely Built Swift
 
 This is a guide to using remotely built Swift modules in local debug builds.
 
-At the time of writing, `lldb` depends on debugging options embedded in `.swiftmodule` files. These options include paths that are only valid on the build host. For local builds, this all just works, but for remote builds, it doesn't.
+By default `lldb` depends on debugging options embedded in `.swiftmodule` files. These options include paths that are only valid on the build host. For local builds, this all just works, but for remote builds, it doesn't.
 
 The solution is two parts:
 
-1. Use `-no-serialize-debugging-options` globally, to prevent embedded paths
-2. Use `-serialize-debugging-options` locally in one empty module
-
-Globally disabling debugging options makes those `.swiftmodule`s usable on any machine. Locally enabling debugging options for one module provides lldb with enough info to make debugging work.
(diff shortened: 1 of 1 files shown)


brentleyjones:
I've seen otherwise. I was going to have to set lots of LLDB settings, and it wouldn't fully cover every situation that `_LocalDebugOptions` would cover.

I'll test again next week and report some detailed findings here. Maybe I was holding it wrong.

keith:
Yea I'd love to see any case you can come up with. I dropped this internally and it seems to work in my testing. Although it's a bit sketchy sometimes because I found I had to kill lldb-rpc-server sometimes to invalidate it's cache

brentleyjones:
I re-read your changes. They are technically correct, since setting `target.swift-framework-search-paths`, `target.clang-module-search-paths`, etc. will work around _most_ issues. I asked about this here though: https://ios-dev-at-scale.slack.com/archives/C019VJ6LK2T/p1651788632722239, and my main concern is if you end up using different flags for PCM compilation, or having same named frameworks at different paths, for various top-level targets. Maybe that will never really happen in practice, but it's been my main hesitation to not telling people about the `_LocalDebugOptions` technique.

keith:
If I understand that conversation correctly I don't think we're really solving that in the original solution or this new solution as we still recommend you disable that globally, so if there isn't 1 set of flags that works for you (that are also the ones passed to the local debug options module), it wouldn't work either way since you're dropping these for all the others regardless? So you could just set the one set of flags that would otherwise apply to your single local module to the lldb setting? I think the real solution in general is https://github.com/bazelbuild/rules_swift/issues/842 but I don't have time to investigate that at the moment

keith:
@brentleyjones bump

brentleyjones:
> Setup a `lldbinit` that restores any of the options that were intended to be set in the `swiftmodule` files

It does require more than you state in the PR description, when dealing with Objective-C dependencies. And `` Use `-serialize-debugging-options` locally in one empty module `` can be a nice solution for some. Doing the `lldbinit` correctly can be tricky: https://github.com/buildbuddy-io/rules_xcodeproj/blob/f95830d20cbee5fb30ccca95db1574cc10844548/tools/generator/src/Generator/CreateFilesAndGroups.swift#L634-L841

But, since rules_xcodeproj now solves this for any use case I'm going to recommend, I'm fine with removing that portion 🤷‍♂️.

PR #847, opened 2022-06-30
Propagate developer SDK paths to Objc providers

### LINKED ISSUE #846: Developer libraries not linked in Apple bundles — 'XCTestSwiftSupport' symbol not found
1. SHA: ac77c2da9a4c24cf0696f56a7941aa0def532da1, or anything post https://github.com/bazelbuild/rules_swift/pull/840
2. Build and test a swift `ios_unit_test`

I see:

```
ld: warning: Could not find or use auto-linked library 'XCTestSwiftSupport'
Undefined symbols for architecture arm64:
  "XCTest.XCTAssertNil(_: @autoclosure () throws -> Any?, _: @autoclosure () -> Swift.String, file: Swift.StaticString, line: Swift.UInt) -> ()", referenced from:
      LoggingLibraryTests.RingBufferTests.testFlush() -> () in libLoggingLibraryTests.library.a(RingBufferTests.swift.o)
...
```

Previous to the change, the developer paths would be added via the `xcode_swift_toolchain` providers, which would then end up in all `Objc` providers. In my testing, manually adding the developer paths to the `Objc` provider resolves the error.

erikkerber:
> **Note:** Experimental solution, looking for design feedback

Resolves #846 

Removing the automatic linking of developer frameworks in #840 may result in undefined symbol errors if the test being built is an Apple bundle. This is because the developer path linkopts weren't being propagated to the `ObjC` providers from the `swift_library` rule.

This solution is inelegant, but it feels like bridging `CcInfo` to/from `ObjC` providers is something that has been hacked around before. Knowing this solution is ~70% of the way there, I'd love feedback on a better way to design this.


swift/internal/swift_protoc_gen_aspect.bzl
```diff
@@ -416,7 +416,7 @@ def _swift_protoc_gen_aspect_impl(target, aspect_ctx):
             copts = ["-parse-as-library"],
             deps = proto_deps + support_deps,
             feature_configuration = feature_configuration,
-            is_test = False,
+            is_test = True,
```
brentleyjones, line 419: 🤨

swift/internal/linking.bzl
```diff
@@ -183,7 +186,7 @@ def create_linking_context_from_compilation_outputs(
         disallow_static_libraries = False,
         disallow_dynamic_library = True,
         grep_includes = None,
-    )
+    ), developer_paths_linkopts
```
brentleyjones, line 189: Because of how CcInfo and Objc currently interopt, I think we need to return this here, and probably not even pass it into `cc_common.create_linking_context_from_compilation_outputs` (does it fail for other cases if you remove it from here?). Setting it on `user_link_flags` even smells bad now that I think about it more. And below, when you set it on `new_objc_provider`, the linkopts should come before the existing ones, to allow user linkopts to override this stuff.

tinder-maxwellelliott (commented):
Thanks for catching this!

swift/internal/linking.bzl
brentleyjones: I think we want it for any that call this API?

erikkerber: In that case it's just a bit of an awkward dataflow, since these paths wouldn't be used within _this_ macro which is primarily for creating those linking contexts. It then feels like this should be part of the `new_objc_provider` macro, as an idea.

I'll get a commit with what I'm thinking about, and will hope the tests (and my test case) can prove one way or the other.

maxwellE (approved):

swift/internal/compiling.bzl
```diff
@@ -2769,7 +2851,10 @@ def new_objc_provider(
         feature_configuration,
         libraries_to_link,
         module_context,
-        user_link_flags = []):
+        user_link_flags = [],
+        apple_fragment,
+        xcode_config,
+        is_test):
```
brentleyjones, line 2857: Can you make these alphabetical?

swift/internal/compiling.bzl
```diff
@@ -2790,6 +2875,9 @@ def new_objc_provider(
         module_context: The module context as returned by
             `swift_common.compile`.
         user_link_flags: Linker options that should be propagated to dependents.
+        is_test: Represents if `testonly` is set on the target to be compiled.
+        apple_fragment: The `apple` configuration fragment.
+        xcode_config: The Xcode configuration.
```
brentleyjones, line 2880: Same

swift/internal/linking.bzl
```diff
@@ -142,6 +147,9 @@ def create_linking_context_from_compilation_outputs(
     if not name:
         name = label.name
 
+    # The developer_paths_linkopts, derived from is_test, need to be propegated
+    # to Objc providers if they are used in this linking context.
+    # Return them in addition to the linking context and output.
```
brentleyjones, line 152: Not needed anymore?
```suggestion
```

swift/internal/linking.bzl
```diff
@@ -23,6 +23,11 @@ load(
     "should_embed_swiftmodule_for_debugging",
 )
 load(":derived_files.bzl", "derived_files")
+load(
+    ":developer_frameworks.bzl",
+    "developer_framework_paths",
+    "swift_developer_lib_dir",
+)
```
brentleyjones, line 30: Not needed anymore?
```suggestion
load(
    ":developer_frameworks.bzl",
    "developer_framework_paths",
    "swift_developer_lib_dir",
)
```

brentleyjones:
#868

PR #854, opened 2022-07-07
Stop automatically setting no private headers feature

keith:
This is a partial commit out of google that was described as:

Enable explicit modules by default

`PACKAGES_USING_EXPLICIT_MODULES_BY_DEFAULT` is replaced with `PACKAGES_NOT_YET_COMPATIBLE_WITH_EXPLICIT_MODULES` and now opts out projects that are not yet migrated.

PiperOrigin-RevId: 444279958
(cherry picked from commit 9da5a9923be7171095eee765a94e6ff7939648d0)


keith:
there are actually other commits that remove this entirely, but they have other conflicts atm

brentleyjones (approved):

PR #857, opened 2022-07-07
Migrate ObjcConfiguration usage to equivalent info in CppConfiguration

keith:
We would like to migrate all such usage to CppConfiguration so that we
can delete the info in ObjcConfiguration.

PiperOrigin-RevId: 457500060
(cherry picked from commit 912d40dade63be23c325ac79a1854868ded16601)

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -78,11 +78,12 @@ def _swift_developer_lib_dir(platform_framework_dir):
         "lib",
     )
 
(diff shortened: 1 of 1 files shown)


brentleyjones:
Requires 6.x.

keith:
they did move it https://github.com/bazelbuild/bazel//commit/8c590d8941e8581ceb502d7ae190b0e5fa480b67

keith:
ok pushed something that should work for backwards compat, and i'll submit a revert PR of it for once we drop 5.x

PR #858, opened 2022-07-07
Use the target triple from the C++ toolchain instead of deriving it from values in the Apple configuration fragment.

keith:
When using `--host_cpu=darwin_x86_64` on an ARM (Apple Silicon) host, the value returned by the Apple configuration fragment is wrong, and we end up compiling Swift for `arm64`. Since the C++ toolchain already has the correct triple verbatim as we want it in the `target_gnu_system_name` field, just use that, and add some generally useful utilities for manipulating the triples and their components.

Note to open-source rules maintainers: This change requires/assumes that your C++ toolchain configuration returns a complete target triple that includes minimum OS version and target environment; for example, `x86_64-apple-ios13.0-simulator`.

PiperOrigin-RevId: 429897884
(cherry picked from commit 62b33ed5ff5d699cb1573309e0714e30f4d6e225)

### DIFF
--- a/swift/internal/target_triples.bzl
+++ b/swift/internal/target_triples.bzl
@@ -1,0 +1,216 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


keith:
https://github.com/bazelbuild/rules_swift/pull/768

brentleyjones (approved):

PR #865, opened 2022-07-26
Fix missing dependency

thii:
Fixes the following compilation error:
```
external/com_github_apple_swift_nio_transport_services/Sources/NIOTransportServices/NIOTSConnectionChannel.swift:20:8: error: no such module 'NIOTLS'
```


segiddins (approved):

keith:
We should build this on CI somewhere, probably as an example dep?

thii:
Looks like it was a non-hermetic bug (it's being built on CI already) https://github.com/bazelbuild/rules_swift/pull/866

PR #868, opened 2022-08-01
Conditionally add developer framework paths

maxwellE:
Condtionally adds developer framework paths to
compile and linking actions. These developer paths
should only be enabled for targets that have
`testonly = True`. Currently they are added for
all targets by default.


brentleyjones:
Can you kickoff a rules_apple build similar to https://github.com/bazelbuild/rules_apple/pull/1511 to ensure this works there?

maxwellE:
> Can you kickoff a rules_apple build similar to [bazelbuild/rules_apple#1511](https://github.com/bazelbuild/rules_apple/pull/1511) to ensure this works there?

Absolutely

maxwellE:
Made a rules_apple pr [here](https://github.com/bazelbuild/rules_apple/pull/1599) with these changes

swift/internal/providers.bzl
```diff
@@ -101,6 +101,20 @@ from the `.proto` files.
     },
 )
 
+SwiftToolchainDeveloperPathInfo = provider(
```
erikkerber, line 104: Does this need to be a provider? I think it could be just a struct.

maxwellE: Struct sounds good to me, will refactor

maxwellE:
Rules apple PR is passing [here](https://github.com/bazelbuild/rules_apple/pulls)

erikkerber (approved):
Verified in our codebase:

- Omitting `testonly` on XCTest targets result in a linker failure
- Adding `testonly` successfully builds

brentleyjones (approved):

brentleyjones (approved):

keith:
This change adds inconsistency between macOS and Linux, where Linux still isn't required to mark tests as `testonly`, can we fix that so it matches the same behavior? Or can we not because of the differences in toolchain layout? (in which case I wonder if we should revert?)

brentleyjones:
> This change adds inconsistency between macOS and Linux, where Linux still isn't required to mark tests as `testonly`, can we fix that so it matches the same behavior

Which paths in linux should be gated behind this? I think we should make the same change if possible, but definitely not revert it, as it's still better even if only one OS is enforcing it. Without this change you can accidentally submit production code that links against test frameworks.

keith:
Not sure which paths, needs investigation into if that's possible, or maybe there's another solution otherwise

keith:
It doesn't appear that there's a way to do this on Linux as the XCTest.swiftmodule is beside the Foundation.swiftmodule. I disagree that it's better to have this on a single platform at the cost of having semantic differences across platforms, but I won't revert this. I would appreciate someone working on making this consistent on Linux.

brentleyjones:
Explicit modules would be able to solve this, as `XCTest` would be labeled as `testonly` in that case. Without that, because of the directory structure of Linux, it doesn't seem like there is a way to enforce `testonly` in the same way.

We could put this behavior behind a flag, but that seems pretty gross to me.

Btw, does this not concern you?

> Without this change you can accidentally submit production code that links against test frameworks.

 (I have to admit, I haven't tried this, and maybe App Store Connect catches this.)

keith:
> We could put this behavior behind a flag, but that seems pretty gross to me.

agreed, def don't want to add a flag for this

> Btw, does this not concern you?

Hrm yea I do think you get caught later for this, I don't recall how.

jerrymarino:
@maxwellE do you have a gist of what motivates this?

It looks like it assumes the invariant of `testonly == "Add xctest bits`. This update has caused a few issues for me. I've got a few scenarios that:
- use testing of `swift_library` without XCTest
- use `XCTest` in other contexts than a test 😅 

I wonder if we could do something a bit different:
- add a `feature` or expose a configurator to turn this on
- other conditional hinges we can govern w/o flipping `testonly`

jerrymarino:
The other situation is for Apple compiling in general - e.g. what about objc_testing?

PR #872, opened 2022-08-08
Use host arch for linux autoconfiguration

dierksen:
The autoconfigured toolchain on linux uses `x86_64` currently. Changing to use the host arch makes it easier to start working on arm64 (for instance... in a docker container on an M1 mac).

It doesn't look like Swift on Windows supports `aarch64` currently; it has the same hard-coded value, but if that changes, it's the same easy fix.


keith (approved):
Thanks!

PR #873, opened 2022-08-08
Fix symbol graph with derived files

chiragramani:
This PR attempts to fix the following issue when `split_derived_files_generation` feature is used.

When using the `split_derived_files_generation` feature, we observed that the `emit-symbol-graph` and `-emit-symbol-graph-dir` options are not passed to the `Generating derived files...` action because of which the symbol.json files were not produced.

Though the flags were passed to the compilation action, it was a no-op because of the issue described here: https://github.com/apple/swift/issues/60412.

This PR does the following:
1. If the derived files feature is not enabled, then the symbol-graph related options are passed to the compilation action. (this is the default).
2. If the derived files feature is enabled, then the symbol-graph related options are passed to the derived files action.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -748,7 +748,7 @@ def compile_action_configs(
             configurators = [_emit_symbol_graph_configurator],
             features = [
                 SWIFT_FEATURE_EMIT_SYMBOL_GRAPH,
-                SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION
+                SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION,
             ],
         ),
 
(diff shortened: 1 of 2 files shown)


keith:
Nice find! Can you sign the CLA, and add a test here? It should cover it to add one validating the args are passed as you expect to the right action. You can probably add it in this file https://github.com/bazelbuild/rules_swift/blob/a707e5491ebcf61e0e6ca7346755695dd1786f47/test/split_derived_files_tests.bzl

chiragramani:
Thanks @keith for the quick review. I have added the tests. Please let me know if there are any other changes needed. Thanks

keith:
can you fix the CI failure here?

keith (approved):

keith:
Thanks!

PR #877, opened 2022-08-15
Add basic sanitizer support for swift_test

keith:
rules_apple handles bundling sanitizers with test bundles that it
creates, swift_test does not handle this as it creates a very minimal
xctest bundle (ideally it wouldn't have to create that bundle at all and
you could run tests directly). Because of this we don't have the
sanitizer libs easily extracted into the bundle. Sanitizers require us
to load them when launching the binary with DYLD_INSERT_LIBRARIES, to
know which to load we can inspect the binary for these. Clang inserts an
absolute rpath to the Xcode directory containing the libraries, as well
as the library loads for each we depend on.

### DIFF
--- a/examples/xplatform/xctest/BUILD
+++ b/examples/xplatform/xctest/BUILD
@@ -9,3 +9,42 @@ swift_test(
         "main.swift",
     ],
 )
+
+swift_test(
+    name = "xctest_tsan",
+    srcs = [
+        "SimpleTests.swift",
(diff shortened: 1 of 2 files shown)


jpsim (approved):
👍 works in my testing with tsan & SwiftLint

Catches races when present, passes tests otherwise.

brentleyjones (approved):

PR #878, opened 2022-08-18
Add `apple.swizzle_absolute_xcttestsourcelocation` feature

brentleyjones:
When this feature is enabled, the `@build_bazel_apple_support//lib:swizzle_absolute_xcttestsourcelocation` target is linked into `swift_test` targets.

### DIFF
--- a/swift/internal/BUILD
+++ b/swift/internal/BUILD
@@ -1,4 +1,5 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//lib:selects.bzl", "selects")
 
 licenses(["notice"])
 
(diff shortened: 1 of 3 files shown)


brentleyjones:
Depends on https://github.com/bazelbuild/apple_support/pull/128

keith (approved):

swift/internal/BUILD
```diff
@@ -404,6 +404,30 @@ bzl_library(
     srcs = ["build_settings.bzl"],
 )
 
+config_setting(
+    name = "can_swizzle_xcttestsourcelocation",
+    # Ideally this would detect building for an Apple platform
+    values = {
+        "features": "apple.swizzle_absolute_xcttestsourcelocation",
+    },
+)
```
brentleyjones, line 413: Ideally this would detect building for an Apple platform. As it stands, if someone on Linux enables this feature, they will get very weird errors about darwin cpu usage.

@keith Do you have an idea for a better check? Do we have an `is_apple` `config_setting`?

keith: what if you put this feature in the unsupported features of the toolchain on linux?

keith: is that supported if it was using the `swift.` prefix logic?

brentleyjones: No. `config_setting.values` looking at the `features` build setting is only affected by the command-line and transitions. If a transition doesn't remove `apple.swizzle_absolute_xcttestsourcelocation` from `features`, it will always exist, causing this config to be true, and the `alias` to resolve to an `objc_library`. Adding something to the unsupported features will prevent it from being enabled in the `is_enabled` checks (which we don't use, but could if it had a `swift.` prefix), but I believe that's too late for the darwin cpu analysis phase failure.

keith: if we annotated the apple_support target with target_compatible_with, could you check that before adding it?  (might only work with bazel 6.x)

brentleyjones: Went with an Envoy inspired `apple` `config_setting` for now. I think we could do some better platform based things in the future, when the apple rules fully support and by default use platforms.

keith (approved):

PR #886, opened 2022-08-29
Supporting Swift 5.7 Bare Slash Regex Literal

chiragramani:
This PR adds support for passing the `-enable-bare-slash-regex` flag by default to all the swiftc actions for Swift 5.7 and above.

**Context**

The following fails to compile on Xcode 14 toolchain:

```
func regexTest() {
    let testString = "04/07/2022"
    let digits = /\d+/
    print(testString.ranges(of: digits))
}
```
**Analysis**

Swift evolution proposal: [Regex Literals](https://github.com/apple/swift-evolution/blob/main/proposals/0354-regex-literals.md)

As per the proposal, it explicitly needs the `-enable-bare-slash-regex` flag to successfully compile the above code. Xcode 14 passes the same by default.


brentleyjones (approved):

thii (approved):

brentleyjones:
Thanks @chiragramani!

PR #889, opened 2022-08-30
Use runfiles for `index-import` path

brentleyjones:
Addresses https://github.com/bazelbuild/rules_swift/pull/567/files#r645580328

### DIFF
--- a/swift/internal/swift_toolchain.bzl
+++ b/swift/internal/swift_toolchain.bzl
@@ -354,7 +354,7 @@ def _swift_toolchain_impl(ctx):
             ],
             requested_features = requested_features,
             root_dir = toolchain_root,
-            swift_worker = ctx.executable._worker,
+            swift_worker = ctx.attr._worker[DefaultInfo].files_to_run,
             test_configuration = struct(
                 env = env,
                 execution_requirements = {},
--- a/tools/worker/worker_main.cc
+++ b/tools/worker/worker_main.cc
@@ -17,10 +17,23 @@
 #include <string>
 #include <vector>
 
+#include "tools/cpp/runfiles/runfiles.h"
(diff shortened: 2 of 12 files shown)


swift/internal/swift_toolchain.bzl
```diff
@@ -354,7 +354,7 @@ def _swift_toolchain_impl(ctx):
             ],
             requested_features = requested_features,
             root_dir = toolchain_root,
-            swift_worker = ctx.executable._worker,
+            swift_worker = ctx.attr._worker[DefaultInfo].files_to_run,
```
brentleyjones, line 357: This was a head-scratcher.

tools/worker/worker_main.cc
```diff
@@ -17,10 +17,23 @@
...
+#include "tools/cpp/runfiles/runfiles.h"
 #include "tools/worker/compile_with_worker.h"
 #include "tools/worker/compile_without_worker.h"
 
+using bazel::tools::cpp::runfiles::Runfiles;
+
 int main(int argc, char *argv[]) {
+  std::string error;
+  std::unique_ptr<Runfiles> runfiles(Runfiles::Create(argv[0], &error));
+  if (!runfiles) {
```
thii, line 29: The [sample code](https://github.com/bazelbuild/bazel/blob/bff973031a65c139a4878c87e5f335a895964ab9/tools/cpp/runfiles/runfiles_src.h#L44) uses this. Do we want to change it?
```suggestion
  if (runfiles == nullptr) {
```

thii (approved):

tools/worker/worker_main.cc
brentleyjones: Was following the style of our codebase, but I don't know enough about C++, so I'll change it back.

Issue #869, opened 2022-08-03
State of feature `swift.use_c_modules`


dierksen:
We're looking into using explicit clang modules, but it looks like right now the swift stdlib isn't accessible when the feature is enabled.

I saw https://github.com/bazelbuild/rules_swift/pull/856 had gone in recently, so wanted to see if there was any plan for this in the near-term, or if the necessary toolchain modifications already exist and just didn't make it over to github.

## Current state

Building an extremely simple hello world `swift_binary` with `--features swift.use_c_modules` results in
```
<unknown>:0: error: missing required module 'SwiftShims'
```

Happy to help build this out, but wanted to check and see if the work had already been started/done. 😃

keith:
I believe this set of features is being used inside google, but not currently publicly. The biggest missing piece I know of currently is to generate the definitions for the system modules, at Lyft we started working on that a while ago but tabled it so we don't have anything to share there yet either. I think if you had those definitions and dependencies setup correctly the rules would work with them.

dierksen:
OK, I'm looking at just adding a `swift_c_module` (with some tweaks) for `SwiftShims` since that seems to fix the "Hello world" case at least. Possibly a little tricky since the shims module is defined per-platform (i.e. `/Applications/Xcode_13.4.1.app/Contents/Developer/Platforms/AppleTVOS.platform/Developer/SDKs/AppleTVOS.sdk/usr/lib/swift/shims` but shouldn't be too hard to work with.

keith:
Yea I think for the productionized version of this it would make sense to generate the BUILD file for the system frameworks as a repository rule

PR #895, opened 2022-09-06
Enable precompiling swift implicit dep PCMs

dierksen:
- Modify `swift_c_module` to enable behavior for system framework PCMs
- add bzl test to ensure it stays functional
- modified expected arg behavior to cover cases like `-fmodule-file=ModuleName=/indeterminate/path/to/module`

After opening #869 I started looking at what it would take to enable this. At base level, everything has to depend on `SwiftShims`, but it quickly balloons as soon as you start adding additional frameworks. The test I added here is just a simple proof of concept for making this work on macos/iOS only.

Since going any further than this would probably require adding additional flags on `swift_library` (not to mention splitting the logic further between linux/mac), I wanted to start with a very simple case to make sure that the necessary build logic is exercised in CI.

Future work I'm looking at around this:
- create a repo rule to scan all the swift framework deps in local versions of xcode (example: https://github.com/dierksen/swift_scratch/blob/main/framework_repos.bzl)
- potentially add an attribute for `sdk_frameworks` (like objc uses for linking), but this might go in `rules_ios` if it doesn't make sense here
- make sure that it all works on linux as well and add tests


swift/internal/swift_c_module.bzl
```diff
@@ -85,18 +124,19 @@ file, so it must be provided directly. Therefore, one may have multiple
...
-        "deps": attr.label_list(
-            allow_empty = False,
-            doc = """\
+                mandatory = True,
+            ),
+            "deps": attr.label_list(
+                allow_empty = True,
+                doc = """\
 A list of C targets (or anything propagating `CcInfo`) that are dependencies of
 this target and whose headers may be referenced by the module map.
```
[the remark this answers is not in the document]
dierksen: Looks like this change is unnecessary; I probably put that in when I was troubleshooting something previously and never removed it. Reverted.

dierksen:
Re-requesting review removed everyone else, and now I can't re-add them 🤦

.bazelci/presubmit.yml
```diff
@@ -55,11 +55,21 @@ tasks:
...
     shell_commands:
-    # Update the WORKSPACE to use head versions of some deps to ensure nothing
-    # has landed on them breaking this project.
-    - .bazelci/update_workspace_to_deps_heads.sh
+      # Update the WORKSPACE to use head versions of some deps to ensure nothing
+      # has landed on them breaking this project.
+      - .bazelci/update_workspace_to_deps_heads.sh
     <<: *mac_common
 
+  macos_swift_explicit_modules:
```
keith, line 63: instead of making a separate CI job here it would be best to use a bzl test that forces the features you care about https://github.com/bazelbuild/rules_swift/blob/ae6a0b602784c947f5c5fb22f874ffce4ed47889/test/debug_settings_tests.bzl#L24-L26 and depend on a fixture target, which will also build it

examples/apple/swift_explicit_modules/BUILD
```diff
@@ -0,0 +1,13 @@
+load("//swift:swift.bzl", "swift_binary", "swift_c_module")
+
+swift_c_module(
+    name = "shims",
+    module_name = "SwiftShims",
```
jerrymarino, line 5: 🚀

.bazelci/presubmit.yml
dierksen: Oh nice, I hadn't messed with these before, but that's really slick for testing the behavior from both sides.

swift/internal/swift_c_module.bzl
```diff
@@ -48,32 +81,43 @@ def _swift_c_module_impl(ctx):
...
-into Swift.
+into Swift. This is mutally exclusive with `system_module_map`.
 """,
-            mandatory = True,
-        ),
-        "module_name": attr.string(
-            doc = """\
+                mandatory = False,
+            ),
+            "system_module_map": attr.string(
```
jerrymarino, line 108: My other question - pertaining to first party modules: do you anticipate a need for `copts` and `defines`? Say, Say, I've got some headers with conditional compilation hinged by defines that clangimporter needs to set. This may not be necessary for SDK level stuff and is potentially forward looking 👁️

jerrymarino (approved):
Nice one @dierksen - can't wait to see how this works out!

swift/internal/swift_c_module.bzl
dierksen: It's possible, but this is really mostly focused on enabling the SDK frameworks. There aren't many examples in the wild of the `swift_c_module` rule in use to see how necessary it might be (at least by a cs.github.com search that mostly shows forks of this repo).

If we do need to add much more, it's probably worth tweaking `attrs.bzl` to get the necessary shared attributes combined and validated.

dierksen:
I've got it mostly working on linux as well, but I'm not as sure on all the settings: https://github.com/dierksen/rules_swift/pull/1

Mostly just dropping that here for posterity as it's not something I'm planning to pursue a ton just now.

jerrymarino:
@dierksen awesome! Yeah for rules_ios the conditional compilation in headers is something we can sort out later as we'll need it for our general iOS framework building rules. The helper rules you've got in this PR should get SDK side of the [explicit PCM story](https://github.com/bazel-ios/rules_ios/issues/484) locked down

swift/internal/swift_c_module.bzl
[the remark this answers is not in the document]
thiagohmcruz: Yep, either one should be fine, as I said it's very NIT but better to be consistent IMO

examples/apple/swift_explicit_modules/BUILD
```diff
@@ -0,0 +1,18 @@
+load("//swift:swift.bzl", "swift_binary", "swift_c_module")
```
keith, line 1: might not need this one anymore with the test fixtures instead?

swift/internal/swift_c_module.bzl
```diff
@@ -14,23 +14,56 @@
...
+        (not ctx.file.module_map and not ctx.attr.system_module_map)
+    ):
+        fail("Must specify one (and only one) of module_map and system_module_map.")
+    swift_toolchain = ctx.attr._toolchain[SwiftToolchainInfo]
+    module_map = ctx.file.module_map or ctx.attr.system_module_map
+    is_system = True if ctx.attr.system_module_map else False
 
     deps = ctx.attr.deps
     cc_infos = [dep[CcInfo] for dep in deps]
+    swift_infos = [dep[SwiftInfo] for dep in deps if SwiftInfo in dep]
```
keith, line 36: if this bit is required for this to work I don't think this is covered by tests? is there something you could add to cover it if so?

keith (approved):
looks good! 2 minor comments

examples/apple/swift_explicit_modules/BUILD
dierksen: Yeah, it's not strictly necessary. I tweaked it into just an example to demonstrate how to use them, but it's definitely not critical at this point.

keith: ok sounds fine to keep

swift/internal/swift_c_module.bzl
dierksen: Good callout; I added another case that has dependencies that will exercise this.

brentleyjones (approved):

keith (approved):

keith:
Thanks!

PR #896, opened 2022-09-06
Remove space-in-path replacement

brentleyjones:
This replacement causes issues with importing indexstores.

In 965c37304cbc0263b6a5b0a945529c66a593c34c this replacement was added to work around issues with `ar_wrapper`. I don't see `ar_wrapper` used anywhere in Bazel, so I believe this is safe to remove.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1448,7 +1448,7 @@ def _declare_per_source_output_file(actions, extension, target_name, src):
         The declared `File`.
     """
     objs_dir = "{}_objs".format(target_name)
-    owner_rel_path = owner_relative_path(src).replace(" ", "__SPACE__")
+    owner_rel_path = owner_relative_path(src)
     basename = paths.basename(owner_rel_path)
     dirname = paths.join(objs_dir, paths.dirname(owner_rel_path))
 


brentleyjones:
Seems I'll need to find a way to escape some paths in tool invocations, and if not possible give up ☹️.

thii:
> This replacement causes issues with importing indexstores.

What issues does it cause?

brentleyjones:
Unit files have a hash based on the object file path. When we import unit files we need to make the object file path equal to what it would be if Xcode generates them. Xcode will generate these paths with spaces in their name, while rules_swift will replace the spaces.

So during an import we would have to undo this, but with index-import (which is just regex) we can't. Even if we could, theoretically there are some paths out there with `__SPACE__` in their name.

brentleyjones:
And after saying all of that, I think I'll add something to index-import to address this for the time being.

brentleyjones:
https://github.com/MobileNativeFoundation/index-import/pull/65

We can merge this PR once we drop Bazel 5.x support.

mattrobmattrob (approved):

luispadron (approved):

brentleyjones:
Looks like there are new failures. I'll see if I'm able to address them.

brentleyjones:
So clang (at least on linux) requires quotes for spaces in the params file. Seeing if I can influence that, or if it will require a Bazel patch.

brentleyjones:
It looks like enabling the `gcc_quoting_for_param_files` feature should fix this, but I can't get it to change anything locally.

brentleyjones:
https://github.com/bazelbuild/bazel/pull/21699

keith:
is this fixed with your upstream fix?

luispadron (approved):

PR #901, opened 2022-09-11
Fix unable to opt out of `swift.use_response_files` feature

thii:


keith (approved):
would be nice to test if possible

thii:
Dropping this - didn't see any improvements, also this was removed upstream.

PR #904, opened 2022-09-14
Add support for -file-prefix-map $PWD=.

keith:
This is a new flag in Swift 5.7, that mirrors the flag with the same
name in clang, meaning that it implies both -debug-prefix-map, and
-coverage-prefix map. Currently it is also the only flag that applies to
indexing info if you want to produce hermetic indexes. Ideally going
forward users should prefer this flag if they want their builds to be
hermetic instead of picking the other flags as needed, since this will
potentially also apply to new absolute paths in the future.

https://github.com/apple/swift/pull/58946

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -57,6 +57,14 @@ SWIFT_FEATURE_DEBUG_PREFIX_MAP = "swift.debug_prefix_map"
 # of remote builds.
 SWIFT_FEATURE_COVERAGE_PREFIX_MAP = "swift.coverage_prefix_map"
 
+# If enabled, builds will use the `-file-prefix-map` feature to remap the
+# current working directory to `.`, which avoids embedding non-hermetic
+# absolute path information in build artifacts. Specifically what this flag
+# does is subject to change in Swift, but it should imply all other
+# `-*-prefix-map` flags. How those flags compose is potentially complicated, so
+# using only this flag, or the same values for each flag, is recommended.
+SWIFT_FEATURE_FILE_PREFIX_MAP = "swift.file_prefix_map"
+
 # If enabled, C and Objective-C libraries that are direct or transitive
 # dependencies of a Swift library will emit explicit precompiled modules that
 # are compatible with Swift's ClangImporter and propagate them up the build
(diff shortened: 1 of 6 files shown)


swift/internal/feature_names.bzl
```diff
@@ -57,6 +57,14 @@ SWIFT_FEATURE_DEBUG_PREFIX_MAP = "swift.debug_prefix_map"
 # of remote builds.
 SWIFT_FEATURE_COVERAGE_PREFIX_MAP = "swift.coverage_prefix_map"
 
+# If enabled, builds will use the `-file-prefix-map` feature to remap the
+# current working directory to `.`, which avoids embedding non-hermetic
+# absolute path information in build artifacts. Specifically what this flag
+# does is subject to change in Swift, but it should imply all other
+# `-*-prefix-map` flags. How those flags compose is potentially complicated, so
+# using only this flag, or the same values for each flag, is recommended.
```
brentleyjones, line 65: Do we instead want to add `not_features = [SWIFT_FEATURE_FILE_PREFIX_MAP]` to all the other spots and say in this comment that enabling this prevents the other features from doing anything?

keith: hrm yea good question. given how we use these in the rules, the values should always be the same, so it's _probably_ fine, but idk. i guess it's a question of trust in the swiftc behavior in general

keith: thinking about it more, i guess i like not disabling the others based on this just to reduce the feature matrix related to this, unless it becomes an actual issue

tools/worker/swift_runner.cc
```diff
@@ -284,6 +284,12 @@ bool SwiftRunner::ProcessArgument(
         consumer("-coverage-prefix-map");
         consumer(std::filesystem::current_path().string() + "=.");
         changed = true;
+      } else if (new_arg == "-file-prefix-pwd-is-dot") {
+        // Get the actual current working directory (the workspace root), which
```
brentleyjones, line 288: FYI, these comments are wrong, since the working directory is the bazel execution root, not the workspace root.

thii (approved):

PR #909, opened 2022-09-28
Move the internal `new_objc_provider` helper function from `compiling.bzl` to `linking.bzl`.

thii:
The `Objc` provider is only created by Swift to propagate linking information to the Apple rules until that logic migrates to `CcInfo`, so this is a better home for it until it can disappear for good.

PiperOrigin-RevId: 423827285
(cherry picked from commit 899eb1e5ca6f5c8b488dbbe4cb64653ccf77604f)

 Conflicts:
	swift/internal/compiling.bzl
	swift/internal/linking.bzl
	swift/internal/swift_grpc_library.bzl
	swift/internal/swift_import.bzl
	swift/internal/swift_library.bzl
	swift/internal/swift_module_alias.bzl
	swift/internal/swift_protoc_gen_aspect.bzl

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -24,7 +24,6 @@ load(
     "run_toolchain_action",
     "swift_action_names",
 )
-load(":debugging.bzl", "should_embed_swiftmodule_for_debugging")
 load(":derived_files.bzl", "derived_files")
 load(
     ":feature_names.bzl",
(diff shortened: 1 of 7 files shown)


thii:
I've made sure to move our version of `new_objc_provider` since it's a little bit different from `upstream`.

keith (approved):

PR #910, opened 2022-09-28
Enable -wmo with opt by default

keith:
This matches Xcode's behavior and has a huge impact on performance in
some cases.


brentleyjones (approved):

jpsim:
I saw a 90% improvement in SwiftLint lint times with this. +1

PR #912, opened 2022-09-30
Filter out non-standard header extensions from module maps

dierksen:
In https://github.com/bazel-ios/rules_ios/pull/562, I realized that the hmap and vfsoverlay files are passed as `headers` on `CcInfo` as a workaround for the lack of custom providers in the Java-side rules. As a result, `rules_swift` attempts to add them to the modulemaps, which leads to failures.

### DIFF
--- a/swift/internal/module_maps.bzl
+++ b/swift/internal/module_maps.bzl
@@ -93,7 +93,7 @@ def write_module_map(
 
     def _add_headers(*, headers, kind):
         content.add_all(
-            headers,
+            [h for h in headers if h.extension.lower() in ("h", "hpp", "hh")],
             allow_closure = True,
             format_each = '    {} "%s"'.format(kind),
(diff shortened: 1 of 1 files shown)


swift/internal/module_maps.bzl
```diff
@@ -93,7 +93,7 @@ def write_module_map(
 
     def _add_headers(*, headers, kind):
         content.add_all(
-            headers,
```
jerrymarino, line 96: Is this ending up being called in `rules_swift` ? I'd expect this to [not be called](https://github.com/bazel-ios/rules_ios/blob/master/rules/library.bzl#L986) - I wonder if there's been a regression in rules_swift - or it's not conditionally flipped off

dierksen:
Probably better fix via https://github.com/bazel-ios/rules_ios/pull/603

PR #917, opened 2022-10-10
Add support for imported_library when linking

keith:
Required for https://github.com/bazelbuild/rules_apple/pull/1698 when
linking swift_binary targets


brentleyjones (approved):

brentleyjones:
In a rules_apple that doesn't have https://github.com/bazelbuild/rules_apple/pull/1698, does this result in a double link?

keith:
nah since it's not added to imported_library, and it's an either or with both setups

PR #920, opened 2022-10-14
Add initial bzlmod support

BalestraPatrick:
This is an initial implementation of bzlmod support in `rules_swift`. Still some things left to figure out, but I was able to compile a basic iOS app with WIP implementations of `rules_apple`, `rules_swift` and `apple_support`.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,31 @@
+module(
+    name = "rules_swift",
+    compatibility_level = 1,
+    repo_name = "build_bazel_rules_swift",
+    version = "1.2.0",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "apple_support", repo_name = "build_bazel_apple_support", version = "1.3.1")
+bazel_dep(name = "rules_cc", version = "0.0.2")
+bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "3.19.2")  # To be removed once rules_proto is bzlmod-ready.
+
+non_module_deps = use_extension("//swift:extensions.bzl", "non_module_deps")
+
+use_repo(
+    non_module_deps,
+    "build_bazel_rules_swift_local_config",
+    "com_github_apple_swift_protobuf",
+    "com_github_grpc_grpc_swift",
+    "com_github_apple_swift_nio",
+    "com_github_apple_swift_nio_http2",
+    "com_github_apple_swift_nio_transport_services",
+    "com_github_apple_swift_nio_extras",
+    "com_github_apple_swift_log",
+    "com_github_nlohmann_json",
+    "rules_proto",
+    "build_bazel_rules_swift_index_import",
+)
+
+# Dev dependencies
+bazel_dep(name = "stardoc", dev_dependency = True, repo_name = "io_bazel_skydoc", version = "0.5.3")
--- a/swift/extensions.bzl
+++ b/swift/extensions.bzl
@@ -1,0 +1,22 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,28 @@
...
+    name = "rules_swift",
+    version = "1.2.0",
+    compatibility_level = 1,
+    repo_name = "build_bazel_rules_swift",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "apple_support", version = "1.3.1", repo_name = "build_bazel_apple_support")
+bazel_dep(name = "rules_cc", version = "0.0.2")
+bazel_dep(name = "protobuf", version = "3.19.2", repo_name = "com_google_protobuf") # To be removed once rules_proto is bzlmod-ready.
```
BalestraPatrick, line 11: At the moment `rules_swift` uses a newer version of `rules_proto` (with even a custom patch) that is not yet in the central-registry. For this reason, we need to specify a dependency on `protobuf` here and load `rules_proto` through a module extension for now. Once `rules_proto` fully supprots bzlmod, we can simply add `bazel_dep(name = "rules_proto", version = "6.0")` and be done with it.

MODULE.bazel
```diff
@@ -0,0 +1,28 @@
...
+
+non_module_deps = use_extension("//swift:extensions.bzl", "non_module_deps")
```
BalestraPatrick, line 13: Through a module extension we can load dependencies that are not yet exposing modules, together with the `swift_autoconfiguration` repository rule. Over time these dependencies should be part of the central registry, but some of them could also stay here forever?

swift/extensions.bzl
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
```
BalestraPatrick, line 1: I'm open to moving this file elsewhere in the repo too, I just took inspiration from what other rulesets have been doing:
- https://github.com/bazelbuild/rules_cc/blob/main/cc/extensions.bzl
- https://github.com/bazelbuild/rules_jvm_external/blob/master/extensions.bzl

swift/repositories.bzl
```diff
@@ -31,31 +31,32 @@ def _maybe(repo_rule, name, **kwargs):
     if not native.existing_rule(name):
         repo_rule(name = name, **kwargs)
 
-def swift_rules_dependencies():
+def swift_rules_dependencies(include_bzlmod_ready_dependencies = True):
```
BalestraPatrick, line 34: This is a way to keep compatibility with non-bzlmod projects, but also slowly start moving dependencies to bzlmod as they become compatible.

MODULE.bazel
keith: this is interesting! yea so what's the long term move here, out of this list there are very few that i would expect to make it to the central registry, is there no provided abstraction for this kind of case?

BalestraPatrick: That's my understanding at least. The [bzlmod migration guide](https://docs.google.com/document/d/1JtXIVnXyFZ4bmbiBCr5gsTH4-opZAFf5DMMb-54kES0/edit#heading=h.axme6pa0eunx) mentions this:
>**Module extensions should be used when introducing the following type of dependencies:**
...
Introduce “internal dependencies”: Sometimes you may have dependencies that don't make sense to be Bazel modules, nor to be introduced via module extensions that integrate third party package managers, eg. downloading a data file or some specific dependency that will only be used in your own project (no resolution needed). You can introduce them as “internal dependencies” by invoking repository rules in a module extension that’s used only in your project, and they won’t be visible to other repositories in your dependency graph

Anything that should be natively supported with bzlmod needs to be maintained in the BCR, so if a dependency isn't supported by default in Bazel (no `BUILD.bazel` checked-in in their repos for example), I don't see a reason to add it o the BCR.

keith (approved):

BalestraPatrick:
I'll merge this and we can keep on iterating in follow-ups.

PR #924, opened 2022-10-22
Transition `--proto_compiler` instead of patching `rules_proto`

thii:
This avoids having to patch `rules_proto` in order to support the `@build_bazel_rules_swift//swift:universal_tools` flag.

One caveat is that this won't apply if `proto_library` is a top-level target.


thii:
The Windows failure seems unrelated https://buildkite.com/bazel/rules-swift-swift/builds/3782#0183fde5-427c-4a60-8eb0-7e8187783018

keith (approved):
seems nicer.  i  guess we should recommend folks pass this flag in the case they want it always, given the caveat about how this won't apply to top level targets (or those through cc_library)

thii:
I'll update the docs in a follow-up.

PR #927, opened 2022-10-26
Use `Label(...)` in proto transition

mattrobmattrob:
# Summary
Use `Label` to reference the repo value for `//tools/protoc_wrapper:protoc` in the `proto_compiler_transition` transition implementaiton.

The errors introduced in our use case in https://github.com/bazelbuild/rules_swift/commit/1a86b0292de737e1251118e6738eba25c7037ad9 + https://github.com/bazelbuild/rules_swift/pull/924:
```
ERROR: /private/var/tmp/_bazel_matt.robinson/4eb5286194401ed417482737bd8dd3cc/external/foo/BUILD.bazel:3:27: every rule of type proto_library implicitly depends upon the target '//tools/protoc_wrapper:protoc', but this target could not be found because of: no such package 'tools/protoc_wrapper': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
```

### DIFF
--- a/swift/internal/swift_grpc_library.bzl
+++ b/swift/internal/swift_grpc_library.bzl
@@ -409,14 +409,14 @@ The kind of definitions that should be generated:
             "_protoc": attr.label(
                 cfg = "exec",
                 default = Label(
-                    "@build_bazel_rules_swift//tools/protoc_wrapper:protoc",
+                    "//tools/protoc_wrapper:protoc",
                 ),
                 executable = True,
             ),
             "_protoc_gen_swiftgrpc": attr.label(
                 cfg = "exec",
                 default = Label(
-                    "@build_bazel_rules_swift//tools/protoc_wrapper:protoc-gen-grpc-swift",
+                    "//tools/protoc_wrapper:protoc-gen-grpc-swift",
                 ),
                 executable = True,
             ),
--- a/swift/internal/swift_grpc_library.bzl
+++ b/swift/internal/swift_grpc_library.bzl
@@ -409,14 +409,14 @@ The kind of definitions that should be generated:
             "_protoc": attr.label(
                 cfg = "exec",
                 default = Label(
-                    "//tools/protoc_wrapper:protoc",
+                    "@build_bazel_rules_swift//tools/protoc_wrapper:protoc",
                 ),
                 executable = True,
             ),
             "_protoc_gen_swiftgrpc": attr.label(
                 cfg = "exec",
                 default = Label(
-                    "//tools/protoc_wrapper:protoc-gen-grpc-swift",
+                    "@build_bazel_rules_swift//tools/protoc_wrapper:protoc-gen-grpc-swift",
                 ),
                 executable = True,
             ),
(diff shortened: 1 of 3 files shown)


keith:
should i be surprised that both work on our CI? (assuming this is getting hit)

mattrobmattrob:
Wouldn't it always work on CI since it's running in the workspace context of `rules_swift` where https://github.com/bazelbuild/rules_swift/blob/master/tools/protoc_wrapper/BUILD exists? Perhaps I'm misunderstanding how this should work.

mattrobmattrob:
My guess is that this change is still correct (in the case where this flag isn't provided) but using `--proto_compiler=@build_bazel_rules_swift//tools/protoc_wrapper:protoc` fixes the issue for me since the transition doesn't return `//tools/protoc_wrapper:protoc`.

keith:
I'm not really sure, there are other cases like this though https://github.com/bazelbuild/rules_swift/blob/fd39daf5f5e000c162f5033e71c489557f2b6bfb/swift/internal/swift_toolchain.bzl#L426

I know some things change with bzlmod with this too, and this label may no longer work in that case. @thii what do you think here?

mattrobmattrob:
Are there other cases in a transition? It might just be this that's necessary: https://github.com/bazelbuild/rules_swift/pull/927/commits/185c31660c8a0ade82a756a3c8205749de33f508#diff-b0b4458451f8ecb5c84c21a380933cbc72b01d6ba9a3e4a1f77d2d0f8f80d3f9R24

swift/internal/swift_grpc_library.bzl
```diff
@@ -409,14 +409,14 @@ The kind of definitions that should be generated:
             "_protoc": attr.label(
                 cfg = "exec",
                 default = Label(
-                    "//tools/protoc_wrapper:protoc",
+                    "@build_bazel_rules_swift//tools/protoc_wrapper:protoc",
                 ),
```
brentleyjones, line 413: `Label(//a:target)` is supposed to resolve to the repo that the `bzl` file is in. So I'm surprised this is needed.

brentleyjones:
> It might just be this that's necessary: https://github.com/bazelbuild/rules_swift/commit/185c31660c8a0ade82a756a3c8205749de33f508#diff-b0b4458451f8ecb5c84c21a380933cbc72b01d6ba9a3e4a1f77d2d0f8f80d3f9R24

Yeah, I think this is the one that is needed. Can we set a `Label` value here as well? I think the `@build_bazel_rules_swift` string version isn't ideal with bzlmod.

mattrobmattrob:
Generated a simple repro case for the issue in https://github.com/mattrobmattrob/rules_swift_testbed/tree/mr/rules_swift/pr/927#summary

And confirmed just the transitioned needed to be updated and `Label` works fine. 👍

brentleyjones (approved):

thii:
@mattrobmattrob Thanks for the fix! 🙇‍♂️

thii:
`@bazel_tools//tools/proto:protoc` is an alias to `@com_google_protobuf//:protoc`, I think we will need to cover that case too. I'm not sure if there is a better way than comparing strings like this.

PR #940, opened 2022-11-29
Add missing `bzl_library` targets

brentleyjones:


keith (approved):
looks good but would be nice if we fixed why ever this didn't catch it here, maybe we're missing some docs

PR #943, opened 2022-11-30
Skip building universal targets by default

thii:
Since `--cpu` won't change unless a `platform_mappings` file is set up,
these won't build.


keith (approved):
as long as these still get built transitively on CI this sgtm. although adding a platform_mappings file would also be fine imo

PR #945, opened 2022-12-01
Fix incremental compilation when generated header is nested in a directory

brentleyjones:


tools/worker/work_processor.cc
```diff
@@ -148,6 +149,20 @@ void WorkProcessor::ProcessWorkRequest(
   std::ostringstream stderr_stream;
 
   if (is_incremental) {
+    std::set<std::string> dir_paths;
```
brentleyjones, line 152: This should overall be more performant now, not trying to create the same directory multiple times.

keith (approved):
nice. it sounds like you could add an example with a nested dir + header to slightly regression test this?

keith (approved):

PR #955, opened 2022-12-16
Test --incompatible_disable_starlark_host_transitions

keith:


keith:
looks good with the skylib update

PR #960, opened 2022-12-23
[bzlmod] Add dependency on platforms

BalestraPatrick:
This was discovered while registering the latest rules_swift release into the BCR and running the validation pipelines there. It was tricky to debug, but it seems like either a recent change in rules_swift or bzlmod and Bazel caused this to be required.

https://github.com/bazelbuild/bazel-central-registry/pull/327

Now that bzlmod is more stable, we should likely look into adding a bzlmod verification pipeline directly in this repo so we can catch regressions earlier.


brentleyjones (approved):

brentleyjones:
> Now that bzlmod is more stable, we should likely look into adding a bzlmod verification pipeline directly in this repo so we can catch regressions earlier.

I agree.

PR #962, opened 2023-01-04
Add generate_internal_visibility_from_proto_files feature

### LINKED ISSUE #961: [swift_proto_library] Generated types are always public
Currently the [`Visibility` generation option](https://github.com/apple/swift-protobuf/blob/main/Documentation/PLUGIN.md#generation-option-visibility---visibility-of-generated-types) is hardcoded to `Public` for all protobuf generation: https://github.com/bazelbuild/rules_swift/blob/7ea19709611d40776f05bf6d35b950e1d1a3a506/swift/internal/swift_protoc_gen_aspect.bzl#L185

Certain use cases require a non-public access level for generated objects. For example, when creating a library without exposing messages to the consumer.

tylervick:
Fixes #961

**Change**
Create a new feature `swift.generate_internal_visibility_from_proto_files` which will set protoc's swift_opt `Visibility` from `Public` to `Internal`.

**Test**
- Create a valid rule for `swift_grpc_library` or `swift_proto_library`
  - Verify the generated sources still have the "public" access modifier
- Add the `swift.generate_internal_visibility_from_proto_files` to the `features` attribute for the above rule
  - Verify the generated sources now omit access modifiers (implicitly `internal`)

**Note**
The ideal solution would be to make the `Public` setting opt-in for this feature, thereby mirroring default protoc Swift plugin behavior. Since rules_swift has always defaulted to public visibility this should instead remain the default.

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -295,6 +295,11 @@ SWIFT_FEATURE_GENERATE_FROM_RAW_PROTO_FILES = "swift.generate_from_raw_proto_fil
 # generating a Swift file from a proto file.
 SWIFT_FEATURE_GENERATE_PATH_TO_UNDERSCORES_FROM_PROTO_FILES = "swift.generate_path_to_underscores_from_proto_files"
 
+# If enabled, the toolchain will use `--swift_opt=Visibility=Internal`
+# (instead of `--swift_opt=Visibility=Public`) for the protoc Swift plugin arguments when
+# generating a Swift file from a proto file.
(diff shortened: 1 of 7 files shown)


mattrobmattrob (changes requested):
Thanks for the addition, @tylervick!

Two things:
1. Mind adding an example/validation of this flag similar to `//examples/xplatform/proto:example_path_to_underscores_proto` with `features = ["swift. generate_internal_visibility_from_proto_files"]` that validates this functionality?
2. Can you move the `//third_party/...` formatting changes to a separate PR?

brentleyjones:
The proto rules have been completely rewritten. I don't think this is needed anymore.

PR #963, opened 2023-01-05
Fix runfiles lookup under bzlmod in Bazel 6

BalestraPatrick:
When building Swift code with `--enable_bzlmod`, it currently fails due to the worker failing to find the `index-import` executable in the runfiles. This is because the naming is different under bzlmod and the C++ runfiles library was updated to support this new mapping as described in this [talk](https://www.youtube.com/watch?v=5NbgUMH1OGo). Since rules_swift supports both Bazel 5 and 6, we need to check `BAZEL_CURRENT_REPOSITORY` is defined to  simply make use of the new constructor which will make it so the runfiles lookup will happen relative to the `rules_swift` repo under Bazel 6.

I tested this locally with a `local_path_override` while iterating on getting the `rules_xcodeproj` to work under bzlmod (and also continues to work with bzlmod disabled in Bazel 6): https://github.com/buildbuddy-io/rules_xcodeproj/pull/1502

### DIFF
--- a/tools/worker/worker_main.cc
+++ b/tools/worker/worker_main.cc
@@ -25,7 +25,11 @@ using bazel::tools::cpp::runfiles::Runfiles;
 
 int main(int argc, char *argv[]) {
   std::string index_import_path;
-  std::unique_ptr<Runfiles> runfiles(Runfiles::Create(argv[0]));
+  #ifdef BAZEL_CURRENT_REPOSITORY
+    std::unique_ptr<Runfiles> runfiles(Runfiles::Create(argv[0], BAZEL_CURRENT_REPOSITORY));
+  #else
+    std::unique_ptr<Runfiles> runfiles(Runfiles::Create(argv[0]));
+  #endif  // BAZEL_CURRENT_REPOSITORY
   if (runfiles != nullptr) {
     // We silently ignore errors here, we will report an error later if this
     // path is accessed


tools/worker/worker_main.cc
```diff
@@ -25,7 +25,7 @@ using bazel::tools::cpp::runfiles::Runfiles;
 
 int main(int argc, char *argv[]) {
   std::string index_import_path;
-  std::unique_ptr<Runfiles> runfiles(Runfiles::Create(argv[0]));
```
brentleyjones, line 28: We should use `#ifdef BAZEL_CURRENT_REPOSITORY` to gate this to Bazel 6. In Bazel 5 we can call it the old way.

brentleyjones (changes requested):

tools/worker/worker_main.cc
BalestraPatrick: Ah forgot `rules_apple` is the only Bazel 6, of course...

BalestraPatrick:
Mmh, failure looks related to `last-green-bazel` using Bazel at 45ca62c042f3679e60664b9ebe6d24239b390e1d which doesn't have this commit: https://github.com/bazelbuild/bazel/commit/d3a92e8ebd65a143baeef15ef5bf15d9b02716b4

tools/worker/worker_main.cc
keith: https://github.com/bazelbuild/rules_swift/pull/965

PR #976, opened 2023-01-11
Remove checked in lib for tests

keith:
This wasn't ideal in general but also caused issues across multiple
archs. This requires a simple rule to extract the swiftmodule path but
it also lets us actually verify that things build.


keith:
FYI @maxwellE

brentleyjones (approved):

tinder-maxwellelliott (approved):

maxwellE (approved):

PR #978, opened 2023-01-12
Add workaround for Swift + lld bug

keith:
This is what SwiftPM does now as well.

### DIFF
--- a/swift/internal/swift_autoconfiguration.bzl
+++ b/swift/internal/swift_autoconfiguration.bzl
@@ -142,7 +142,7 @@ def _check_use_response_files(repository_ctx, swiftc_path, temp_dir):
     )
 
 def _check_supports_lld_gc_workaround(repository_ctx, swiftc_path, temp_dir):
-    """Returns True if `swiftc` supports implementation-only imports."""
+    """Returns True if lld is being used and it supports nostart-stop-gc"""
     source_file = _scratch_file(
         repository_ctx,
         temp_dir,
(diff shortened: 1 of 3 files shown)


swift/internal/swift_autoconfiguration.bzl
```diff
@@ -130,6 +131,27 @@ def _check_use_response_files(repository_ctx, swiftc_path, temp_dir):
         "@{}".format(param_file),
     )
 
+def _check_supports_lld_gc_workaround(repository_ctx, swiftc_path, temp_dir):
+    """Returns True if `swiftc` supports implementation-only imports."""
```
brentleyjones, line 135: Wrong comment?

keith: fixed

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -130,6 +131,27 @@ def _check_use_response_files(repository_ctx, swiftc_path, temp_dir):
         "@{}".format(param_file),
     )
 
+def _check_supports_lld_gc_workaround(repository_ctx, swiftc_path, temp_dir):
```
brentleyjones, line 134: How do we enable this only if someone is using lld (to match what the Swift driver is doing)?

keith: if lld is found the cc toolchain always uses it

keith: https://github.com/bazelbuild/bazel/blob/ccfeeff06cdc0266b92781213bbdf33c4288b1c5/tools/cpp/unix_cc_configure.bzl#L440

brentleyjones (approved):

PR #979, opened 2023-01-12
Clear WORKSPACE for bzlmod test

keith:
This validates that we're not relying on anything from the WORKSPACE itself. It works even with stardoc still having issues since we avoid those targets in this job


keith:
The problem is it might produce false negatives where we break bzlmod compat I think

keith:
https://github.com/bazelbuild/rules_swift/pull/1003

PR #985, opened 2023-01-12
Add minimum bazel version as 6.0.0 in bzlmod

keith:


brentleyjones (commented):
Is this needed, as bzlmod didn't exist before 5, and really it doesn't work before 6?

keith:
i just figure we should throw a value in here that we can bump as we move forward, since it will be a benefit as we drop new ones

BalestraPatrick:
I think it should be 6.0. A very simple `MODULE.bazel` that uses `repo_name` will fail in Bazel 5 because that attribute didn't exist in 5.0.

PR #992, opened 2023-01-30
Bump shas of non-release rules_swift deps

jerrymarino:
Update shas of rules_swift non release dependencies.

Github guarantees the release archives only unfortunately so this broke CI for a number of OSS repos that don't have other mirroring strategies. 

Per github's official comms
> The default compression for Git archives has recently changed.
> As result, archives downloaded from GitHub may have different checksums
> even though the contents are completely unchanged.

https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/
https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1409390654

### DIFF
--- a/swift/repositories.bzl
+++ b/swift/repositories.bzl
@@ -88,7 +88,7 @@ def swift_rules_dependencies(include_bzlmod_ready_dependencies = True):
         http_archive,
         name = "com_github_apple_swift_protobuf",
         urls = ["https://github.com/apple/swift-protobuf/archive/1.19.0.tar.gz"],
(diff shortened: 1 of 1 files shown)


brentleyjones:
Is the `[Hold]` still intended? I'm fine with merging for now, and reverting once GitHub does their revert.

jerrymarino:
Yeah SGTM atleast it gets us unblocked for now 🚀

brentleyjones:
Revert is incoming, so maybe I won't merge: https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1409438954

PR #995, opened 2023-01-31
Build with incompatible_disallow_empty_glob

limdor:
In order to flip the flag, all downstream projects should be adapted. However, it is hard to fix them all if there are constant regressions. Adding it to the CI will ensure that once the project can build with incompatible_disallow_empty_glob it can keep building like that.
See: https://github.com/bazelbuild/bazel/pull/15327

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,6 +1,9 @@
 # https://github.com/bazelbuild/stardoc/issues/112
 common --incompatible_allow_tags_propagation
 
+# Fail if a glob doesn't match anything (https://github.com/bazelbuild/bazel/issues/8195)
+build --incompatible_disallow_empty_glob
+
 # Since there's no way to set the deployment version for swift_{binary,test},
 # this forces all targets' minimum macOS to Catalina until Bazel CI has
 # upgraded their Mac machines to Big Sur.


brentleyjones:
The errors will need to be fixed before we can merge this, but :+1: to the idea.

limdor:
@brentleyjones I just fixed the errors

third_party/com_github_apple_swift_nio/BUILD.overlay
```diff
@@ -94,7 +94,7 @@ cc_library(
     srcs = glob([
         "Sources/CNIOAtomics/**/*.c",
         "Sources/CNIOAtomics/**/*.cc",
-    ]),
+    ], allow_empty=True),
```
brentleyjones, line 97: So did these actually return empty globs before? If so, that seems like an actual bug. Otherwise, these should be `allow_empty = False`.

@thii @BalestraPatrick

limdor: Yes, here I'm just being explicit on the Bazel behavior until now.

thii: The error was about the `*.cc` one that didn't match anything—there were no C++ files returned from the list. I think we should remove that from the globs instead.

```
Error in glob: glob pattern 'Sources/CNIOAtomics/**/*.cc' didn't match anything, but allow_empty is set to False
```

jszumski: https://github.com/bazelbuild/rules_swift/pull/996

brentleyjones (approved):

Issue #997, opened 2023-02-05
Feature Request: Implement `--test_filter` Support


cwybro:
### Description
Implement Bazel [`--test_filter`](https://bazel.build/docs/user-manual#test-filter) feature, [currently marked with a `TODO`](https://github.com/bazelbuild/rules_swift/blob/a4f17129f052062bc3084d56b9f25c8b51b855f9/tools/xctest_runner/xctest_runner.sh.template#L57).

This will allow `swift_test` consumers to limit the test scope of a target to a specific test class / method, which can be very useful for particularly large targets invoked commonly the command-line (i.e. without any Xcode conveniences).

As mentioned below, I attempted this edit myself and have it working in a consumer repository. I would be more than happy to work with maintainers to contribute this back to `bazelbuild/rules_swift`, and also add tests.

In the meantime, I'm starting with an issue, per the [contribution guide](https://github.com/bazelbuild/rules_swift/blob/master/CONTRIBUTING.md#file-or-claim-an-issue).

### Additional Context

- I've implemented `--test_filter` support in a [fork of bazelbuild/rules_swift](https://github.com/cwybro/rules_swift/commit/48ebc62c7cd83d3ddc1a1f0b79c4f245175eda15).
- I've [created a repository](https://github.com/cwybro/bazel-swift-test-filter) demonstrating the consumption of this new support (and simultaneously the missing support depending on the backing of the `build_bazel_rules_swift` `http_archive`).

keith:
Looks good to me! Happy to merge a related PR!

cwybro:
> Looks good to me! Happy to merge a related PR!

Thanks, @keith! Just opened https://github.com/bazelbuild/rules_swift/pull/999 -- I'll work on adding a test for this prior to upgrading from a draft.

keith:
great! looks good so far

keith:
https://github.com/bazelbuild/rules_swift/pull/999

PR #999, opened 2023-02-12
Implement --test_filter support

cwybro:
### Description

Implement `--test_filter` support, as detailed in https://github.com/bazelbuild/rules_swift/issues/997. This can be engaged via a `swift_test` target like so:

```
bazel test //:Tests --test_filter=TestModuleName.TestClassName/testMethodName
```

**Note:** If `--test_filter` is not passed into the test invocation, we fallback to the original behavior (pass `All` into the xctest invocation.

### Changelog

- IMPROVED:
    - Implement `test_filter`, allowing users to inject a specific test scope into `swift_test` execution

### TODO

- [x] Add tests
- [x] Add docs (I'm open to updating / moving the new documentation as necessary)

### DIFF
--- a/doc/rules.md
+++ b/doc/rules.md
@@ -499,6 +499,12 @@ have the paths made absolute via swizzling by enabling the
 set the `BUILD_WORKSPACE_DIRECTORY` environment variable in your scheme to the
 root of your workspace (i.e. `$(SRCROOT)`).
 
+A subset of tests for a given target can be executed via the `--test_filter` parameter:
+
+```
+bazel test //:Tests --test_filter=TestModuleName.TestClassName/testMethodName
+```
+
 
 **ATTRIBUTES**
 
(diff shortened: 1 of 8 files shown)


cwybro:
@keith The current test suite validates the following:
1. Tests fail without test filtering
2. Test scope is reduced via test filtering

However 2 can currently be subverted if zero tests are executed (e.g. you specify a filter that matches no tests). Given this, I'd like to validate that the tests are *actually* executed, for example:
- The `test.log` contains: `Test Suite 'All tests'` / `Test Suite 'Selected tests'`
- The `test.log` contains: `Executed 1 test, with 0 failures`

Is it possible to write tests against the test log output (or another test artifact)? There might be a better way to assert against this output, or perhaps this is not a desirable test. Open to suggestions / feedback either way. Thanks!

keith:
blah yea, i think you might be able to with a shell test, but that is worse in other ways, so i think this is probably good enough for now

keith (approved):

keith:
Thanks!

PR #1001, opened 2023-02-13
Fix `swift.use_global_index_store` when using `swift.file_prefix_map`

brentleyjones:


brentleyjones:
Depends on an `index-import` release with https://github.com/MobileNativeFoundation/index-import/pull/78 in it.

keith (approved):
seems good once that release is up

PR #1002, opened 2023-02-14
Flip swift.coverage_prefix_map default

keith:
Using this flag makes objects, archives, and binaries built with
coverage data entirely hermetic. The only downside is that the coverage
reports generated from these binaries cannot be viewed in Xcode since it
doesn't support handling relative paths. If users want the previous
behavior they can specify `--features=-swift.coverage_prefix_map`


keith:
seems like this is the right default, and what we just did for objc https://github.com/bazelbuild/apple_support/pull/190

brentleyjones (approved):

PR #1014, opened 2023-02-17
Enable swift.file_prefix_map by default if supported

keith:
We already enable debug and coverage prefix maps, this one supersedes
those and potentially handles more cases in the future.


swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -635,6 +635,7 @@ def _xcode_swift_toolchain_impl(ctx):
 
     # Xcode 14 implies Swift 5.7.
     if _is_xcode_at_least_version(xcode_config, "14.0"):
+        requested_features.append(SWIFT_FEATURE_FILE_PREFIX_MAP)
```
brentleyjones, line 638: Do we also want to remove `SWIFT_FEATURE_DEBUG_PREFIX_MAP`, or are we confident that they play nice together?

keith: I am confident they play fine together since we've been running both internally. I think you're right that we could remove that but it's a tad risky

brentleyjones (approved):

PR #1018, opened 2023-03-02
Add support for removing -ObjC flag

keith:
-ObjC is used when linking to automatically force the linker to
over-discover things that appear unused to find things like Objective-C
categories and Swift protocol extensions that actually do affect
behavior. In the case that you've audited libraries for these, avoiding
passing these flags can reduce binary size.


keith:
I think this feature will go away at some point in the future, but the planning of how to remove -ObjC in general has just begun, so this is a nice stopgap since folks have asked to try this

brentleyjones (approved):

keith:
https://github.com/bazelbuild/apple_support/pull/210
https://github.com/bazelbuild/rules_swift/pull/1019

brentleyjones (approved):

PR #1019, opened 2023-03-02
Remove -ObjC link flag by default

keith:
This is moving into the cc toolchain instead so it can be more easily
disabled. If users are using the toolchain built in to bazel it's possible
their cc_binary / swift_binary / swift_test targets need to add this flag
manually until they move to the new toolchain.

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -230,7 +230,6 @@ def _swift_linkopts_providers(
         # of "objc-executable" because the latter requires additional
         # variables not provided by cc_common. Figure out how to handle this
         # correctly.
-        "-ObjC",
         "-Wl,-objc_abi_version,2",
     ]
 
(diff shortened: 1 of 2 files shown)


brentleyjones:
> If users are using the toolchain built in to bazel it's possible
their cc_binary / swift_binary / swift_test targets need to add this flag
manually until they move to the new toolchain.

I'm not sure we should merge this yet then.

keith:
yea i mean this is the potential problem we get with how the crosstool + bazel versions interact. if we don't merge this users can't actually disable -ObjC if their dependency tree has swift in it, unless we implement it another way (like the other PR I closed). my thinking in this case is that the risk is pretty low since it's only non-apple binary types that are affected

brentleyjones:
> since it's only non-apple binary types that are affected

I know there are people using swift_binary/swift_test for tools.

I support us doing this, it just seems a little soon given that many people might not even know about the new toolchain, let alone are using it.

keith:
yea fair enough. i was thinking the risk is so low that this simplification would be fine, but we can go back to https://github.com/bazelbuild/rules_swift/pull/1018

PR #1020, opened 2023-03-08
Revert "Add support for imported_library when linking (#917)"

keith:
With recent rules_apple updates, libraries are added both to CcInfo and
ObjcProvider's imported_library, which results in duplicate symbols.
Removing this and only relying on CcInfo for this is the path moving
forward. Theoretically a custom rule that produced objc info with
imported_librarys would be broken by this, but they will have to migrate
either way as this provider goes away for linking.

This reverts commit b83db34c861c1dec08c2907afd2bb9850ac88d62.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -395,10 +395,6 @@ def register_link_binary_action(
                 "-framework",
                 objc.static_framework_names.to_list(),
             ))
-            dep_link_flags.extend([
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

brentleyjones:
This breaks Bazel 5, since those users can't upgrade to a newer version of rules_apple.

PR #1024, opened 2023-03-14
Add a private feature for force alwayslink=True.

keith:
This can then be used by a toolchain that needs the behavior to ensure
all Swift code makes it into the final binary.

PiperOrigin-RevId: 516553733
(cherry picked from commit b48dda98f02cb6ea63990560eb3c21d55af0e4b0)


keith:
im lightly testing this internally as during this transition with this solution it would be impossible to set `alwayslink = False`

brentleyjones (approved):

thii (approved):

keith:
this is still disabled for now, so it doesn't do anything yet

PR #1025, opened 2023-03-21
Add proper support for custom Swift toolchains

keith:
Previously we did some amount of handling if you passed a custom
toolchain identifier. This didn't handle the case that the toolchain
required new compatibility libraries to actually link your binary. To
handle this we have to pass a new -L to the path in the root of the
toolchain. We do this with a custom env var that we replace, similar to
the DEVELOPER_DIR variable. Ideally this would be cached in memory, but
we can't do that without editing bazel, luckily xcrun has a cache that
makes this likely fast enough (about 0.01s for a cached invocation).

### DIFF
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -195,7 +195,8 @@ def _sdk_developer_framework_dir(apple_toolchain, target_triple, xcode_config):
 def _swift_linkopts_providers(
         apple_toolchain,
         target_triple,
-        toolchain_label):
+        toolchain_label,
+        toolchain_root):
     """Returns providers containing flags that should be passed to the linker.
 
     The providers returned by this function will be used as implicit
@@ -207,6 +208,8 @@ def _swift_linkopts_providers(
         target_triple: The target triple `struct`.
         toolchain_label: The label of the Swift toolchain that will act as the
             owner of the linker input propagating the flags.
+        toolchain_root: The path to a custom Swift toolchain that could contain
+            libraries required to link the binary
 
     Returns:
         A `struct` containing the following fields:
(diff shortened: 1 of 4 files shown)


keith:
https://github.com/bazelbuild/apple_support/pull/213

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -591,10 +597,20 @@ def _xcode_swift_toolchain_impl(ctx):
     # script, use `--define=SWIFT_CUSTOM_TOOLCHAIN=<id>` as shown below.
     swift_executable = get_swift_executable_for_toolchain(ctx)
     toolchain_root = ctx.var.get("SWIFT_USE_TOOLCHAIN_ROOT")
-    custom_toolchain = ctx.var.get("SWIFT_CUSTOM_TOOLCHAIN")
+    custom_toolchain = ctx.var.get("SWIFT_CUSTOM_TOOLCHAIN") or ctx.configuration.default_shell_env.get("TOOLCHAINS")
```
brentleyjones, line 600: Should we add a comment to remove the `SWIFT_CUSTOM_TOOLCHAIN` part eventually? Maybe when we do a major version bump?

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -591,10 +597,20 @@ def _xcode_swift_toolchain_impl(ctx):
...
+    custom_xcode_toolchain_root = None
     if toolchain_root and custom_toolchain:
         fail("Do not use SWIFT_USE_TOOLCHAIN_ROOT and SWIFT_CUSTOM_TOOLCHAIN" +
```
brentleyjones, line 603: This should be adjusted.

brentleyjones (approved):

swift/internal/xcode_swift_toolchain.bzl
keith: this is removed in https://github.com/bazelbuild/rules_swift/pull/1026

swift/internal/xcode_swift_toolchain.bzl
keith: yea im not sure, since it didn't really work i don't think it's very relied on. i could keep it for now, or just error on it if it's passed at all?

brentleyjones: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/690ef6814670ac1e51a59988585d4bfe1d31f3e0/xcodeproj/internal/bazel_integration_files/bazel_build.sh#L121

🙃

keith: yea but in the common case where your deployment target is <iOS "current" that won't be enough because of the missing linker options

keith: added a comment but we can probably remove sooner in a separate PR IMO

brentleyjones: Just saying that erroring out would force people to upgrade rules_xcodeproj. Better to keep both for now and determine how long we want to support not breaking people (in particular rules_xcodeproj).

brentleyjones (approved):

swift/internal/xcode_swift_toolchain.bzl
keith: yep that's just fine, but im saying it must be that no one is using it because it doesn't work today either.

brentleyjones: It worked for me when I tested it, but I haven't tested it recently.

keith: yea you have to get lucky with minimum deployment target, and what changes have happened recently in swift wrt compatibility changes

PR #1026, opened 2023-03-21
Remove old custom Swift build options

keith:
The toolchain root option has been superseded by passing a custom
swift_executable, the SWIFT_CUSTOM_TOOLCHAIN option is being replaced by
using the native TOOLCHAINS xcrun option.


keith:
this is just cleaning stuff up before https://github.com/bazelbuild/rules_swift/pull/1025

PR #1042, opened 2023-04-13
Mark swift_autoconfiguration as local

### LINKED ISSUE #1041: Caching issue when updating Swift
I'm trying to update Swift in our CI setup, and I realized that the intermediate artifacts cache is not invalidated when using another swift version.

Here's an example error message I get:
`external/swiftpkg_swift-parsing/Sources/Parsing/Conversions/Enum.swift:1:8: error: module compiled with Swift 5.7.3 cannot be imported by the Swift 5.8 compiler: /home/builder/.cache/bazel/a55e444fc86752cc6129d960be3bceff/sandbox/linux-sandbox/2/execroot/prx/bazel-out/k8-opt/bin/external/swiftpkg_swift-case-paths/CasePaths.swiftmodule`

The strange thing is that it looks like this was designed to work fine, as there's a `_write_swift_version` function in `swift_autoconfiguration.bzl` but somehow, it's not used by Bazel to invalidate its cache keys.

I'm really not sure where to look to make sure this works correctly, and any pointer is appreciated.

gferon:
This makes sure the rule is evaluated whenever a new Bazel server starts, which gives it a chance to regenerate the `swift-version` file which will be used in the resulting action keys.

One example to reproduce the issue without `local = True`:

```sh
ln -sf /opt/swift-5.7/bin/swift* /usr/local/bin/
bazel build //:swift-binary
cat $(bazel info output_base)/external/build_bazel_rules_swift_local_config/swift_version
# Swift version 5.7.3 (swift-5.7.3-RELEASE)
# Target: x86_64-unknown-linux-gnu

bazel shutdown
ln -sf /opt/swift-5.8/bin/swift* /usr/local/bin/

bazel build //:swift-binary # cache hit (should be a miss)
cat $(bazel info output_base)/external/build_bazel_rules_swift_local_config/swift_version
# Swift version 5.7.3 (swift-5.7.3-RELEASE)
# Target: x86_64-unknown-linux-gnu
```

with `local = True` the example above works as expected.

Fixes #1041

### DIFF
--- a/swift/internal/swift_autoconfiguration.bzl
+++ b/swift/internal/swift_autoconfiguration.bzl
@@ -413,4 +413,5 @@ def _swift_autoconfiguration_impl(repository_ctx):
 swift_autoconfiguration = repository_rule(
     environ = ["CC", "PATH", "ProgramData", "Path"],
     implementation = _swift_autoconfiguration_impl,
+    local = True,
 )


keith (approved):

keith:
nice find, thanks!

PR #1043, opened 2023-05-04
Remove Bitcode support

keith:
Now that Xcode 14.1 is the minimum supported to the App Store, and Xcode
14.0 dropped support for bitcode, we don't need to handle these flags
anymore.

This keeps support for emitting LLVM Bitcode (confusing naming) for use
with LTO. This changes the relevant tests to test that behavior instead.

(cherry picked from commit 797660a758f924d9121689ba666d3f92044c2e5a)


keith:
https://github.com/bazelbuild/bazel/issues/18321

mattrobmattrob (approved):

PR #1046, opened 2023-05-16
Add support for -package-name

keith:


keith:
i don't think there's a need to gate this on swift version since if you actually use this on a target, and then use the package ACL, that would fail anywayas

swift/internal/attrs.bzl
```diff
@@ -129,6 +129,12 @@ The name of the Swift module being built.
 If left unspecified, the module name will be computed based on the target's
 build label, by stripping the leading `//` and replacing `/`, `:`, and other
 non-identifier characters with underscores.
+""",
+            ),
+            "package_name": attr.string(
+                doc = """\
+The semantic package of the Swift target being built. Targets with the same
```
brentleyjones, line 136: We should mention that it required Swift 5.9+

brentleyjones (approved):

keith:
sg

keith:
FWIW  this does work now, we can still wait until a bit closer to merge tho

brentleyjones:
Is this in an Xcode beta? If so, I think we can merge now.

keith:
yea it is, sg

Issue #1055, opened 2023-05-23
modulemaps and umbrella headers from private objc deps are getting propagated to rdeps


ivan-golub:
When objc lib is added as private deps to swift library its modulemap and umbrella header doesn't need to be propagated in apple_common.Objc as they are not part of linking and expose extra info that than can cause sandboxed build errors as modulemap file itself is not an explicit input. 

Example of sandbox error we experiences due to transitive swift/objc private deps
```
bazel clean; bazel build //Features/Foo:Foo --disk_cache="" --remote_cache="" --spawn_strategy=sandboxed --sandbox_debug 
INFO: Invocation ID: 1b86e27a-f366-4627-921c-1ec9f403e02f
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.

INFO: Analyzed target //Features/Foo:Foo (74 packages loaded, 1012 targets configured).
INFO: Found 1 target...
ERROR: /Users/ivan.golub/Dev/app/Features/Foo/BUILD.bazel:3:8: Compiling Features/Foo/Sources/SCFooUtils.m failed: (Exit 1): sandbox-exec failed: error executing command 
  (cd /Users/ivan.golub/Dev/.cache/bazel/arm64/e7e39dff3153e0ed28e9964aecdc85fa/sandbox/darwin-sandbox/68/execroot/app && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=16.1 \
    DEVELOPER_DIR=/Applications/Xcode14.1_14B47b.app/Contents/Developer \
    PATH=/bin:/usr/bin:/usr/local/bin:/opt/homebrew/bin \
    SDKROOT=/Applications/Xcode14.1_14B47b.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator16.1.sdk \
    TMPDIR=/var/folders/bl/t9zx_pl53qxg6t3t4bkzxyr80000gn/T/ \
    XCODE_VERSION_OVERRIDE=14.1.0.14B47b \
    ZERO_AR_DATE=1 \
  /usr/bin/sandbox-exec -f /Users/ivan.golub/Dev/.cache/bazel/arm64/e7e39dff3153e0ed28e9964aecdc85fa/sandbox/darwin-sandbox/68/sandbox.sb ........
[...]

PR #1056, opened 2023-05-23
Do not propagate module_map and umbrella_headers of objc private deps

ivan-golub:
Address https://github.com/bazelbuild/rules_swift/issues/1055


brentleyjones:
Can you add tests to verify this behavior so we don't regress it?

PR #1069, opened 2023-07-05
Fix stderr with workers

keith:
std::cerr is ignored when this code is being run with workers.

### DIFF
--- a/tools/common/process.cc
+++ b/tools/common/process.cc
@@ -368,8 +368,8 @@ int RunSubProcess(const std::vector<std::string> &args,
     } while ((wait_status == -1) && (errno == EINTR));
 
     if (wait_status < 0) {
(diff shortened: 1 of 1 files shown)


tools/common/process.cc
```diff
@@ -368,8 +368,8 @@ int RunSubProcess(const std::vector<std::string> &args,
     } while ((wait_status == -1) && (errno == EINTR));
 
     if (wait_status < 0) {
-      std::cerr << "Error waiting on child process '" << args[0] << "'. "
-                << strerror(errno) << "\n";
+      (*stderr_stream) << "error: 1 waiting on child process '" << args[0]
```
brentleyjones, line 371: What's with the "1"?

keith: oops thanks, debugging code

brentleyjones (approved):

PR #1070, opened 2023-07-05
Add test validating global index store builds

keith:
Previously only arguments were validated, now we validate that the
actions actually succeed, which makes sure we successfully run
index-import


keith:
this should fail until https://github.com/MobileNativeFoundation/index-import/pull/92 is updated here

brentleyjones (approved):

keith:
fails until we update to https://github.com/bazelbuild/apple_support/pull/225

PR #1078, opened 2023-07-13
Add target name to output file path

tymurmustafaiev:
To be able to build multiple flavours of same target, we needed to add `target_name` to output paths as bazel were conflicting with same output path names.
Added feature flag `SWIFT_FEATURE_ADD_TARGET_NAME_TO_OUTPUT`, which enables adding `target_name` to output.


swift/internal/derived_files.bzl
```diff
@@ -222,7 +275,11 @@ def _swiftc_output_file_map(actions, target_name):
     Returns:
         The declared `File`.
     """
-    return actions.declare_file("{}.output_file_map.json".format(target_name))
+    return _declare_file(
+        actions,
+        target_name,
+        "{}.output_file_map.json".format(target_name),
```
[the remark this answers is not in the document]
acecilia: Agree. I would say better keep the subfolder (aligned with all others) and change the name:
```
    return _declare_file(
        actions,
        target_name,
        "output_file_map.json",
    )
```

swift/internal/compiling.bzl
```diff
@@ -2879,7 +2885,7 @@ def _declare_validated_generated_header(actions, generated_header_name):
             "extension (got '{}').".format(generated_header_name),
         )
 
-    return actions.declare_file(generated_header_name)
+    return actions.declare_file(paths.join(target_name, generated_header_name))
```
acecilia, line 2888: Would it make sense to move this line to the file `derived_files.bzl`?
Context: this `declare_file` is the only one in the repository that is not inside `derived_files.bzl`

tymurmustafaiev: Thanks, done

swift/internal/compiling.bzl
```diff
@@ -2764,6 +2798,7 @@ def _declare_multiple_outputs_and_write_output_file_map(
         srcs: The list of source files that will be compiled.
         target_name: The name (excluding package path) of the target being
             built.
+        add_target_name_to_output_path: Add target_name in output path. More info at SWIFT_FEATURE_ADD_TARGET_NAME_TO_OUTPUT description.
```
acecilia, line 2801: [detail] I think is `More info in` instead of `More info at`

swift/internal/feature_names.bzl
```diff
@@ -334,3 +334,8 @@ SWIFT_FEATURE_OBJC_LINK_FLAGS = "swift.objc_link_flag"
 # A private feature that is set by the toolchain if the given toolchain wants
 # all Swift compilations to always be linked.
 SWIFT_FEATURE__FORCE_ALWAYSLINK_TRUE = "swift._force_alwayslink_true"
+
+# A feature that enables adding target_name in output path to support
+# build of multiple flavours of same framework which may have different
+# target name, but same module name.
```
acecilia, line 340: I would propose not mention `flavours`, as is a term that is not widely understood.

```
# A feature that adds target_name in output path to support building
# multiple frameworks with different target name, but same module name.
```

test/private_deps_tests.bzl
```diff
@@ -27,6 +27,16 @@ private_deps_provider_test = make_provider_test_rule(
     },
 )
 
+# Test with enabled `swift.add_target_name_to_output` feature
```
acecilia, line 30: [detail] This comment helps, may be good to add it in every related `make_provider_test_rule`

swift/internal/derived_files.bzl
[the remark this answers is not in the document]
tymurmustafaiev: I prefer keep names as it is to avoid changes in `.cc` files 🙏

PR #1089, opened 2023-08-15
Fix compile_module_interface with explicit module map

keith:
This failed to compile previously because we passed -Xfrontend to
something that was already a frontend invocation

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1718,16 +1718,13 @@ def _dependencies_swiftmodules_vfsoverlay_configurator(prerequisites, args, is_f
 
     # Bug: `swiftc` doesn't pass its `-vfsoverlay` arg to the frontend.
     # Workaround: Pass `-vfsoverlay` directly via `-Xfrontend`.
-    if is_frontend:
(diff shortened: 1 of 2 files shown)


swift/internal/compiling.bzl
```diff
@@ -1688,31 +1712,42 @@ def _dependencies_swiftmodules_configurator(prerequisites, args):
...
-        "-Xfrontend",
-        "-vfsoverlay{}".format(prerequisites.vfsoverlay_file.path),
-    )
+    if is_frontend:
+        args.add(
+            "-vfsoverlay{}".format(prerequisites.vfsoverlay_file.path),
+        )
+    else:
+        args.add(
+            "-Xfrontend",
```
brentleyjones, line 1727: Could we instead do a single `add` `if not is_frontend:` to prevent duplication? Same for the other function.

keith: ah yea that would work for this one but in the other one it might not be as nice since we'd have to do 2 conditions for it since we'd be conditionalizing `before_each`

brentleyjones (approved):

PR #1090, opened 2023-08-16
Move check for LLVM toolchain outside of repository_rule

### LINKED ISSUE #991: Flag --incompatible_enable_cc_toolchain_resolution will break rules_swift in Bazel 7.0
Incompatible flag `--incompatible_enable_cc_toolchain_resolution` will be enabled by default in the next major release (Bazel 7.0), thus breaking rules_swift. Please migrate to fix this and unblock the flip of this flag.

The flag is documented here: https://github.com/bazelbuild/bazel/issues/7260.

Please check the following CI builds for build and test results:

- [macOS, OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1411#01860033-ed64-42fd-9c8c-1449d770c654)

Never heard of incompatible flags before? We have [documentation](https://docs.bazel.build/versions/master/backward-compatibility.html) that explains everything.
If you have any questions, please file an issue in https://github.com/bazelbuild/continuous-integration.

gferon:
Checking `CC=clang` in the implementation of the `repository_rule` doesn't work when using `--incompatible_enable_cc_toolchain_resolution` (for example when using https://github.com/grailbio/bazel-toolchain to bring a hermetic LLVM toolchain).

Moving the check for LLVM the `repository_rule` to the Swift toolchain impl used by all `rule`s instead solves this issue.

I tested that the check (the failure message) is triggered on Linux with:
- only `gcc` installed
- a LLVM toolchain configured in `WORKSPACE` but without `--incompatible_enable_cc_toolchain_resolution`

TODO (maybe?):
- [ ] Add a CI job, or modify a CI job to test the case with a LLVM toolchain?

Fixes #991

### DIFF
--- a/swift/internal/swift_autoconfiguration.bzl
+++ b/swift/internal/swift_autoconfiguration.bzl
@@ -237,11 +237,11 @@ def _create_linux_toolchain(repository_ctx):
     Args:
       repository_ctx: The repository rule context.
     """
-    cc = repository_ctx.os.environ.get("CC") or ""
-    if "clang" not in cc:
-        fail("ERROR: rules_swift uses Bazel's CROSSTOOL to link, but Swift " +
-             "requires that the driver used is clang. Please set `CC=clang` " +
-             "in your environment before invoking Bazel.")
+    # cc = repository_ctx.os.environ.get("C_COMPILER") or ""
+    # if "clang" not in cc:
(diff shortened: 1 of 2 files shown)


keith (approved):
nice thanks! note that with swift it's possible the hermetic llvm toolchain may not work since swift's llvm fork could differ in required ways, but this still seems like a nice fix

PR #1099, opened 2023-09-05
Remove grep_includes

keith:
This was removed from bazel in https://github.com/bazelbuild/bazel//commit/ca8b39895bda7a48d7765b97ff9491794cb34edc


keith:
we'll see if this works for both HEAD and 6.x

keith:
it did fail on CI before https://buildkite.com/bazel/rules-swift-swift/builds/4499#018a62ba-0886-4efa-9d2d-05cd5bb3cee7 so if this is entirely green i think we're good

mattrobmattrob (approved):

PR #1102, opened 2023-09-07
Add platform_mappings

comius:
This makes it possible to use C++ toolchains with rules_swift.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -56,8 +56,10 @@ tasks:
     <<: *mac_common
     build_flags:
       - "--enable_bzlmod"
+      - "--platform_mappings=platform_mappings.empty_for_bzlmod" # TODO(bazelbuild/bazel#17127): remove when resolved
     test_flags:
       - "--enable_bzlmod"
+      - "--platform_mappings=platform_mappings.empty_for_bzlmod"
 
   macos_last_green:
     name: "Last Green Bazel"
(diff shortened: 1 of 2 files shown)


comius:
Blocked on: https://github.com/bazelbuild/bazel/issues/17127

keith:
in the meantime we can probably either duplicate everything in the platform_mappings file to use the bzlmod identifier, or specify a different `--platforms_mappings=` with our bzlmod config on CI

comius:
> in the meantime we can probably either duplicate everything in the platform_mappings file to use the bzlmod identifier

I tried using @apple_support in `platform_mapping` and even removing `repo_name` from MODULE.bazel and it just doesn't work

> or specify a different --platforms_mappings= with our bzlmod config on CI

that I can do

comius:
Done

keith:
erm won't we need it to have content to support the other changes you're making in this job?

comius:
> erm won't we need it to have content to support the other changes you're making in this job?

Sorry I’m not following. I’m just adding plaform mappings from rules_apple and disabling temporary for bzlmod. What other changes?

keith:
i mean are we going to have to un-disable for bzlmod very soon after you land https://github.com/bazelbuild/bazel/pull/19441?

comius:
> i mean are we going to have to un-disable for bzlmod very soon after you land [bazelbuild/bazel#19441](https://github.com/bazelbuild/bazel/pull/19441)?

Correct. Yesterday I didn’t know it will be fixed so quickly. You can either merge this and I prepare a second pr, or I revert this one back to the old form. Whatever you prefer.

PR #1104, opened 2023-09-07
Add platform mappings

keith:
Replace https://github.com/bazelbuild/rules_swift/pull/1102 with support for bzlmod

### DIFF
--- a/platform_mappings_bzlmod
+++ b/platform_mappings_bzlmod
@@ -1,120 +1,120 @@
 platforms:
-  @@apple_support~1.9.0//platforms:macos_x86_64
+  @apple_support~1.9.0//platforms:macos_x86_64
     --cpu=darwin_x86_64
 
-  @@apple_support~1.9.0//platforms:macos_arm64
+  @apple_support~1.9.0//platforms:macos_arm64
     --cpu=darwin_arm64
 
-  @@apple_support~1.9.0//platforms:darwin_arm64e
+  @apple_support~1.9.0//platforms:darwin_arm64e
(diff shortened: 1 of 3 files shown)


platform_mappings_bzlmod
```diff
@@ -0,0 +1,120 @@
+platforms:
+  @apple_support~1.9.0//platforms:macos_x86_64
+    --cpu=darwin_x86_64
+
+  @apple_support~1.9.0//platforms:macos_arm64
```
brentleyjones, line 5:
```suggestion
  @@apple_support~1.9.0//platforms:macos_arm64
```

keith: hrm is that required / better? I don't have that in our internal one (although we're tying this to `~override`, not sure if that matters)

keith: hrm is that required / better? I don't have that in our internal one (although we're tying this to `~override`, not sure if that matters)

brentleyjones: Needed. See the CI failure. This is a canonical label, so it needs two `@`.

keith: wasn't the issue, any other ideas? im surprised it doesn't repro locally

keith: nvm, version issue, just needed rebase

brentleyjones: Ahh. I wouldn't revert your change though. I think it should have the extra `@`, and it only works because of a fallback in the repo mapping that will translate it to canonical if it can't find a normal mapping.

PR #1108, opened 2023-09-13
Add macros support for Linux

keith:
This uses Swift's builtin feature detection to check what flags are
available. This probably means previous toolchains where macros weren't
complete would return true for this, but that's probably fine.

This also requires absolutizing TMPDIR since that is validated in a case
on Linux and errors otherwise. Hoping this doesn't cause issues.

### DIFF
--- a/tools/worker/swift_runner.cc
+++ b/tools/worker/swift_runner.cc
@@ -301,7 +301,7 @@ bool SwiftRunner::ProcessArgument(
       } else if (StripPrefix("-macro-expansion-dir=", new_arg)) {
         changed = true;
         std::filesystem::create_directories(new_arg);
-        job_env_["TMPDIR"] = new_arg;
+        job_env_["TMPDIR"] = std::filesystem::absolute(new_arg);
       } else if (new_arg == "-ephemeral-module-cache") {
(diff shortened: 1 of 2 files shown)


tools/worker/swift_runner.cc
```diff
@@ -301,7 +301,7 @@ bool SwiftRunner::ProcessArgument(
       } else if (StripPrefix("-macro-expansion-dir=", new_arg)) {
         changed = true;
         std::filesystem::create_directories(new_arg);
-        job_env_["TMPDIR"] = new_arg;
```
brentleyjones, line 304: I think this needs to be allowed to be relative, otherwise you run into caching issues.

keith: it seems like they expect it to be absolute https://github.com/apple/swift-tools-support-core/blob/33ae5abc07ea505a9865b4c11329db7a83cb3094/Sources/TSCBasic/FileSystem.swift

keith: ok updated with a hack that only applies to linux because it has a minor downside mentioned in the comment

brentleyjones (approved):

PR #1110, opened 2023-09-13
Fix fetching env on windows

keith:
Turns out there is _environ. This isn't all of this support, we also have to process the env before exec'ing the process


brentleyjones (approved):

keith:
thanks @compnerd

PR #1114, opened 2023-09-19
Enable use_default_shell_env everywhere

keith:
Now that bazel 7.x has
https://bazel.build/reference/command-line-reference#flag--incompatible_merge_fixed_and_default_shell_env
enabled by default, we can set `use_default_shell_env = True` moving
forward. This makes actions respect `--action_env=FOO` which can be
useful for things like `TOOLCHAINS`, `ZERO_AR_DATE`, `TMPDIR`, etc.
Previously these were ignored.

### DIFF
--- a/swift/internal/actions.bzl
+++ b/swift/internal/actions.bzl
@@ -19,6 +19,8 @@ load("@bazel_skylib//lib:types.bzl", "types")
 load(":features.bzl", "are_all_features_enabled")
 load(":toolchain_config.bzl", "swift_toolchain_config")
 
+USE_DEFAULT_SHELL_ENV = not hasattr(apple_common, "apple_crosstool_transition")
+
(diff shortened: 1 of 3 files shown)


swift/internal/actions.bzl
```diff
@@ -19,6 +19,8 @@ load("@bazel_skylib//lib:types.bzl", "types")
 load(":features.bzl", "are_all_features_enabled")
 load(":toolchain_config.bzl", "swift_toolchain_config")
 
+USE_DEFAULT_SHELL_ENV = not hasattr(apple_common, "apple_crosstool_transition")
```
brentleyjones, line 22: Can we add a comment above this saying it's a stand-in check for Bazel 7 which has `--incompatible_merge_fixed_and_default_shell_env` defaulted?

brentleyjones (approved):

PR #1121, opened 2023-10-06
Add SwiftGRPCInfo provider

AttilaTheFun:
Hi folks! I'm working on a tool to copy generated .pb.swift and .grpc.swift files out of the Bazel build tree into my repo. These files aren't used directly, they're just meant for reference / easy lookup / non-Xcode editor support (VSCode).

Currently, getting this information from GRPC targets is a pain, and they don't differentiate the providers based on flavor. Also in the case of the existing SwiftProtoInfo you get a depset of *all* of the transitive, generated pbswift files, not just those for direct dependencies. And the SwiftProtoInfo doesn't include the GRPC files at all.

This change is purely additive so it should not break anyone. It introduces a new SwiftGRPCInfo provider returned from the swift_grpc_library rule which exposes the flavor and direct_grpc_files generated by the target. I also added an analogous  direct_pbswift_files field to the SwiftProtoInfo provider. These are just flat lists of the generated .pb.swift and .grpc.swift files respectively for direct (proto) deps of those targets.

### DIFF
--- a/swift/internal/providers.bzl
+++ b/swift/internal/providers.bzl
@@ -62,6 +62,19 @@ the allowlist.
     },
 )
 
+SwiftGRPCInfo = provider(
+    doc = "Propagates Swift-specific information about a `swift_grpc_library`.",
+    fields = {
+        "flavor": """\
+The flavor of GRPC that was generated. E.g. server, client, or client_stubs.
+""",
+        "direct_pbgrpc_files": """\
+`Depset` of `File`s. The Swift source files (`.grpc.swift`) generated
+from the `.proto` files in direct dependencies.
+""",
+    },
+)
+
 SwiftInfo = provider(
     doc = """\
 Contains information about the compiled artifacts of a Swift module.
@@ -120,6 +133,10 @@ libraries to import those modules in generated code.
         "pbswift_files": """\
 `Depset` of `File`s. The transitive Swift source files (`.pb.swift`) generated
 from the `.proto` files.
+""",
+        "direct_pbswift_files": """\
+`list` of `File`s. The Swift source files (`.pb.swift`) generated
+from the `.proto` files in direct dependencies.
 """,
     },
 )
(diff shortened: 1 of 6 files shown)


swift/swift.bzl
```diff
@@ -92,6 +93,7 @@ load(
 # Re-export providers.
 SwiftInfo = _SwiftInfo
 SwiftProtoInfo = _SwiftProtoInfo
+SwiftGRPCInfo = _SwiftGRPCInfo
```
brentleyjones, line 96:
```suggestion
SwiftGRPCInfo = _SwiftGRPCInfo
SwiftInfo = _SwiftInfo
SwiftProtoInfo = _SwiftProtoInfo
```

swift/internal/swift_protoc_gen_aspect.bzl
```diff
@@ -265,6 +265,7 @@ def _build_swift_proto_info_provider(
             direct = pbswift_files,
             transitive = [dep[SwiftProtoInfo].pbswift_files for dep in deps],
         ),
+        direct_pbswift_files = pbswift_files,
```
brentleyjones, line 268: Can you move this to above `module_mappings` to keep it alphabetical?

swift/internal/providers.bzl
```diff
@@ -120,6 +120,23 @@ libraries to import those modules in generated code.
...
 from the `.proto` files.
+""",
+        "direct_pbswift_files": """\
+`list` of `File`s. The Swift source files (`.pb.swift`) generated
+from the `.proto` files in direct dependencies.
+""",
+    },
+)
+
+SwiftGRPCInfo = provider(
```
brentleyjones, line 131: Can you move this to be above `SwiftInfo` to keep this file alphabetical?

AttilaTheFun:
Thanks @brentleyjones ! I addressed your comments in the latest commit.

brentleyjones (approved):

PR #1132, opened 2023-11-06
Add gazelle for bzl_library generation

keith:
This makes it easier to keep these up to date. Also if we have conflicts
with upstream we can just ignore those changes and re-run this
afterwards


test/rules/BUILD
```diff
@@ -1,15 +1,8 @@
...
-    ],
-    deps = [
-        "//swift",
-        "@bazel_skylib//lib:collections",
-        "@bazel_skylib//lib:types",
-        "@bazel_skylib//lib:unittest",
-    ],
+    name = "swift_library_artifact_collector",
+    srcs = ["swift_library_artifact_collector.bzl"],
+    visibility = ["//visibility:public"],
```
brentleyjones, line 6: Are these visibility changes needed?

brentleyjones:
Nit: can the changes and the addition of the new tool be different PRs?

keith:
split it into 2 commits here so we can be sure it still works

test/rules/BUILD
keith: fixed, for reference whenever gazelle changes a target it resets the visibility unless you have `package(default_visibility`, kind of annoying that it only does it when it changes something, so it's not obvious that in the future it might overwrite yours

luispadron (approved):

PR #1133, opened 2023-11-07
Update `xctest_runner` to fail if it doesn't execute any tests

jszumski:
Ports https://github.com/bazelbuild/rules_apple/pull/1978 into `rules_swift`.

- Updates the `xctest_runner` to fail if it doesn't execute any tests
- Adds tests for basic runner usage
- Adds tests for empty bundle cases

### DIFF
--- a/test/rules/swift_shell_test.bzl
+++ b/test/rules/swift_shell_test.bzl
@@ -1,0 +1,80 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Test rule for verifying output of Swift binaries."""
+
+load("@bazel_skylib//lib:shell.bzl", "shell")
+
+def _swift_shell_test_impl(ctx):
+    """Implementation of the swift_shell_test rule."""
+
+    test_executable = ctx.attr.target_under_test.files_to_run.executable
+
+    runfiles = ctx.runfiles(
+        files = [test_executable],
+    ).merge_all([
+        ctx.attr.target_under_test.default_runfiles,
+        ctx.attr._tool.default_runfiles,
+    ])
+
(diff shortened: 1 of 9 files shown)


.bazelrc
```diff
@@ -38,3 +38,6 @@ build --cxxopt='-std=c++14'
 # There is an open issue about it on the zlib repository here:
 # https://github.com/madler/zlib/issues/633
 build --per_file_copt="external/.*zlib.*/.*.c@-Wno-deprecated-non-prototype"
+
+# Needed to load @build_bazel_apple_support//:BUILD for integration tests
+build --nocheck_visibility
```
jszumski, line 43: rules_apple seemingly has this enabled for other reasons but it happened to help the integration tests

test/bazel_testrunner.sh
jszumski, line 1: This was copied and I manually pruned the rules_apple bits

test/swift_shell_testutils.sh
jszumski, line 1: This was copied and I manually pruned the rules_apple bits

test/unittest.bash
jszumski, line 1: This was copied without modifications

tools/xctest_runner/xctest_runner.sh.template
```diff
@@ -55,12 +56,21 @@ fi
...
 
 if [[ "$exit_status" -ne 0 ]]; then
   exit "$exit_status"
 fi
 
+# Fail when bundle executes nothing
+test_target_execution_count=$(grep -e "Executed [[:digit:]]\{1,\} tests*," "$testlog" | tail -n1)
+if echo "$test_target_execution_count" | grep -q -e "Executed 0 tests, with 0 failures"; then
+  echo "error: no tests were executed, is the test bundle empty?" >&2
+  exit 1
```
jszumski, line 71: This is the runner behavior change

luispadron (approved):
I wonder if longer term the swift_shell_test infrastructure could be shared

keith:
I think instead of re-execing bazel in these tests we can make a `sh_test` where the `data` is the `swift_test` target, and exec that in the sh_test script, checking the error code. We haven't done that in the iOS case because of simulator management, but I think that might be fine for this case, in which case it would mean a lot less infra

test/rules/swift_shell_test.bzl
```diff
@@ -0,0 +1,80 @@
...
+
+    return [
+        DefaultInfo(
+            executable = output_script,
+            runfiles = runfiles,
+        ),
+        ctx.attr.target_under_test[OutputGroupInfo],
+    ]
+
+swift_shell_test = rule(
```
luispadron, line 52: nice!

luispadron (approved):

keith (approved):
nice!

PR #1136, opened 2023-11-12
Add features to derive module file names in PascalCase and omit package

AttilaTheFun:
The module names generated for targets which did not or were not able to supply one via an attribute are currently derived in "snake_case" and include the package name. The convention for Swift module names is PascalCase (see: Foundation, UIKit, SwiftUI, SwiftSyntax, Dispatch etc), and the package name is only necessary if developers don't already have a naming convention for targets which avoids collisions.

I want to add the capability to derive module names without the package name and in PascalCase. I can't just pass the module name as an attribute for some targets such as swift_proto_library because these are actually generated through an aspect.

(As an aside, I dislike this design -- I would rather just have one swift_proto_library / swift_grpc_library which generates the protos and compiles them from each proto_library. I believe this was because the swift protoc plugin is designed to compile all of the protos in a single invocation.)

To support this behavior, I added two new features:
```
# Omit the package name when deriving the module name.
# E.g. //package:some_library would normally have the module name "package_some_library"
# With this feature enabled, the module name would be "some_library".
SWIFT_FEATURE_DERIVED_MODULE_NAME_OMIT_PACKAGE = "swift.derived_module_name_omit_package"

# Convert to PascalCase when deriving the module name.
# E.g. //package:some_library would normally have the snake_case module name "package_some_library"
# With this feature enabled, the module name would be "PackageSomeLibrary".
SWIFT_FEATURE_DERIVED_MODULE_NAME_PASCAL_CASE = "swift.derived_module_name_pascal_case"
```

Which can either be applied at the target level or to all targets in .bazelrc. I updated the xplatform grpc example to leverage these features and demonstrate them in action.

Additionally, while working on this, I noticed that the derive_module_name function had been duplicated for use in the swift compiler plugin rules even though it was already available to the plugin via the swift_common import. I deleted the duplicate and unified all calls to this function.


doc/api.md
```diff
@@ -533,6 +533,47 @@ This mapping is intended to be fairly predictable, but not reversible.
...
+<pre>
+swift_common.derive_proto_module_name(<a href="#swift_common.derive_proto_module_name-label">label</a>, <a href="#swift_common.derive_proto_module_name-omit_package">omit_package</a>, <a href="#swift_common.derive_proto_module_name-pascal_case">pascal_case</a>)
+</pre>
+
+Returns a derived proto module name from the given build label.
+
+For swift_proto_library and swift_grpc_library targets,
+the module name is derived from the corresponding proto library target.
+
+The name computed using the following algorithm:
```
luispadron:
```suggestion
The name is computed using the following algorithm:
```

swift/internal/compiling.bzl
```diff
@@ -2022,6 +2022,52 @@ def derive_module_name(*args):
...
+        All _interior_ sequences of non-identifier characters (anything other
+        than `a-z`, `A-Z`, `0-9`, and `_`) are replaced by a single underscore
+        (`_`). Any leading or trailing non-identifier characters are dropped.
+    *   If the package component is non-empty after the above transformation,
+        it is joined with the transformed name component using an underscore.
+        Otherwise, the transformed name is used by itself.
+    *   If this would result in a string that begins with a digit (`0-9`), an
+        underscore is prepended to make it identifier-safe.
+
+    This mapping is intended to be fairly predictable, but not reversible.
```
luispadron, line 2042: I'd suggest just referencing the algo from the feature flag doc, might be the case these get out of sync

luispadron (commented):
Some comments on changes. Over all I think this makes sense and is helpful for making module names match their Swift names.

I do wonder though the use of splitting this into two flags, did we consider just one feature flag that "swiftifies" these module names vs. the two?

swift/internal/compiling.bzl
AttilaTheFun: Yeah that's a good callout, I can add that.

AttilaTheFun:
Thanks @luispadron ! Re: one feature flag or two, I think it's still worthwhile to separate the behavior. For example, someone might want the PascalCase names but their proto packages are not structured in such a way that module names would be unique if the package was omitted. 

What would you think of expanding the scope of this change to include *all* derived module names and for the feature flags to control the PascalCase and package-including behavior for any module name that is not set explicitly. For example, when using rules_swift_package_manager, I added a directive to explicitly convert module names to PascalCase, but if this could just be configured at the rules_swift level that could be interesting.

AttilaTheFun:
Also @luispadron I think because I accepted your commit suggestions, it's complaining because you haven't accepted the Google CLA. Could you accept that here?
https://github.com/bazelbuild/rules_swift/pull/1136/checks?check_run_id=18673968570
If not I could try to revert those changes and re-do them myself.

luispadron:
> What would you think of expanding the scope of this change to include all derived module names and for the feature flags to control the PascalCase and package-including behavior for any module name that is not set explicitly. For example, when using rules_swift_package_manager, I added a directive to explicitly convert module names to PascalCase, but if this could just be configured at the rules_swift level that could be interesting.

That seems reasonable to me, I'll wait for others to leave their thoughts as well.

AttilaTheFun:
Hmm I found that rules_xcodeproj is calling this function here as well:
https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/xcodeproj/internal/build_settings.bzl#L51
I can add default arguments for backwards compatibility, but the way they're handling this feels very brittle.
I think they should be getting the module name through the provider. I can try to make a PR there as well.

AttilaTheFun:
closing in favor of https://github.com/bazelbuild/rules_swift/pull/1140

PR #1137, opened 2023-11-17
Add support for importing swift_binary

luispadron:
Currently when attempting to use `swift_binary` within the `deps` of a `swift_test` the module is not importable, for example:

BUILD:

```bzl
load("//swift:swift.bzl", "swift_binary", "swift_test")

swift_binary(
    name = "hello_world",
    srcs = ["main.swift"],
    module_name = "HelloWorld",
)

swift_test(
    name = "hello_world_test",
    srcs = ["main_test.swift"],
    deps = [":hello_world"],
)
```

Error:

```
examples/xplatform/hello_world/main_test.swift:3:18: error: no such module 'HelloWorld'
@testable import HelloWorld
                 ^
error: fatalError
```

Importing the library of a Swift binary is a common pattern in Swift Package Manager, and given `swift_binary` is unlike the Apple packaging rules in that it allows `srcs`, one could conceivably want to test (or use these sources in another module) without needing to split the target up into a `swift_library` and `swift_binary`. 

See [this issue](https://github.com/cgrindel/rules_swift_package_manager/issues/752) for a real use case where breaking it up into `swift_library` and `swift_test` pair complicates things. 

This PR adds support for being able to `import` the underlying library of a `swift_binary` target.

### DIFF
--- a/swift/internal/swift_binary_test.bzl
+++ b/swift/internal/swift_binary_test.bzl
@@ -19,9 +19,9 @@ load(":compiling.bzl", "output_groups_from_other_compilation_outputs")
 load(":derived_files.bzl", "derived_files")
 load(":feature_names.bzl", "SWIFT_FEATURE_BUNDLED_XCTESTS")
 load(":linking.bzl", "binary_rule_attrs", "configure_features_for_binary", "register_link_binary_action")
-load(":providers.bzl", "SwiftCompilerPluginInfo", "SwiftToolchainInfo")
+load(":providers.bzl", "SwiftCompilerPluginInfo", "SwiftInfo", "SwiftToolchainInfo")
 load(":swift_common.bzl", "swift_common")
-load(":utils.bzl", "expand_locations", "get_providers")
+load(":utils.bzl", "compact", "expand_locations", "get_providers")
 load(":env_expansion.bzl", "expanded_env")
 
 def _maybe_parse_as_library_copts(srcs):
@@ -84,8 +84,11 @@ def _swift_linking_rule_impl(
     cc_feature_configuration = swift_common.cc_feature_configuration(
         feature_configuration = feature_configuration,
     )
+    deps = ctx.attr.deps
     srcs = ctx.files.srcs
     user_link_flags = list(linkopts)
+    output_files = []
+    providers = []
 
     # If the rule has sources, compile those first and collect the outputs to
     # be passed to the linker.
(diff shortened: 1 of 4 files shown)


swift/internal/swift_binary_test.bzl
```diff
@@ -127,16 +130,36 @@ def _swift_linking_rule_impl(
...
+                compilation_context = module_context.clang.compilation_context,
+                linking_context = linking_context,
+            ),
+            coverage_common.instrumented_files_info(
+                ctx,
+                dependency_attributes = ["deps", "private_deps"],
+                extensions = ["swift"],
+                source_attributes = ["srcs"],
+            ),
+        ]
```
luispadron, line 160: Most of this was taken from `swift_library`, not exactly sure what is and is not required here.

swift/internal/swift_binary_test.bzl
```diff
@@ -127,16 +130,36 @@ def _swift_linking_rule_impl(
...
     else:
         module_context = None
-        cc_compilation_outputs = cc_common.create_compilation_outputs()
```
luispadron, line 139: This was unused

swift/internal/swift_binary_test.bzl
```diff
@@ -327,20 +342,14 @@ def _swift_test_impl(ctx):
...
                 files = ctx.files.data + additional_test_outputs,
                 transitive_files = ctx.attr._apple_coverage_support.files,
             ),
         ),
-        coverage_common.instrumented_files_info(
-            ctx,
-            dependency_attributes = ["deps"],
-            extensions = ["swift"],
-            source_attributes = ["srcs"],
-        ),
```
luispadron, line 343: This now happens above for both `swift_binary` (with srcs) and `swift_test`

swift/internal/swift_binary_test.bzl
luispadron: I can also imagine wanting to do this only for `swift_binary` and not `swift_test` since this is shared. Should i move this outside of this shared method

brentleyjones:
I think we want to take 6cae838e1fcbcc344fa2a29dc272261e719cba6e from upstream, and then it might unlock this more (or for free)?

> `swift_binary` and `swift_compiler_plugin` propagate this provider that wraps
`CcInfo` and `SwiftInfo` providers, instead of propagating them directly, so
that `swift_test` targets can depend on those binaries and test their modules
(similar to what Swift Package Manager allows) without allowing any
`swift_library` to depend on an arbitrary binary.

luispadron:
Ah didn't see that

luispadron:
Is that something we could get in soon 🙏. I'll close this for now, going off that comment it seems like exactly what we want

brentleyjones:
You can try cherry picking it. I'm not sure if it depends on other upstream changes.

PR #1139, opened 2023-11-27
Replace --swiftcopt with starlark flag

keith:
The Swift configuration fragment is being removed from bazel, this adds
the starlark flags necessary to replace the functionality. This differs
from the upstream solution since google doesn't have a need to support a
--host_swiftcopt replacement

(cherry picked from commit 8be724720ee6d1b73c52851106cf95b3cf158b3f)

### DIFF
--- a/swift/internal/swift_toolchain.bzl
+++ b/swift/internal/swift_toolchain.bzl
@@ -21,6 +21,7 @@ toolchain, see `swift.bzl`.
 
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load(":actions.bzl", "swift_action_names")
 load(":attrs.bzl", "swift_toolchain_driver_attrs")
@@ -292,11 +293,19 @@ def _swift_toolchain_impl(ctx):
             toolchain_root,
         )
 
+    # TODO: b/312204041 - Remove the use of the `swift` fragment once we've
+    # migrated the `--swiftcopt` flag via `--flag_alias`.
+    swiftcopts = list(ctx.fragments.swift.copts())
+    if "-exec-" in ctx.bin_dir.path:
+        swiftcopts.extend(ctx.attr._exec_copts[BuildSettingInfo].value)
+    else:
+        swiftcopts.extend(ctx.attr._copts[BuildSettingInfo].value)
+
     # Combine build mode features, autoconfigured features, and required
     # features.
     requested_features = (
         features_for_build_modes(ctx) +
-        features_from_swiftcopts(swiftcopts = ctx.fragments.swift.copts())
+        features_from_swiftcopts(swiftcopts = swiftcopts)
     )
     requested_features.extend([
         SWIFT_FEATURE_CACHEABLE_SWIFTMODULES,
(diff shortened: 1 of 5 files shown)


keith:
This can be used with `--flag_alias` to make the calling syntax more concise:

```
  --flag_alias=host_scopt=@build_bazel_rules_swift//swift:exec_copt
  --flag_alias=scopt=@build_bazel_rules_swift//swift:copt
```

Looks to me like it's safe to make those aliases `swiftcopt` and `host_swiftcopt` to mirror the flags that exist today, but that might be too confusing until those are gone for good from bazel

swift/internal/swift_toolchain.bzl
```diff
@@ -292,11 +293,19 @@ def _swift_toolchain_impl(ctx):
             toolchain_root,
         )
 
+    # TODO: b/312204041 - Remove the use of the `swift` fragment once we've
+    # migrated the `--swiftcopt` flag via `--flag_alias`.
+    swiftcopts = list(ctx.fragments.swift.copts())
+    if "-exec-" in ctx.bin_dir.path:
```
brentleyjones, line 299: I think this can break with path remapping?

brentleyjones: https://github.com/bazelbuild/bazel/blob/4008966449aef78c0de1d7ce527e9277b77186f5/src/main/java/com/google/devtools/build/lib/starlarkbuildapi/BuildConfigurationApi.java#L81-L82 would work, if we were allowed to call it (I think we should ask for it to become public API). If not, I think we can look at https://github.com/bazelbuild/bazel/blob/4008966449aef78c0de1d7ce527e9277b77186f5/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java#L639-L646 in a transition?

keith: Doesn't look like that first one is exposed anywhere. According to this comment https://github.com/bazelbuild/bazel/blob/106aec622f0cd011467bdc63b50982afc961e2d5/src/main/starlark/builtins_bzl/common/python/py_executable.bzl#L690-L695 the second is the same case

brentleyjones:
https://github.com/bazelbuild/rules_swift/commit/95f49c91b7eaec25ef99c4541d45cc3cdefb90fa

swift/internal/swift_toolchain.bzl
brentleyjones: I left a comment on the issue. We can do this for now.

brentleyjones (approved):

swift/internal/swift_toolchain.bzl
gregestren: How often do you want distinct exec `copt` options? And who sets them? 

We've Starlarkized Bazel's exec transition ([details](https://docs.google.com/document/d/1fyeR2eYP7wlN36btaEV2gjZOYW_PANpJI-XNVErZXwQ/edit#heading=h.5mcn15i0e1ch)) with hope to turn this into a broader API that removes the need for `--host_*` / `--exec_*` flags. 

i.e. it's no longer the case that every Bazel build has the same exec transition semantics for all users.

I'm hoping we can move toward a model of "build with exec configuration X" or "build with exec configuration Y"  instead of encoding target config flags that only exist to trigger the exec config.

Caveats: this is post Bazel-7.0 and it's clearly bigger work than, e.g., just exposing one of the existing "is tool config" settings.

Aside,  I think this is safe for theoretical path remapping because it's Bazel analysis logic, not part of action execution.

keith: I think that largely varies per project and the --host_swiftcopt uses very closely mirror any case you'd have for --host_copt and the other related flags. I do think that in our use case it would be pretty fine for us to encode those in a transition, but the idea of having multiple transitions for the times you want to slightly vary one flag sounds a bit heavier than how we do it today

PR #1140, opened 2023-12-04
Implement new swift_proto_library rule

AttilaTheFun:
## Overview

This PR introduces a new swift_proto_library rule which is intended to replace both the swift_proto_library and swift_grpc_library rules. The PR marks those rules as deprecated and provides a comprehensive migration guide to the new rules.

## Why rewrite?

The new swift_proto_library rule was created to address several issues with the previous
swift_proto_library and swift_grpc_library rules which would be difficult if not impossible 
to fix while maintaining full backwards compatibility.

### 1. swift_protoc_gen_aspect

Firstly, the previous swift_proto_library rule used the aspect which traversed its direct 
proto_library dependencies and their transitive graph of proto_library dependencies,
and compiled all of these into swift modules. 

This meant that a single swift_proto_library target could generate and and compile swift source files
for *all* of the proto_library targets in potentially a very large dependency graph,
which could span across multiple repositories.

In practice, though, consumers typically preferred to have a 1:1 mapping of proto_library to swift_proto_library targets.
Additionally, this aspect meant that the module name and other compilation options could not be configured on a per-target basis, since the aspect needed to be able to determine all of this information by just looking at the providers of the proto_library targets.

This lack of flexibility was extremely frustrating, especially for consumers who wanted to assign "Swift-like" module names
to the swift proto library modules. 

Using the deprecated_grpc example in this repository to demonstrate,
consumers had to import the generated service swift proto module with:

```
import examples_xplatform_deprecated_grpc_echo_server_services_swift
```

But with the new swift_proto_library rule, 
they can assign an arbitrary module name to the target and import it with:

```
import ServiceServer
```

This removes the guesswork of trying to determine what the module name will be,
and it allows engineers to work with the new swift_proto_library targets exactly the same
way they would if these were swift_library targets they had written by hand,
except the source files are generated for them.

### 2. Implicit, internal proto compiler and runtime dependencies.

The second main issue the new implementation is intended to address is the specific proto compiler,
proto compiler plugin, and compile time proto library dependencies the old rules required.

The old rules depended internally on the SwiftProtobuf and GRPC targets defined in the targets under
the third_party directory. This was problematic for consumers who wanted to use different versions
of protoc, the protoc plugins or compile time dependencies.

Specifically, this issue came to a head when implementing support for proto targets in
rules_swift_package_manager. There was a lengthy slack conversation here on the topic:
https://bazelbuild.slack.com/archives/CD3QY5C2X/p1692055426375909

The problem was that some Swift packages like Firebase have package dependencies on protobuf
libraries which contain the same code (though perhaps a different version) as the protobuf
targets defined in this repository which were dependencies of those rules.

This meant that consumers could *either* use the Swift package dependencies *or*
swift_proto_library / swift_grpc_library but not both or else they would encounter linker errors.

The new swift_proto_library rule is completely configurable in terms of:
- the proto compiler target
- the proto compiler plugin target(s)
- the swift compile-time dependency target(s)

This is accomplished with a new swift_proto_compiler rule which is passed as an attribute to the new swift_proto_library rule.

The swift_proto_compiler rule represents the combination of a protoc binary,
protoc plugin binary (e.g. the SwiftProtobuf or grpc-swift protoc plugin),
and the set of options passed to the protoc plugin.

We provide pre-configured swift_proto_compiler targets for:
- swift protos
- swift server grpc protos
- swift client grpc protos

And 3rd parties are welcome to create their own, configured based on their needs.
In the case of rules_swift_package_manager, this will enable the creation of a swift_proto_compiler target
configured to use the same dependencies as the other Swift packages.

Finally, this allows us to only have a single, new swift_proto_library rule in lieu of the two
swift_proto_library and swift_grpc_library rules because the only real difference between them
is the plugin being passed to protoc and the options being passed to that plugin.


doc/proto_migration.md
```diff
@@ -0,0 +1,247 @@
...
+    deps = [":echo_proto_swift"],
+)
+```
+
+These can be migrated to the following new_swift_proto_library targets:
+
+```
+new_swift_proto_library(
+    name = "service_server_swift_proto",
+    compilers = ["//swift/proto/compilers:swift_server_proto"],
```
cgrindel, line 206: Should this include the repository name ?

```
    compilers = ["@rules_swift//swift/proto/compilers:swift_server_proto"],
```

proto/compilers/BUILD
```diff
@@ -0,0 +1,25 @@
+"""Aliases to provided proto compiler targets."""
```
cgrindel, line 1: Curious. Is there something about the actual targets that you don't like? These aliases seem to shorten target labels by one level.

swift/proto/swift_proto_compiler.bzl
```diff
@@ -0,0 +1,324 @@
...
+        )
+        module_mappings.append(module_mapping)
+    module_mappings_file = register_module_mapping_write_action(ctx.label.name, ctx.actions, module_mappings)
+
+    # Build the arguments for protoc:
+    args = ctx.actions.args()
+    args.use_param_file("--param=%s")
+
+    # Add the plugin argument with the provided name to namespace all of the options:
+    args.add("--plugin=protoc-gen-" + swift_proto_compiler_info.internal.plugin_name + "=" + swift_proto_compiler_info.internal.plugin.path)
```
cgrindel, line 172: I recommend using `"".format()` so that you can wrap these long lines.

cgrindel (approved):
I have a couple of comment/questions. Otherwise, LGTM. Thanks! This was a big effort.

doc/proto_migration.md
AttilaTheFun: @cgrindel If they're using the rules_swift proto compilers, yeah, I can add that. Though in the rules_swift_package_manager GRPC example, I demonstrated creating a custom compiler to use the swiftpkg GRPC / SwiftProtobuf dependencies which didn't need a workspace name because it was in the same workspace.

proto/compilers/BUILD
AttilaTheFun: I added these because, for backwards compatibility, the rule named "swift_proto_library" exported by swift/swift.bzl still points to the old implementation. (It logs a deprecation warning now.) The new implementation is also exported in that file, but as "new_swift_proto_library". Consumers *can* import that, but if they want cleaner import statements they can use the export from proto/proto.bzl instead. This also more closely aligns with rules_go which has proto/def.bzl:
https://github.com/bazelbuild/rules_go/blob/master/proto/def.bzl

AttilaTheFun:
@thii yes regarding multiple protos with the same name, it sounds like you already figured this out, but for others who read this, I'm deferring this responsibility to SwiftProtobuf which has a FileNaming option for the protoc plugin.

https://github.com/apple/swift-protobuf/blob/main/Documentation/PLUGIN.md#generation-option-filenaming---naming-of-generated-sources

The default value here is FullPath where "foo/bar/baz.proto" makes "foo/bar/baz.pb.swift" (relative to the package in bazel-bin). However, you can set this to PathToUnderscores via additional_plugin_options (or configure a swift_proto_compiler target to set this by default).

I updated //examples/xplatform/proto:example_path_to_underscores_proto_swift to support this:
```
swift_proto_library(
    name = "example_path_to_underscores_proto_swift",
    additional_plugin_options = {
        "FileNaming": "PathToUnderscores",
    },
    protos = [":example_path_to_underscores_proto"],
)
```

swift/proto/swift_proto_library.bzl
```diff
@@ -0,0 +1,330 @@
...
+                default = [],
+                doc = """\
+                A list of targets that are dependencies of the target being built, which will be
+                linked into that target, but will be ignored by the proto compiler.
+                """,
+            ),
+            "compilers": attr.label_list(
+                default = ["//proto/compilers:swift_proto"],
+                providers = [SwiftProtoCompilerInfo],
+            ),
```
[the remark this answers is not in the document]
AttilaTheFun: Done, thanks!

doc/proto_migration.md
```diff
@@ -0,0 +1,247 @@
+# Overview
+
+This document aims to provide context to future contributors on why we decided to rewrite
+the swift_proto_library and swift_grpc_library rules into the new swift_proto_library rule,
```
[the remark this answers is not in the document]
AttilaTheFun: Sure, I went through and updated the migration doc to improve the formatting.

AttilaTheFun:
Thanks so much @thii ! I went through and updated the documentation as requested, and also responded re: the PathToUnderscores parameter though it looks like you figured that one out already.

AttilaTheFun:
Thanks for all the feedback @thii !

Firstly, my new rule actually *does* support globbing multiple proto packages into a single proto_library which is generated by a single swift_proto_library. I added a proto_glob example to the repo to demonstrate this. The only drawback is that the new rule only generates a single swift module per swift_proto_library similar to swift_library. 

The deprecated rule used the aspect to generate a swift module per proto package, but you could not really configure anything about these modules because they had to be entirely derived from the ProtoInfo. If if you tried to glob thousands of protos with the new rule, you would potentially have long compile times. SwiftProtobuf prefixes the names of messages with their package name, though, so name conflicts are not possible. In the example you can see I have Message1 in package_1 and Message2 in package_2 which SwiftProtobuf generates as Package1_Message1 and Package2_Message2. These are both in the same swift module.

Would this be sufficient to cover your use case? Or are the compile times too problematic?

If the compile times are too much of an issue, I could possibly follow this up by adding a swift_proto_library_group rule that tries to replicate the previous swift-module-per-proto-package behavior while using the new internals including swift_proto_compiler. I'm still of the opinion that a single rule generating hundreds if not thousands of modules is a bad idea, though, and I would still encourage people to either handwrite the BUILD files or use a BUILD file generator like gazelle.

Regarding gazelle, yes a go toolchain is not ideal. 
I'd love to see an alternative where the core build file generation logic is all in Starlark,
but the plugins are Bazel executables written in an arbitrary language, so e.g. go could use go, swift could use swift etc.
Maybe that will be my next project after merging these PRs.

The gazelle plugin for generating `proto_library` targets is already included in the main gazelle repo:
https://github.com/bazelbuild/bazel-gazelle/blob/da9ab082db51af091e6de25fdb495e9657b9c62e/language/proto/BUILD.bazel#L37

The gazelle plugin for generating `swift_proto_library` (and `swift_library`) targets is in the rules_swift_package_manager repo, but it does not require the use of SPM. I've actually flagged this as an issue to @cgrindel before -- I believe that keeping the gazelle plugin in the rules_swift_package_manager repo limits its discoverability, and discourages people from using it who erroneously believe that they would need to adopt SPM to use it. IMHO we should extract it into a separate repo named `rules_swift_gazelle_plugin` or something similar.

I have a PR here to update it to support the new `swift_proto_library` rule introduced in this PR:
https://github.com/cgrindel/rules_swift_package_manager/pull/834

doc/proto_migration.md
```diff
@@ -0,0 +1,250 @@
...
+which traversed its direct `proto_library` dependencies and their transitive graph of `proto_library` dependencies,
+and compiled all of these into swift modules. 
+
+This meant that a single `swift_proto_library` target could generate and and compile swift source files
+for *all* of the `proto_library` targets in potentially a very large dependency graph,
+which could span across multiple repositories.
+
+In practice, though, consumers typically preferred to have a 1:1 mapping of `proto_library` to `swift_proto_library` targets.
+Additionally, this aspect meant that the module name and other compilation options could not be configured on a per-target basis,
+since the aspect needed to be able to determine all of this information by just looking at the providers of the `proto_library` targets.
```
luispadron, line 29:
```suggestion
In practice, though, consumers typically preferred to have more control over the mapping of `proto_library` to `swift_proto_library` targets.
Additionally, this aspect meant that the module name and other compilation options could not be configured on a per-target basis,
since the aspect needed to be able to determine all of this information by just looking at the providers of the `proto_library` targets.
```

doc/proto_migration.md
```diff
@@ -0,0 +1,250 @@
...
+
+This lack of flexibility was extremely frustrating, especially for consumers who wanted to assign "Swift-like" module names
+to the `swift_proto_library` modules.
```
luispadron, line 32: nit: imo think the point is made in the previous and next paragraph.

```suggestion
```

doc/proto_migration.md
```diff
@@ -0,0 +1,250 @@
...
+the `third_party` directory. This was problematic for consumers who wanted to use different versions
+of protoc, the protoc plugins or compile time dependencies.
+
+Specifically, this issue came to a head when implementing support for proto targets in
+`rules_swift_package_manager`. There was a lengthy slack conversation here on the topic:
+https://bazelbuild.slack.com/archives/CD3QY5C2X/p1692055426375909
+
+The problem was that some Swift packages like Firebase have package dependencies on protobuf
+libraries which contain the same code (though perhaps a different version) as the protobuf
+targets defined in this repository which were dependencies of those rules.
```
luispadron, line 68:
```suggestion
The problem was that some Swift packages have package dependencies on protobuf
libraries which contain the same code (though perhaps a different version) as the protobuf
targets defined in this repository which were dependencies of those rules.
```

doc/proto_migration.md
```diff
@@ -0,0 +1,250 @@
...
+This is because TestClient protoc plugin option was deprecated with Swift 5.6,
+as the test clients do not conform to Sendable.
+This capability may be removed in a future major version update.
+
+The alternative recommended by the `grpc-swift` authors is to register a mock server on localhost.
+
+## F.A.Q.
+
+As consumers raise questions during their migrations to the new `swift_proto_library` rule,
+we will add their answers here as a reference.
```
luispadron, line 250: Some I can think of (and potentially we just link to examples here):

- Can I group protos to reduce number of modules?
- Can I use a completely custom protobuf compiler?

doc/rules.md
```diff
@@ -20,15 +20,86 @@ On this page:
...
+
+Generates a Swift static library from one or more targets producing `ProtoInfo`.
+
+```python
+proto_library(
+    name = "foo",
+    srcs = ["foo.proto"],
+)
+
+swift_proto_library(
```
luispadron, line 53:
```suggestion
new_swift_proto_library(
```

doc/rules.md
```diff
@@ -20,15 +20,86 @@ On this page:
...
+swift_proto_library targets which mirror the dependencies between the proto targets.
+
+```python
+proto_library(
+    name = "bar",
+    srcs = ["bar.proto"],
+    deps = [":foo"],
+)
+
+swift_proto_library(
```
luispadron, line 69:
```suggestion
new_swift_proto_library(
```

swift/proto/swift_proto_compiler.bzl
```diff
@@ -0,0 +1,382 @@
...
+            continue
+
+        merged_proto_paths_to_module_names[path] = module_name
+        module_proto_paths = module_names_to_proto_paths[module_name] if module_name in module_names_to_proto_paths else []
+        module_proto_paths.append(path)
+        module_names_to_proto_paths[module_name] = module_proto_paths
+
+    # Write the module mappings to a file:
+    module_mappings = []
+    for module_name in sorted(module_names_to_proto_paths.keys()):
```
luispadron, line 149: Is the sort for keys here needed?

swift/proto/swift_proto_compiler.bzl
```diff
@@ -0,0 +1,382 @@
...
+            doc = """\
+            A proto compiler plugin executable binary.
+            
+            E.g.
+            "//tools/protoc_wrapper:protoc-gen-grpc-swift"
+            "//tools/protoc_wrapper:ProtoCompilerPlugin"
+
+            We provide two compiler targets:
+            "//swift/proto:swift_proto"
+            "//swift/proto:swift_grpc"
```
luispadron, line 299: Should these be the plugin targets?

swift/proto/swift_proto_library.bzl
```diff
@@ -0,0 +1,324 @@
...
+            requires_srcs = False,
+        ),
+        {
+            "protos": attr.label_list(
+                doc = """\
+                Exactly one `proto_library` target (or a target producing `ProtoInfo`),
+                from which the Swift library should be generated.
+                """,
+                providers = [ProtoInfo],
+            ),
```
luispadron, line 261: Thoughts on updating the name / type to make it more obvious its only one (or does it need to be a list)
```suggestion
            "dep": attr.label(
                doc = """\
                Exactly one `proto_library` target (or a target producing `ProtoInfo`),
                from which the Swift library should be generated.
                """,
                providers = [ProtoInfo],
            ),
```

swift/proto/swift_proto_library.bzl
```diff
@@ -0,0 +1,324 @@
...
+    doc = """\
+    Generates a Swift static library from one or more targets producing `ProtoInfo`.
+
+    ```python
+    proto_library(
+        name = "foo",
+        srcs = ["foo.proto"],
+    )
+
+    swift_proto_library(
```
luispadron, line 299: Should the documentation use `new_swift_proto_library` until the other ones are removed?

swift/proto/swift_proto_utils.bzl
luispadron, line 1: Should this file be under `swift/proto/internal`

luispadron: If we intend to make it public, I think putting the functions in a struct like `swift_proto_support` would match other rules

luispadron (approved):
Thank you for your effort on this! It feels like a good improvement over the existing rules and allows for easier extension.

doc/proto_migration.md
AttilaTheFun: Added answers to these questions

doc/rules.md
AttilaTheFun: So I needed to give these another name in swift/swift.bzl to avoid breaking consumers of the old rules, but I created aliases to the new rules under proto/BUILD named swift_proto_library and swift_proto_compiler. These are analogous to rules_go's structure.

I expect new consumers to use these aliases rather than those from swift.bzl (though it's fine if they want to use the non-aliased ones too). I can a load statement here to make it clear that they have the same name but different implementations.

doc/rules.md
AttilaTheFun: Replied above

doc/rules.md
luispadron: That works!

swift/proto/swift_proto_library.bzl
AttilaTheFun: No, actually I forgot do update this doc comment -- it actually works with an arbitrary number of proto targets!
I added the proto_files example demonstrating this in action. I'll update the doc comment.

swift/proto/swift_proto_compiler.bzl
AttilaTheFun: It's not strictly necessary, no, but it makes debugging the generated file a bit easier if the module names are always in lexicographical order.

swift/proto/swift_proto_utils.bzl
AttilaTheFun: No these are not intended to be public -- I can move them under swift/proto/internal.

swift/proto/swift_proto_compiler.bzl
AttilaTheFun: The plugins go into the compiler rule, not the library rule. From the library rule's perspective there is just a compiler. I updated the documentation to reflect this.

swift/proto/swift_proto_utils.bzl
AttilaTheFun: Actually, at least the proto_path function I want to export for 3rd party rule authors. I'll make a `swift_proto_common` struct and export it similar to `swift_common` (we don't need to shove all of the proto stuff in there).

examples/xplatform/custom_swift_proto_compiler/rules/custom_proto_compiler_plugin.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# Echo a message to show that the plugin was run:
+echo "Plugin"
```
brentleyjones, line 4: Nit: can you add a newline at the end of the file?

examples/xplatform/custom_swift_proto_compiler/rules/custom_proto_compiler.sh
```diff
@@ -1,4 +1,18 @@
...
+  source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v3 ---
+
+CUSTOM_PROTO_COMPILER_PLUGIN_PATH=$(rlocation build_bazel_rules_swift/examples/xplatform/custom_swift_proto_compiler/rules/custom_proto_compiler_plugin)
+$CUSTOM_PROTO_COMPILER_PLUGIN_PATH
 
 # Touch all of the declared Swift sources to create an empty file.
 touch $@
```
brentleyjones, line 18: Nit: can you add a newline at the end of the file?

examples/xplatform/custom_swift_proto_compiler/protos/example.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package example;
+
+message Example {
+  string name = 1;
+}
```
brentleyjones, line 7: Nit: can you add a newline at the end of the file?

swift/proto/copy_swift_sources.sh.tpl
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Recursively copy the Swift sources from the temporary directory to the permanent one:
+cp -RL {temporary_output_directory_path}/** {permanent_output_directory_path}/
+
+# Touch all of the declared Swift sources to create an empty file if the plugin didn't generate it:
+touch {swift_source_file_paths}
```
brentleyjones, line 7: Nit: can you add a newline at the end of the file?

doc/rules.md
```diff
@@ -20,15 +20,92 @@ On this page:
   * [swift_compiler_plugin](#swift_compiler_plugin)
   * [universal_swift_compiler_plugin](#universal_swift_compiler_plugin)
   * [swift_feature_allowlist](#swift_feature_allowlist)
-  * [swift_grpc_library](#swift_grpc_library)
   * [swift_import](#swift_import)
   * [swift_library](#swift_library)
   * [swift_library_group](#swift_library_group)
   * [swift_module_alias](#swift_module_alias)
   * [swift_package_configuration](#swift_package_configuration)
+  * [swift_grpc_library](#swift_grpc_library)
```
brentleyjones, line 28: If we are moving this, do we want it after the proto ones?

doc/rules.md
```diff
@@ -20,15 +20,92 @@ On this page:
...
   * [swift_proto_library](#swift_proto_library)
+  * [swift_proto_compiler](#swift_proto_compiler)
+  * [new_swift_proto_library](#new_swift_proto_library)
```
brentleyjones, line 31: If we want people to use this instead, should we list it earlier?

doc/rules.md
```diff
@@ -20,15 +20,92 @@ On this page:
...
+                        <a href="#new_swift_proto_library-generated_header_name">generated_header_name</a>, <a href="#new_swift_proto_library-generates_header">generates_header</a>, <a href="#new_swift_proto_library-linkopts">linkopts</a>, <a href="#new_swift_proto_library-linkstatic">linkstatic</a>, <a href="#new_swift_proto_library-module_name">module_name</a>,
+                        <a href="#new_swift_proto_library-package_name">package_name</a>, <a href="#new_swift_proto_library-plugins">plugins</a>, <a href="#new_swift_proto_library-protos">protos</a>, <a href="#new_swift_proto_library-swiftc_inputs">swiftc_inputs</a>)
+</pre>
+
+Generates a Swift static library from one or more targets producing `ProtoInfo`.
+
+```python
+load("//proto:proto.bzl", "swift_proto_library")
+load("@rules_proto//proto:defs.bzl", "proto_library")
```
brentleyjones, line 49:
```suggestion
load("@rules_proto//proto:defs.bzl", "proto_library")
load("//proto:proto.bzl", "swift_proto_library")
```

doc/rules.md
```diff
@@ -20,15 +20,92 @@ On this page:
...
+    protos = [":foo"],
+)
+```
+
+If your protos depend on protos from other targets, add dependencies between the
+swift_proto_library targets which mirror the dependencies between the proto targets.
+
+```python
+load("//proto:proto.bzl", "swift_proto_library")
+load("@rules_proto//proto:defs.bzl", "proto_library")
```
brentleyjones, line 67:
```suggestion
load("@rules_proto//proto:defs.bzl", "proto_library")
load("//proto:proto.bzl", "swift_proto_library")
```

doc/rules.md
```diff
@@ -248,6 +325,10 @@ package.
 swift_grpc_library(<a href="#swift_grpc_library-name">name</a>, <a href="#swift_grpc_library-deps">deps</a>, <a href="#swift_grpc_library-srcs">srcs</a>, <a href="#swift_grpc_library-flavor">flavor</a>)
 </pre>
 
+DEPRECATED -- Please use new_swift_proto_library rule instead.
+This rule will be removed in the next rules_swift major version update.
+If you're already using this rule, see doc/proto_migration.md for infomation on how to migrate.
```
brentleyjones, line 330: Pretty sure we can have `doc/proto_migration.md` render as a link.

doc/rules.md
```diff
@@ -248,6 +325,10 @@ package.
 swift_grpc_library(<a href="#swift_grpc_library-name">name</a>, <a href="#swift_grpc_library-deps">deps</a>, <a href="#swift_grpc_library-srcs">srcs</a>, <a href="#swift_grpc_library-flavor">flavor</a>)
 </pre>
 
+DEPRECATED -- Please use new_swift_proto_library rule instead.
```
brentleyjones, line 328: Pretty sure we can have `new_swift_proto_library` render as a link.

doc/rules.md
```diff
@@ -479,6 +586,10 @@ target's label is included by the package specifications in the configuration.
 swift_proto_library(<a href="#swift_proto_library-name">name</a>, <a href="#swift_proto_library-deps">deps</a>)
 </pre>
 
+DEPRECATED -- Please use new_swift_proto_library rule instead.
+This rule will be removed in the next rules_swift major version update.
+If you're already using this rule, see doc/proto_migration.md for infomation on how to migrate.
```
brentleyjones, line 591: Same link comments.

examples/xplatform/custom_swift_proto_compiler/BUILD.bazel
brentleyjones, line 1: We use `BUILD` instead of `BUILD.bazel` in this repo if possible. Can you rename all of the new files to follow that format?

brentleyjones (approved):

swift/proto/swift_proto_library.bzl
```diff
@@ -0,0 +1,316 @@
...
+swift_proto_library = rule(
+    attrs = dicts.add(
+        swift_common.library_rule_attrs(
+            additional_deps_aspects = [
+                swift_clang_module_aspect,
+            ],
+            requires_srcs = False,
+        ),
+        {
+            "protos": attr.label_list(
```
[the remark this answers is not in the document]
AttilaTheFun: @brentleyjones Yes, deps *is* used, just as it is on swift_library, for dependencies between targets providing Swift modules (or non-Swift clang modules). I prefer to reserve protos for the dependencies providing ProtoInfo (typically proto_library targets), and deps for the dependencies imported as modules from the generated Swift code.

This is how go_proto_library works as well:
```
go_proto_library(
    name = "agora_service_go_proto",
    compilers = ["@io_bazel_rules_go//proto:go_grpc"],
    importpath = "github.com/attilathefun/app_engine/protos/services/public/agora_service",
    proto = ":agora_service_proto",
    visibility = ["//visibility:public"],
    deps = ["//protos/messages/public/agora_service_messages:agora_service_messages_go_proto"],
)
```

swift/swift.bzl
```diff
@@ -110,15 +122,24 @@ swift_c_module = _swift_c_module
...
 
+# Deprecated swift proto rules:
+swift_proto_library = _deprecated_swift_proto_library
+swift_grpc_library = _deprecated_swift_grpc_library
+SwiftGRPCInfo = _DeprecatedSwiftGRPCInfo
+
+# Swift proto rule:
+SwiftProtoCompilerInfo = _SwiftProtoCompilerInfo
+swift_proto_compiler = _swift_proto_compiler
+new_swift_proto_library = _swift_proto_library
```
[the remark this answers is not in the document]
AttilaTheFun: @brentleyjones sure I can do that. I added aliases under //proto which I expected people to actually use, but we could also just expose a proto.bzl file in //swift/proto publicly instead. I couldn't remove or rename the existing swift_proto_library and swift_grpc_library exports in this file without breaking existing consumers of the deprecated rules though.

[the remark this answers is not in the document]
brentleyjones: Yeah, we can leave those here, and the deprecation note can send people to the new location. Looks like `rules_go` uses the `proto` package as well for this, which makes me want to do that even more.

doc/rules.md
```diff
@@ -1,33 +1,207 @@
...
-rules. Do not import definitions from the `internal` subdirectory directly.
-
-To use the Swift build rules in your BUILD files, load them from
-`@build_bazel_rules_swift//swift:swift.bzl`.
-
-For example:
-
-```build
-load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
-```
```
brentleyjones, line 15: Seems we might want something like this still?

doc/BUILD
```diff
@@ -12,26 +12,29 @@ _DOC_SRCS = {
...
         "SwiftInfo",
         "SwiftToolchainInfo",
+        "SwiftProtoCompilerInfo",
         "SwiftProtoInfo",
         "SwiftUsageInfo",
+        "DeprecatedSwiftGRPCInfo",
     ],
     "rules": [
+        "swift_proto_library",
+        "swift_proto_compiler",
```
brentleyjones, line 24: Can you put these just below `swift_test`?

brentleyjones (approved):

Issue #1142, opened 2024-01-08
Error in new_objc_provider: Key 'linkopt' no longer supported in ObjcProvider



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3575#018ce2f8-2590-45f5-9afe-3d2cdb2e6b7d

 Platform: MacOS

 Logs: 
```(09:30:39) ERROR: /private/var/tmp/_bazel_buildkite/c8f5eeee7f112b07250892b1524f7f4f/external/_main~non_module_deps~build_bazel_rules_swift_local_config/BUILD:9:22: in xcode_swift_toolchain rule @@_main~non_module_deps~build_bazel_rules_swift_local_config//:toolchain:
Traceback (most recent call last):
	File "/Users/buildkite/builds/bk-imacpro-18/bazel-org-repo-root/rules_swift/swift/internal/xcode_swift_toolchain.bzl", line 589, column 57, in _xcode_swift_toolchain_impl
		swift_linkopts_providers = _swift_linkopts_providers(
	File "/Users/buildkite/builds/bk-imacpro-18/bazel-org-repo-root/rules_swift/swift/internal/xcode_swift_toolchain.bzl", line 257, column 51, in _swift_linkopts_providers
		objc_info = apple_common.new_objc_provider(linkopt = depset(linkopts)),
Error in new_objc_provider: Key 'linkopt' no longer supported in ObjcProvider (use CcInfo instead).
```

 Steps:
```
git clone https://github.com/bazelbuild/rules_swift.git 
git reset 7c82ef1e7e503cedb8d4c69b049e76412cc3572a --hard
export USE_BAZEL_VERSION=8e8ddaba0e90c280bfd85644d6ccd91df5b6d353
bazel build //examples/... 
```
 CC Greenteam @fweikert

brentleyjones:
This is from `--incompatible_objc_provider_remove_linking_info` flipping in https://github.com/bazelbuild/bazel/commit/67c5c0fbf99c80aeb17073c79e8017bde443fd4b.

PR #1143, opened 2024-01-08
Fix compatibility with --incompatible_objc_provider_remove_linking_info

keith:
Fixes https://github.com/bazelbuild/rules_swift/issues/1142

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -45,6 +45,9 @@ load(":providers.bzl", "SwiftToolchainInfo")
 load(":swift_clang_module_aspect.bzl", "swift_clang_module_aspect")
 load(":utils.bzl", "get_providers")
 
+# TODO: Remove once we drop bazel 7.x
+_OBJC_PROVIDER_LINKING = hasattr(apple_common.new_objc_provider(), "linkopt")
+
 def binary_rule_attrs(
         *,
         additional_deps_providers = [],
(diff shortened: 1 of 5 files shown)


test/imported_framework_tests.bzl
```diff
@@ -17,17 +17,6 @@ def imported_framework_test_suite(name):
         target_under_test = "@build_bazel_rules_swift//test/fixtures/linking:bin",
     )
 
-    action_command_line_test(
-        name = "{}_duplicate_linking_args".format(name),
-        expected_argv = [
-            "-framework framework1",
```
keith, line 23: this test is no longer applicable since the files are passed directly

swift/internal/linking.bzl
```diff
@@ -439,23 +439,28 @@ def new_objc_provider(
...
             deps,
             apple_common.Objc,
         ) + additional_objc_infos,
-    )
+    }
+    empty_provider = apple_common.new_objc_provider()
+    for key in dict(kwargs):
+        if not hasattr(empty_provider, key):
+            kwargs.pop(key, None)
+    return apple_common.new_objc_provider(**kwargs)
```
brentleyjones, line 463: This seems expensive. Could we instead use a single global check (create this provider globally and look for a sentinel value?) and `key not in _REMOVED_OBJC_PROVIDER_KEYS` instead?

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -243,6 +243,10 @@ def _swift_linkopts_providers(
         "-Wl,-rpath,/usr/lib/swift",
     ])
 
+    objc_info = apple_common.new_objc_provider()
+    if hasattr(objc_info, "linkopt"):
+        objc_info = apple_common.new_objc_provider(linkopt = depset(linkopts))
```
brentleyjones, line 248: Same here

keith: done

swift/internal/linking.bzl
keith: done

brentleyjones (approved):

luispadron (approved):

PR #1146, opened 2024-01-09
Disable builtin C++ toolchain in BCR presubmit

keith:
Now that the BCR uses bzlmod the builtin toolchain is preferred which
doesn't work for us. This wouldn't be an issue if the tests in the bcr
ran within this repo


brentleyjones (approved):

mattrobmattrob:
Which error does this fix, @keith?

keith:
https://buildkite.com/bazel/bcr-presubmit/builds/3048#018cef7b-393c-4cc5-821a-844533aba56d

PR #1147, opened 2024-01-12
Add swift.treat_warnings_as_errors feature

keith:
This makes it easier for you to negate this feature and not have to
worry about command line flag ordering


keith:
we could jump on the CC feature for this but I could see folks wanting these to be separate as well

thii (approved):

Issue #1148, opened 2024-01-16
Error: There is no configuration fragment named 'swift'. Available fragments: 'platform', 'coverage', 'bazel_py', 'android', 'bazel_android', 'apple', 'cpp', 'java', 'j2objc', 'objc', 'proto', 'py'



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3594#018d1012-3183-4ff1-9791-7fffb20420b9

 Platform: MacOS

 Logs: 
```ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/94c266353aabbe7180eecb7aa311392a/external/_main~non_module_deps~build_bazel_rules_swift_local_config/BUILD:9:16: in swift_toolchain rule @@_main~non_module_deps~build_bazel_rules_swift_local_config//:toolchain:
Traceback (most recent call last):
	File "/var/lib/buildkite-agent/builds/bk-docker-5xrg/bazel-org-repo-root/rules_swift/swift/internal/swift_toolchain.bzl", line 298, column 36, in _swift_toolchain_impl
		swiftcopts = list(ctx.fragments.swift.copts())
Error: There is no configuration fragment named 'swift'. Available fragments: 'platform', 'coverage', 'bazel_py', 'android', 'bazel_android', 'apple', 'cpp', 'java', 'j2objc', 'objc', 'proto', 'py'
(02:41:23) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/94c266353aabbe7180eecb7aa311392a/external/_main~non_module_deps~build_bazel_rules_swift_local_config/BUILD:9:16: Analysis of target '@@_main~non_module_deps~build_bazel_rules_swift_local_config//:toolchain' failed
(02:41:23) ERROR: Analysis of target '//examples/xplatform/grpc:echo_client_services_swift' failed; build aborted: Analysis failed
(02:41:23) INFO: Elapsed time: 3.548s, Critical Path: 0.19s
(02:41:23) INFO: 5 processes: 5 internal.
(02:41:23) ERROR: Build did NOT complete successfully
(02:41:23) FAILED:
```
 Steps:
 ```
git clone -v https://github.com/bazelbuild/rules_swift.git
git reset 0919507c7c6c23d410a15b67c01e5f8025ea96b8 --hard
export USE_BAZEL_VERSION=cddec1247571f1cda9939ffb7546e089e528c76e
[...]

PR #1149, opened 2024-01-18
Add support for removed swift fragment

keith:
The swift fragment was removed at bazel HEAD and replaced by our custom
starlark flag for copts in rules swift. This requires a version check
that we're relying on bazel_features for if we want to be 100% accurate.

Fixes https://github.com/bazelbuild/rules_swift/issues/1148

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,7 @@ module(
     repo_name = "build_bazel_rules_swift",
 )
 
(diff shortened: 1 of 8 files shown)


keith:
depends on https://github.com/bazel-contrib/bazel_features/pull/36

luispadron (approved):

brentleyjones (approved):

PR #1155, opened 2024-01-19
Add `swift_library_group` rule

brentleyjones:
The `swift_library_group` rule acts as an alias for multiple other `{cc,objc,swift}_library`s. This will allow rules_swift_package_manager to properly map SPM’s library products to a target that can be added to other target’s `deps`.

### DIFF
--- a/doc/rules.md
+++ b/doc/rules.md
@@ -394,7 +394,11 @@ swift_library_group(<a href="#swift_library_group-name">name</a>, <a href="#swif
 </pre>
 
 Groups Swift compatible libraries (e.g. `swift_library` and `objc_library`).
-Can be used anywhere a `swift_library` can be used.
+The target can be used anywhere a `swift_library` can be used. It behaves
+similar to source-less `{cc,obj}_library` targets.
+
+Unlike `swift_module_alias`, a new module isn't created for this target, you
+need to import the grouped libraries directly.
 
 **ATTRIBUTES**
 
--- a/doc/rules.md
+++ b/doc/rules.md
@@ -23,6 +23,7 @@ On this page:
   * [swift_grpc_library](#swift_grpc_library)
   * [swift_import](#swift_import)
   * [swift_library](#swift_library)
+  * [swift_library_group](#swift_library_group)
   * [swift_module_alias](#swift_module_alias)
   * [swift_package_configuration](#swift_package_configuration)
   * [swift_proto_library](#swift_proto_library)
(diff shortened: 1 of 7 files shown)


swift/internal/linking.bzl
```diff
@@ -320,26 +323,51 @@ def create_linking_context_from_compilation_outputs(
...
+                            user_link_flags = user_link_flags,
+                        ),
+                    ],
+                ),
+            ),
+        )
+        linking_context = cc_common.merge_linking_contexts(
+            linking_contexts = linking_contexts + extra_linking_contexts,
+        )
+        linking_output = None
```
brentleyjones, line 368: Copying how `objc_library` does it: This makes `swift_library` behave more like `{cc,objc}_library`. It allows for a target that:

- propagates `defines`
- propagates `linkopts`
- acts as an alias for multiple other `{cc,objc,swift}_library`s

The last one is the reason I’m making this change. This will alow rules_swift_package_manager to properly map SPM’s library products to a `swift_library` that can be added to other target’s `deps`.

test/fixtures/basic/BUILD
```diff
@@ -9,13 +9,21 @@ licenses(["notice"])
 
 swift_library(
     name = "first",
+    module_name = "first",
```
brentleyjones, line 12: https://github.com/bazelbuild/rules_swift/pull/1154

swift/internal/swift_library.bzl
```diff
@@ -301,7 +307,6 @@ dependent for linking, but artifacts/flags required for compilation (such as
     doc = """\
 Compiles and links Swift code into a static library and Swift module.
 """,
-    outputs = swift_library_output_map,
```
brentleyjones, line 304: This is needed in order to support not having an archive output. Currently it breaks `//examples/xplatform/swift_import:import`. I'm looking into how to fix that.

brentleyjones: Resolved by exposing a new output_group and using `filegroup`. This is technically a breaking change, but a small one, so I think just calling it out in the release notes should be enough.

swift/internal/linking.bzl
```diff
@@ -466,6 +494,47 @@ def new_objc_provider(
...
+    if lib.resolved_symlink_interface_library != None:
+        dynamic_library.append(lib.resolved_symlink_interface_library)
+    elif lib.interface_library != None:
+        dynamic_library.append(lib.interface_library)
+
+    output_groups["archive"] = depset(archive_file)
+    output_groups["dynamic_library"] = depset(dynamic_library)
+
+    return output_groups
+
```
brentleyjones, line 537: Copied how `cc_library` does it: https://github.com/bazelbuild/bazel/blob/ff36d875b9b236ad141dce40e65cae5f4ffbfdcb/src/main/starlark/builtins_bzl/common/cc/cc_library.bzl#L326-L351

cgrindel (approved):

luispadron (approved):

keith:
previous convo https://github.com/bazelbuild/rules_swift/pull/429

cgrindel:
> previous convo https://github.com/bazelbuild/rules_swift/pull/429

From my quick read, it seems like most of the concern is the source-less `swift_library`. Instead of overloading `swift_library`, it could be a new rule `swift_product`. Via the providers mechanism, we could have it only accept `SwiftInfo`, `CcInfo`, etc.

luispadron:
I don't have super strong feelings about a new rule vs. using `swift_library`. If we go with a new rule it's going to be different than the other `*c_library` rules which seems a bit confusing. 

If the only concern is that some folks might spend a bit more analysis time on empty `swift_library` targets, then we should maybe just make the rules align here and use `swift_library`. We can mention the performance concerns in the documentation.

brentleyjones:
I don't agree with the previous convo. I also feel it should be `swift_library` and not a new rule, to match the other library rules. But if a new rule is what's needed to land this, that's fine as well (and the new rule will never produce a swiftmodule, it won't have a `srcs`, etc.).

keith:
no objections one way or the other just wanted to provide previous conext

cgrindel:
For the record, I am fine with `swift_library` doing double duty. Just responding to the previous conversation.

brentleyjones:
I will concede that this is a slight (super super slight) analysis regression for all `swift_library` because of the additional conditionals. It also loses the ability to have the implicit output.

Maybe that is enough to make this a new rule. `cc_library` also has the "header only library" thing going for it that we don't have. I also feel that if `cc_library` was designed today, they would use two rules for this, so that a library always produced a `.a`.

Okay, I talked myself the other way. I'll make a new rule, and the docs for `swift_library` can point to it for people looking to do source-less libraries.

brentleyjones:
Now to bikeshed a name. `swift_library_group`?

cgrindel:
I am ambivalent on the name. If you want to be Swift friendly, `swift_product` would align nicely with SPM's concept of a product.

brentleyjones:
I don't like that name since `swift_binary` and `swift_test` are products too. SPM library products are an odd beast and represent a grouping. So trying to find something that matches more bazel naming.

cgrindel:
Swift_library_group works for me.

luispadron:
That matches well with `filegroup` so makes sense to me.

brentleyjones:
Not exactly. That creates a new module, that you need to import. The point of this new rule is to have something that doesn't create a swiftmodule, but just passes along the merged `SwiftInfo`.

brentleyjones:
Changed the PR to add a new `swift_library_group` rule.

doc/rules.md
```diff
@@ -384,6 +385,26 @@ Compiles and links Swift code into a static library and Swift module.
...
+<a id="swift_library_group"></a>
+
+## swift_library_group
+
+<pre>
+swift_library_group(<a href="#swift_library_group-name">name</a>, <a href="#swift_library_group-deps">deps</a>)
+</pre>
+
+Groups Swift compatible libraries (e.g. `swift_library` and `objc_library`).
+Can be used anywhere a `swift_library` can be used.
```
luispadron, line 397: Could we add some info that this still means you need to `import` the individual libraries using their own module name.

luispadron (approved):

adincebic (approved):

PR #1161, opened 2024-01-24
Enable support for clang modules from cc_proto_library targets

AttilaTheFun:
Right now the clang module aspect hardcodes the "cc_library" rule kind, meaning it does not work with "cc_proto_library" targets. I'm interested in using Swift / C++ interop to share protos with our xplat C++ layer. Ideally this shouldn't hardcode any rule kinds, but at least for now this will unblock us.


keith (commented):
looks good, would be nice to have a test if possible, i think you need a rebase

AttilaTheFun:
@keith unfortunately the C++ protos don't really work through the Swift / C++ interop. I think this is because the base Message interface has some virtual methods which are not supported right now. I'll close this for now, but revisit with Swift 6 if the situation improves.

PR #1163, opened 2024-01-27
chore: format all Starlark files

cgrindel:
Ran `buildifier -lint fix -mode fix -r .` on the repository using Buildifier 6.4.0.


luispadron (approved):
I'm surprised there's no buildifier target like what's in rules_apple. Thanks!

PR #1167, opened 2024-02-13
Update bazel_skylib dependency

AttilaTheFun:
rules_swift currently depends on bazel_skylib 1.3.0 which is 1.5 years old. This PR updates this dependency to 1.5.0.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,7 +7,7 @@ module(
 )
 
 bazel_dep(name = "bazel_features", version = "1.3.0")
-bazel_dep(name = "bazel_skylib", version = "1.3.0")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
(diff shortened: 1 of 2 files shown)


luispadron (approved):

mattrobmattrob:
Did you need some functionality from the new version in this repo?

AttilaTheFun:
@mattrobmattrob not specifically, no, but I was struggling with some WORKSPACE dependency issues related to the old version earlier today. I did eventually manage to work around them, but I think it's generally a good idea to keep dependencies up to date (and pick up bug fixes while we're at it).

brentleyjones:
> but I think it's generally a good idea to keep dependencies up to date (and pick up bug fixes while we're at it)

This is true for `WORKSPACE` macros, but it's best practice for Bazel modules to depend on the lowest version to allow for the greatest chance of dependency resolution succeeding. Root modules can depend on a newer version if they need/want to.

PR #1170, opened 2024-02-19
Add a mechanism to provide a list of protocol names for constant value extraction.

BalestraPatrick:
This is an attempt at cherry-picking support for producing const values produced via the compiler flags `-emit-const-values` and `-const-gather-protocols-file`. It's still a WIP as it's not completely working at the moment.


swift/internal/compiling.bzl
```diff
@@ -3487,7 +3576,58 @@ def _emitted_output_nature(feature_configuration, user_compile_flags):
...
+    const_gather_protocols = []
+    for swift_info in swift_infos:
+        for module_context in swift_info.direct_modules:
+            const_gather_protocols.extend(
+                module_context.const_gather_protocols,
+            )
+
+    # TODO: remove hack, I picked these from Xcode.
+    # Not sure how rules_apple should pass these.
+    const_gather_protocols = [
```
BalestraPatrick, line 3604: I'm not really sure how we're supposed to pass this list of protocols. I took it from a compiler invocation of a simple Xcode project, but I'm open to suggestions on how to handle this properly. In the upstream [commit description](https://github.com/bazelbuild/rules_swift/pull/1170/commits/fa0fb7aa1f049439e91823e122229eda220b9399) I couldn't find more information about it.

keith: Looks like it's hardcoded in Xcode as:

```
                Name = "SWIFT_EMIT_CONST_VALUE_PROTOCOLS";
                Type = StringList;
                DefaultValue = "AppIntent EntityQuery AppEntity TransientEntity AppEnum AppShortcutProviding AppShortcutsProvider AnyResolverProviding AppIntentsPackage DynamicOptionsProvider";
```

So we'll probably have to duplicate it similar to this, seems fine to me

keith:
what's the general progress on this?

BalestraPatrick:
I just verified this seems to be working in our project together with https://github.com/bazelbuild/rules_apple/pull/2418

swift/internal/compiling.bzl
```diff
@@ -3038,6 +3086,9 @@ def _declare_compile_outputs(
             src = srcs[0],
         )]
         other_outputs = []
+        const_values_files = [
```
keith, line 3089: how much of this is from upstream? i don't see offhand why this is an array as opposed to a single file like some of the others, do you?

swift/internal/compiling.bzl
```diff
@@ -3038,6 +3086,9 @@ def _declare_compile_outputs(
...
+            actions.declare_file("{}.swiftconstvalues".format(target_name)),
```
keith, line 3090: seems fine but maybe should be in derived_files? (maybe that's gone from upstream instead)

swift/internal/compiling.bzl
```diff
@@ -3487,7 +3575,58 @@ def _emitted_output_nature(feature_configuration, user_compile_flags):
...
+    ]
+
+    # If there are no protocols to extract, return early.
+    if not const_gather_protocols:
+        return None
+
+    # Create the input file to the compiler, which contains a JSON array of
+    # protocol names.
+    const_gather_protocols_file = actions.declare_file(
+        "{}_const_extract_protocols.json".format(target_name),
```
keith, line 3623: seems like having to generate this for every single module is wasteful, but maybe not enough to care? like I guess this file could be part of the toolchain or something instead, but maybe not worth optimizing

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -637,6 +638,9 @@ def _xcode_swift_toolchain_impl(ctx):
     if _is_xcode_at_least_version(xcode_config, "15.0"):
         requested_features.append(SWIFT_FEATURE__SUPPORTS_MACROS)
 
+    if _is_xcode_at_least_version(xcode_config, "15.3"):
```
keith, line 641: I know I pushed this but if it's supported earlier maybe we should make this supported earlier too? especially if that impacts our ability to test this feature on CI with Xcode 15.2

keith (commented):
were there any tests for this we could grab?

swift/internal/compiling.bzl
brentleyjones: I agree that we shouldn't create it per target.

swift/internal/compiling.bzl
BalestraPatrick: Isn't it because there's one file per source file if it's not using wmo (specified in the for loop below [here](https://github.com/bazelbuild/rules_swift/pull/1170/files#diff-535d0dbea9614c6905a30dfdb0e708df3b16fc68f8930d6aba519342509dc657R3260-R3270))?

swift/internal/compiling.bzl
BalestraPatrick: If you meant to suggest moving this logic to `derived_files.bzl` next to the other functions declare file outputs, I've done it in https://github.com/bazelbuild/rules_swift/pull/1170/commits/046d05b5a454d510f0211d36809df26d81a47d38

swift/internal/compiling.bzl
BalestraPatrick: Moved it to a static file in the toolchain with https://github.com/bazelbuild/rules_swift/pull/1170/commits/b0e3480f7afe44f26ac04a1a207e492d72f7174f

BalestraPatrick:
@keith There were no tests upstream, but I tried my best at adding some. I also had to disable them on Linux due to the fact that we're using Swift 5.7 there, and from what I can see support for the flags we're using started in Swift 5.8.

BalestraPatrick:
When testing this internally with a remote cache I'm getting `Expected output Foo.swiftconstvalues was not created locally.`, so still need to investigate that.

swift/internal/feature_names.bzl
```diff
@@ -357,3 +357,8 @@ SWIFT_FEATURE_EMIT_SWIFTSOURCEINFO = "swift.emit_swiftsourceinfo"
 # A feature that adds target_name in output path to support building
 # multiple frameworks with different target name, but same module name.
 SWIFT_FEATURE_ADD_TARGET_NAME_TO_OUTPUT = "swift.add_target_name_to_output"
+
+# A private feature that is set by the toolchain if it supports constant values
+# extraction. Users should never manually enable, disable, or query this
+# feature.
```
luispadron, line 363: Could we add some info here about what Swift versions support this

swift/internal/providers.bzl
```diff
@@ -341,6 +346,9 @@ def create_module(
             contains artifacts related to Clang modules, such as a module map or
             precompiled module. This may be `None` if the module is a pure Swift
             module with no generated Objective-C interface.
+        const_gather_protocols: A list of protocol names from which constant
```
luispadron, line 349: nit: this attribute reads more like a `bool` than a list, what about `const_protocols_to_gather` or `const_protocols`

swift/internal/swift_toolchain.bzl
```diff
@@ -483,6 +484,14 @@ for incremental compilation using a persistent mode.
 """,
                 executable = True,
             ),
+            "const_gather_protocols": attr.label(
+                default = Label("@build_bazel_rules_swift//swift/toolchains/config:const_extract_protocols.json"),
```
luispadron, line 488: should the attr here and json file name match?

luispadron (approved):

keith (approved):

swift/internal/providers.bzl
BalestraPatrick: Good idea, I renamed it to `const_protocols_to_gather` pretty much everywhere!

swift/internal/swift_toolchain.bzl
BalestraPatrick: We can do that, changed it to `const_protocols_to_gather.json`

swift/internal/feature_names.bzl
BalestraPatrick: Added, it's Swift 5.8 and above.

PR #1173, opened 2024-02-23
Add swift_proto_library_group rule

AttilaTheFun:
I recently merged https://github.com/bazelbuild/rules_swift/pull/1140 which introduced a new, highly-customizable swift_proto_library implementation. 

However, because this dropped support for the proto aspect, consumers of the deprecated rules who relied on this behavior and don't want to use a tool like gazelle to generate their build files will have a hard time migrating to the new rules.

This PR introduces a new swift_proto_library_group rule which uses an aspect and functions very similarly to the deprecated rule to facilitate their migration. Because of the limitations of aspects, though, they will be unable to benefit from some of the customization options afforded to the new swift_proto_library rule including custom module names. This was true of the deprecated rules and it will be true of the aspect-based rule as well.

To avoid repeated / diverging logic, I extracted most of the logic necessary to generate the swift source files and compile them + aggregate the providers into the swift_proto_utils.bzl file and shared this between the swift_proto_library and swift_proto_library_group rules.

I've also added a working GRPC example which shows how to combine swift_proto_library and swift_proto_library_group targets. As with the old rule, swift_proto_library_group does not support GRPC generation directly. However, you can create a swift_proto_library target to generate your service protos which depends on a swift_proto_library_group which generates everything else. This is how the deprecated rules behaved with a swift_grpc_library depending on a (deprecated) swift_proto_library target.


proto/swift_proto_compiler.bzl
```diff
@@ -244,6 +266,24 @@ Typically, these are Well Known Types and proto runtime libraries.
 """,
             providers = [SwiftInfo],
         ),
+        "bundled_proto_paths": attr.string_list(
+            doc = """\
+List of proto paths for which to skip generation because they're bundled in runtime.
```
luispadron, line 271:
```suggestion
List of proto paths for which to skip generation because they're bundled at runtime.
```

proto/swift_proto_library.bzl
```diff
@@ -25,15 +25,12 @@ load(
     "ProtoInfo",
 )
 load(
-    "//proto:swift_proto_common.bzl",
-    "swift_proto_common",
+    "//proto:swift_proto_utils.bzl",
+    "compile_swift_protos_for_target",
```
luispadron, line 29: Should we keep the file name as it was? Seems like this could break any folks who had started using the existing name

proto/swift_proto_library.bzl
```diff
@@ -82,160 +59,42 @@ def _get_module_name(attr, target_label):
...
-            linking_contexts = [
-                dep[CcInfo].linking_context
-                for dep in deps
-                if CcInfo in dep
-            ],
-            module_context = module_context,
-            swift_toolchain = swift_toolchain,
-        )
+    # Compile the source files to a module:
+    direct_output_group_info, direct_proto_cc_info, direct_swift_info, direct_swift_proto_info = compile_swift_protos_for_target(
```
luispadron, line 69: Should this return a struct vs. a tuple? Might be more clear

proto/swift_proto_library.bzl
luispadron: Actually i see that these two files both export a `swift_proto_common` type, is that expected?

proto/swift_proto_utils.bzl
```diff
@@ -101,6 +127,259 @@ def _render_text_module_mapping(mapping):
...
+            direct = generated_swift_srcs,
+            transitive = [swift_proto_info.pbswift_files for swift_proto_info in transitive_swift_proto_infos],
+        ),
+    )
+
+    return direct_output_group_info, direct_swift_proto_cc_info, direct_swift_info, direct_swift_proto_info
+
 # The exported `swift_proto_common` module, which defines the public API
 # for rules that compile Swift protos.
 swift_proto_common = struct(
```
luispadron, line 385: Should the compile function be added here too?

proto/swift_proto_compiler.bzl
AttilaTheFun: They're actually not bundled *at* runtime, I meant that they're bundled in the SwiftProtobuf module along with the primitive types for protos which I was referring to as the runtime. That said, I can reword this to make it clearer.

proto/swift_proto_library.bzl
AttilaTheFun: Ah no, I forgot to delete the redundant swift_proto_common export. I can do that now.

Re: compile_swift_protos_for_target, my original thinking was that I should not export it because it relied on some implementation details of the swift_proto_compiler rule. 

But I just looked and I could move bundled_proto_paths from an internal field on the provider to a public field and remove that dependency on an implementation detail so I can export it.

proto/swift_proto_utils.bzl
AttilaTheFun: Responded above -- it used to rely on an implementation detail but I was able to remove that and export it.

proto/swift_proto_library.bzl
AttilaTheFun: Sure I can make it a struct

luispadron (approved):

proto/swift_proto_compiler.bzl
```diff
@@ -38,14 +38,14 @@ def _swift_proto_compile(ctx, swift_proto_compiler_info, additional_compiler_inf
...
-        module_mappings: the module_mappings field of the `SwiftProtoInfo` for the `swift_proto_library` target
+        ctx: The context of the aspect or rule.
+        swift_proto_compiler_info: The `SwiftProtoCompilerInfo` provider.
+        additional_compiler_info: Information passed from the library target to the compiler.
+        proto_infos: List of `ProtoInfo` providers to compile.
+        module_mappings: The module_mappings field of the `SwiftProtoInfo` for the library target.
 
     Returns:
-        A list of .swift Files generated by the compiler.
+        A list of .swift files generated by the compiler.
```
brentleyjones, line 48:
```suggestion
        A `list` of `.swift` `File`s generated by the compiler.
```

proto/swift_proto_compiler.bzl
```diff
@@ -38,14 +38,14 @@ def _swift_proto_compile(ctx, swift_proto_compiler_info, additional_compiler_inf
     """Compiles Swift source files from `ProtoInfo` providers.
 
     Args:
-        ctx: the `swift_proto_library` target's context
-        swift_proto_compiler_info: this `SwiftProtoCompilerInfo` provider
-        additional_compiler_info: information passed from the `swift_proto_library` target to the compiler
-        proto_infos: the list of `ProtoInfo` providers to compile
-        module_mappings: the module_mappings field of the `SwiftProtoInfo` for the `swift_proto_library` target
+        ctx: The context of the aspect or rule.
```
brentleyjones, line 41: Same comment.

proto/swift_proto_compiler.bzl
```diff
@@ -220,6 +241,7 @@ def _swift_proto_compiler_impl(ctx):
         SwiftProtoCompilerInfo(
             compile = _swift_proto_compile,
             compiler_deps = ctx.attr.deps,
+            bundled_proto_paths = ctx.attr.bundled_proto_paths,
```
brentleyjones, line 244:
```suggestion
            bundled_proto_paths = ctx.attr.bundled_proto_paths,
            compile = _swift_proto_compile,
            compiler_deps = ctx.attr.deps,
```

proto/swift_proto_compiler.bzl
```diff
@@ -244,6 +266,25 @@ Typically, these are Well Known Types and proto runtime libraries.
 """,
             providers = [SwiftInfo],
         ),
+        "bundled_proto_paths": attr.string_list(
```
brentleyjones, line 269: nit: can you move this up above `deps`?

proto/swift_proto_library.bzl
```diff
@@ -82,161 +65,55 @@ def _get_module_name(attr, target_label):
...
-            direct_pbswift_files = generated_swift_srcs,
-            pbswift_files = depset(
-                direct = generated_swift_srcs,
-                transitive = [dep[SwiftProtoInfo].pbswift_files for dep in deps if SwiftProtoInfo in dep],
-            ),
-        ),
+        direct_output_group_info,
+        direct_swift_proto_cc_info.cc_info,
+        direct_swift_proto_cc_info.objc_info,
+        direct_swift_info,
```
brentleyjones, line 113:
```suggestion
        direct_swift_info,
        direct_swift_proto_cc_info.cc_info,
        direct_swift_proto_cc_info.objc_info,
```

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+        DefaultInfo(
+            files = depset(
+                direct_files,
+                transitive = [direct_swift_proto_info.pbswift_files],
+            ),
+        ),
+        direct_output_group_info,
+        direct_swift_proto_cc_info.cc_info,
+        direct_swift_proto_cc_info.objc_info,
+        direct_swift_info,
```
brentleyjones, line 129:
```suggestion
        direct_swift_info,
        direct_swift_proto_cc_info.cc_info,
        direct_swift_proto_cc_info.objc_info,
```

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+    outputs = ["//proto:_swift_proto_compiler"],
+)
+
+# swift_proto_library_group
+
+def _swift_proto_library_group_impl(ctx):
+    proto_target = ctx.attr.proto
+    direct_output_group_info = proto_target[OutputGroupInfo]
+    direct_swift_proto_cc_info = proto_target[SwiftProtoCcInfo]
+    direct_swift_info = proto_target[SwiftInfo]
```
brentleyjones, line 109:
```suggestion
    direct_swift_info = proto_target[SwiftInfo]
    direct_swift_proto_cc_info = proto_target[SwiftProtoCcInfo]
```

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+        "proto": attr.label(
+            aspects = [_swift_proto_library_group_aspect],
+            doc = """\
+Exactly one `proto_library` target (or target producing `ProtoInfo`),
+from which the Swift source files should be generated.
+""",
+            providers = [ProtoInfo],
+            mandatory = True,
+        ),
+        "compiler": attr.label(
```
brentleyjones, line 144: nit: can you move this above `proto`?

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+    },
+    cfg = _swift_proto_compiler_transition,
+    doc = """\
+Generates a collection of Swift static library from a target producing `ProtoInfo` and its dependencies.
+
+This rule is intended to facilitate migration from the deprecated swift proto library rules to the new ones.
+Unlike swift_proto_library which is a drop-in-replacement for swift_library,
```
brentleyjones, line 161:
```suggestion
Unlike `swift_proto_library` which is a drop-in-replacement for `swift_library`, 
```

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+    cfg = _swift_proto_compiler_transition,
+    doc = """\
+Generates a collection of Swift static library from a target producing `ProtoInfo` and its dependencies.
+
+This rule is intended to facilitate migration from the deprecated swift proto library rules to the new ones.
+Unlike swift_proto_library which is a drop-in-replacement for swift_library, 
+this rule uses an aspect over the direct proto library dependency and its transitive dependencies,
+compiling each into a swift static library.
+
+For example, in the following targets, the proto_library_group_swift_proto target only depends on
```
brentleyjones, line 165:
```suggestion
For example, in the following targets, the `proto_library_group_swift_proto` target only depends on
```

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+package_2_proto target, and this transitively depends on package_1_proto.
```
brentleyjones, line 166:
```suggestion
`package_2_proto` target, and this transitively depends on `package_1_proto`.
```

proto/swift_proto_library_group.bzl
```diff
@@ -0,0 +1,219 @@
...
+When used as a dependency from a swift_library or swift_binary target,
```
brentleyjones, line 167:
```suggestion
When used as a dependency from a `swift_library` or `swift_binary` target, 
```

proto/swift_proto_utils.bzl
```diff
@@ -48,25 +75,24 @@ def proto_path(proto_src, proto_info):
...
     return proto_src.path[len(prefix):]
 
-def register_module_mapping_write_action(label, actions, module_mappings):
+def register_module_mapping_write_action(ctx, module_mappings):
     """Registers an action that generates a module mapping for a proto library.
 
     Args:
-        label: The label of the target being analyzed.
-        actions: The context's actions object.
+        ctx: Context of the aspect or rule.
```
brentleyjones, line 82:
```suggestion
        ctx: Rule context.
```

proto/swift_proto_utils.bzl
```diff
@@ -101,8 +127,264 @@ def _render_text_module_mapping(mapping):
 
     return content
 
-# The exported `swift_proto_common` module, which defines the public API
-# for rules that compile Swift protos.
-swift_proto_common = struct(
-    proto_path = proto_path,
+def _generate_module_mappings(module_name, proto_infos, transitive_swift_proto_deps, bundled_proto_paths):
```
brentleyjones, line 130: I recommend making these all named (starting it with `*,`) and ordering them alphabetically. Same for `compile_swift_protos_for_target`.

proto/swift_proto_utils.bzl
```diff
@@ -101,8 +127,264 @@ def _render_text_module_mapping(mapping):
...
+        pbswift_files = depset(
+            direct = generated_swift_srcs,
+            transitive = [swift_proto_info.pbswift_files for swift_proto_info in transitive_swift_proto_infos],
+        ),
+    )
+
+    return struct(
+        direct_output_group_info = direct_output_group_info,
+        direct_swift_proto_cc_info = direct_swift_proto_cc_info,
+        direct_swift_info = direct_swift_info,
```
brentleyjones, line 388:
```suggestion
        direct_swift_info = direct_swift_info,
        direct_swift_proto_cc_info = direct_swift_proto_cc_info,
```

swift/internal/providers.bzl
```diff
@@ -146,6 +146,10 @@ Typically these are proto runtime libraries.
 
 Well Known Types should be added as dependencies of the swift_proto_library
 targets as needed to avoid compiling them unnecessarily.
+""",
+        "bundled_proto_paths": """\
```
brentleyjones, line 150: Can you move this to be above `compiler_deps`?

examples/xplatform/custom_swift_proto_compiler/rules/custom_swift_proto_compiler.bzl
```diff
@@ -34,14 +34,14 @@ def _custom_swift_proto_compile(ctx, swift_proto_compiler_info, additional_compi
     """Compiles Swift source files from `ProtoInfo` providers.
 
     Args:
-        ctx: the `swift_proto_library` target's context
-        swift_proto_compiler_info: this `SwiftProtoCompilerInfo` provider
-        additional_compiler_info: information passed from the `swift_proto_library` target to the compiler
-        proto_infos: the list of `ProtoInfo` providers to compile
-        module_mappings: the module_mappings field of the `SwiftProtoInfo` for the `swift_proto_library` target
+        ctx: The context of the aspect or rule.
```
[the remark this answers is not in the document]
AttilaTheFun: So I was using the ctx for the label and actions object. I suppose I could just pass those directly.

swift/internal/providers.bzl
AttilaTheFun: Shouldn't it actually be above "compile" if we're going alphabetically?

proto/swift_proto_utils.bzl
AttilaTheFun: I'm not sure what you mean by "I recommend making these all named (starting it with `*,`) and ordering them alphabetically." Are you referring to the arguments? Could you give an example?

brentleyjones (approved):

rasberik:
@AttilaTheFun @brentleyjones In case I am missing something, is there way to apply "additional_compiler_info" to this rule?
New proto rules dont seem to respect `swift.generate_path_to_underscores_from_proto_files` on `proto_library`, and there is no option like 
```
additional_compiler_info = {
        "FileNaming": "PathToUnderscores",
    },
``` 

as with new proto library rule. I was migrating from old proto and due to large scale, tried to use `swift_proto_library_group` but cannot resolve duplicate issues.

AttilaTheFun:
@rasberik the swift_proto_library_group doesn't support that attribute yet, but you can customize the swift_proto_compiler through the compiler attribute here:
https://github.com/bazelbuild/rules_swift/blob/master/proto/swift_proto_library_group.bzl#L141

On your custom swift_proto_compiler target you can customize the options, so you can accomplish this that way.

You're welcome to create a feature request or PR to add the additional_compiler_info attribute to swift_proto_library_group and plumb it into the aspect through the transition similar to the compiler attribute. I think this would be fairly simple?

cc @brentleyjones

PR #1177, opened 2024-03-07
Fix docs build for proto compiler

luispadron:
This was causing `bazel test //...` to fail since `//swift/proto` is not a package

### DIFF
--- a/proto/compilers/BUILD
+++ b/proto/compilers/BUILD
@@ -51,7 +51,7 @@ bzl_library(
     srcs = ["swift_proto_compiler_macros.bzl"],
     visibility = ["//visibility:public"],
     deps = [
-        "//swift/proto:swift_proto_compiler",
+        "//proto:swift_proto_compiler",
         "@bazel_skylib//lib:dicts",
     ],
 )


luispadron:
Not sure what's up with the ubuntu tests

doc/rules.md
```diff
@@ -664,7 +664,7 @@ swift_proto_library(
...
 | <a id="swift_proto_library-alwayslink"></a>alwayslink |  If true, any binary that depends (directly or indirectly) on this Swift module will link in all the object files for the files listed in `srcs`, even if some contain no symbols referenced by the binary. This is useful if your code isn't explicitly called by code in the binary; for example, if you rely on runtime checks for protocol conformances added in extensions in the library but do not directly reference any other symbols in the object file that adds that conformance.   | Boolean | optional |  `False`  |
-| <a id="swift_proto_library-compilers"></a>compilers |  One or more `swift_proto_compiler` targets (or targets producing `SwiftProtoCompilerInfo`), from which the Swift protos will be generated.   | <a href="https://bazel.build/concepts/labels">List of labels</a> | optional |  `["@build_bazel_rules_swift//proto/compilers:swift_proto"]`  |
```
brentleyjones, line 667: IMO this change shouldn't have happened

luispadron: Yeah i think this changed because it was failing to build before (not sure how it made it through PR though). `//swift/proto` doesnt exist.

brentleyjones: The tests are failing because it wants `build_bazel_rules_swift` instead of `rules_swift`. I think this is something to do with using bzlmod or not when generating the docs.

luispadron: yep that was it, ran without bzlmod and its fixed now thanks!

brentleyjones (approved):

PR #1179, opened 2024-03-18
Feature to opt out of tracking .swiftdoc and .swiftsourceinfo Files

chiragramani:
This PR introduces configuration options aimed at optimizing CI build processes by excluding non-essential Swift documentation and source info files, files that don't serve as inputs to other build targets. By introducing flags to control the exclusion of these files, we aim to reduce unnecessary network I/O and improve build performance in specific CI environments. (non-dev facing only)

Features Introduced:

- `swift.emit_swiftdoc`
    - This feature is enabled by default. 
    - Opting out of this feature will prevent .swiftdoc files from being tracked by `SwiftInfo` provider, though they will still be generated by the Swift compiler. 
    - These documentation files are generally used for IDE features and are not required in certain CI contexts, such as non-developer-facing validation suites.

- `swift.emit_swiftsourceinfo`
  - This feature is enabled by default. 
  - When this feature is disabled, .swiftsourceinfo files, which support source-level debugging, will also be excluded from build outputs and will not be tracked by `SwiftInfo` provider.

It is important to note that, by default, the Swift compiler generates .swiftdoc and .swiftsourceinfo files as part of the output when using the `-emit-module` flag. As of now, this behavior is consistent, given the absence of compiler options in Swift that permit controlling the generation of these files.


swift/internal/compiling.bzl
```diff
@@ -2943,6 +2963,8 @@ def _declare_compile_outputs(
         user_compile_flags: The flags that will be passed to the compile action,
             which are scanned to determine whether a single frontend invocation
             will be used or not.
+        include_swiftdoc: If .swiftdoc file should be included or not.
```
brentleyjones, line 2966: Can you move these up to be alphabetical? Same for any other additions.

swift/internal/feature_names.bzl
```diff
@@ -343,3 +343,9 @@ SWIFT_FEATURE__SUPPORTS_MACROS = "swift._supports_macros"
 
 # Pass -warnings-as-errors to the compiler.
 SWIFT_FEATURE_TREAT_WARNINGS_AS_ERRORS = "swift.treat_warnings_as_errors"
+
+# Disable the inclusion of .swiftdoc files in Swift build outputs.
+SWIFT_FEATURE_EXCLUDE_SWIFTDOC = "swift.exclude_swiftdoc"
```
brentleyjones, line 348: I think it's better to have positive flags. Could this be `swift.emit_swift_doc`, enabled by default, and the comment updated (especially mention that the file will still be generated because there is currently no way to tell the compiler to not generated, but we won't track it will Bazel)?

Same for the other flag.

chiragramani: Yes, that sounds better, updated! thanks for pointing it out!

swift/internal/swift_library.bzl
```diff
@@ -120,7 +122,10 @@ def _swift_library_impl(ctx):
         copts.extend(["-static"])
     copts.extend(module_copts)
 
-    extra_features = []
+    extra_features = [
```
brentleyjones, line 125: Instead of here, we should enable it in the toolchains.

chiragramani: got it, updated!

brentleyjones:
Can you also update the PR description with the changes to names and behavior? It will be used for the body of the merged commit.

chiragramani:
> Can you also update the PR description with the changes to names and behavior? It will be used for the body of the merged commit.

Oh yes, updated!

brentleyjones (approved):

luispadron (approved):

brentleyjones:
Ohh, also, do we need to adjust anything in the worker for incremental generation?

jszumski:
> we aim to reduce unnecessary network I/O and improve build performance in specific CI environments
@chiragramani what was the rough magnitude of the performance win?

chiragramani:
> Ohh, also, do we need to adjust anything in the worker for incremental generation?

I tried out the batch mode feature, and the builds worked fine! But I might be missing some scenarios. Can you suggest other tests for incremental generation? I'd like to check those out too.

> > we aim to reduce unnecessary network I/O and improve build performance in specific CI environments
> 
> @chiragramani what was the rough magnitude of the performance win?

I tested this for a smaller set to see the impact when there are no cache-hits.
Without excludes: ~3.6GB uploaded, ~590 seconds.
With excludes: ~3.2Gb uploaded, ~550 seconds.

Also, we are not using `experimental_remote_cache_async` which might explain some of the build time delta (issues with HTTP cache) but maybe this delta would be a lot smaller/negligible for environments making use of `experimental_remote_cache_async`.

brentleyjones:
> Can you suggest other tests for incremental generation? I'd like to check those out too.

There is the `incremental_inputs` stuff in `output_file_map.cc`, since since we are always generating these files, it probably doesn't matter. Thanks for checking.

PR #1180, opened 2024-03-19
Replace instances of to_json() method with json.encode(..)

c-mita:
The to_json and to_proto methods on struct are deprecated and will be
removed. The "json" and "proto" builtin modules should be used instead.

This replaces instances of `foo.to_json()` with `json.encode(foo)`.


brentleyjones (approved):

brentleyjones:
Thanks @c-mita!

PR #1193, opened 2024-04-06
Delete deprecated swift proto library rules

AttilaTheFun:
With the addition of `swift_proto_library_group`, consumers of the deprecated `swift_proto_library` and `swift_grpc_library` rules now have a clear migration path to the new rules. Given that we will be cutting a source-breaking rules_swift version 2.0 soon, this is a good opportunity to delete the old rules. 

We will keep the deprecated rules indefinitely in 1.x, and only the new rules will exist in 2.x.
The release notes will also be useful to make consumers aware of the new rules who might not have used the old ones previously due to their inflexibility.


luispadron (approved):

brentleyjones:
There are some conversations happening in Slack related to how easy it is to migrate. I want those to resolve before I approve this.

mattrobmattrob (approved):
AFAIK we are in a good place to merge this from the Slack conversations. It would be nice to use an RC approach to get more general feedback to those not following in Slack for the next release given the substantial removal. 👍 

Thanks for getting this all through, @AttilaTheFun!

Issue #1204, opened 2024-05-16
XCode SDK's aren't visible in Sandboxed mode


ed-irl:
Hello, I'm interested in using sandboxed mode to achieve something like strict deps, as mentioned in #344 .

With sandboxed mode enabled, however, SDK's provided by XCode (eg. SwiftUI) are unresolvable. I am not sure how I would go about adding bazel targets for these SDK dependencies.

Is there a convenient way to resolve them in sandboxed mode?

Thank you!

keith:
Can you provide a repro case? I don't fully understand what you mean

brentleyjones:
The SDKs not being available is probably that the macros don't work, since they are sandboxed themselves by default: https://github.com/bazelbuild/rules_swift/issues/1202

We need to add the new no-sandbox flag to the swift compiles when it's available (which is Swift 5.10+).

adincebic:
@ed-irl this is fixed.

Issue #1202, opened 2024-05-08
Unable to compile Swift files with system-provided Swift Macro like #Preview when sandboxing is enabled


adincebic:
When performing `bazel build` while `--spawn_strategy=sandboxed` the build fails.

```
➜  bazel-ios-swiftui-template git:(failing-build-with-macros-and-sandboxing) bazel build //app
INFO: Analyzed target //app:app (0 packages loaded, 0 targets configured).
ERROR: /Users/adincebic/developer/bazel-ios-swiftui-template/app/BUILD.bazel:16:14: Compiling Swift module //app:app.library failed: (Exit 1): worker failed: error executing SwiftCompile command (from target //app:app.library) bazel-out/darwin_arm64-opt-exec-ST-13d3ddad9198/bin/external/rules_swift~/tools/worker/worker swiftc ... (remaining 1 argument skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
sandbox-exec: sandbox_apply: Operation not permitted
<unknown>:0: warning: compiler plugin not loaded: '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin/swift-plugin-server; failed to initialize
app/source/ContentView.swift:15:1: error: external macro implementation type 'PreviewsMacros.SwiftUIView' could not be found for macro 'Preview(_:body:)'
#Preview {
^
SwiftUI.Preview:2:41: note: 'Preview(_:body:)' declared here
@freestanding(declaration) public macro Preview(_ name: String? = nil, body: @escaping @MainActor () -> any View) = #externalMacro(module: "PreviewsMacros", type: "SwiftUIView")
                                        ^
error: emit-module command failed with exit code 1 (use -v to see invocation)
sandbox-exec: sandbox_apply: Operation not permitted
<unknown>:0: warning: compiler plugin not loaded: '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin/swift-plugin-server; failed to initialize
app/source/ContentView.swift:15:1: error: external macro implementation type 'PreviewsMacros.SwiftUIView' could not be found for macro 'Preview(_:body:)'
#Preview {
^
SwiftUI.Preview:2:41: note: 'Preview(_:body:)' declared here
[...]

PR #1206, opened 2024-05-22
Disable Swift compiler sandboxing in Xcode 15.3+ to fix nested sandboxing

adincebic:
This should solve https://github.com/bazelbuild/rules_swift/issues/1202
 and #1204

> There were issues with Swift compiler plugins (incl. macros) and nested sandboxes on macOS with Swift 5.9 that have been fixed with 5.10:
> swift: [apple/swift#70079](https://github.com/apple/swift/pull/70079)
> swift-driver: [apple/swift-driver#1493](https://github.com/apple/swift-driver/pull/1493)
> swift-package-manager: [apple/swift-package-manager#7167](https://github.com/apple/swift-package-manager/pull/7167)

I am not sure whether this flag is required on Linux too.

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -341,6 +341,10 @@ SWIFT_FEATURE__FORCE_ALWAYSLINK_TRUE = "swift._force_alwayslink_true"
 # feature.
 SWIFT_FEATURE__SUPPORTS_MACROS = "swift._supports_macros"
 
+# Disables Swift sandbox which prevents issues with nested sandboxing when Swift code contains system-provided macros.
+# If enabled '#Preview' macro provided by SwiftUI fails to build and probably other system-provided macros.
+SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX = "swift._disable_swift_sandbox"
+
 # Pass -warnings-as-errors to the compiler.
 SWIFT_FEATURE_TREAT_WARNINGS_AS_ERRORS = "swift.treat_warnings_as_errors"
 
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -343,6 +343,7 @@ SWIFT_FEATURE__SUPPORTS_MACROS = "swift._supports_macros"
 
 # Disables Swift sandbox which prevents issues with nested sandboxing when Swift code contains system-provided macros.
 # If enabled '#Preview' macro provided by SwiftUI fails to build and probably other system-provided macros.
+# Enabled by default for Swift 5.10+ on macOS.
 SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX = "swift._disable_swift_sandbox"
 
 # Pass -warnings-as-errors to the compiler.
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -344,7 +344,7 @@ SWIFT_FEATURE__SUPPORTS_MACROS = "swift._supports_macros"
 # Disables Swift sandbox which prevents issues with nested sandboxing when Swift code contains system-provided macros.
 # If enabled '#Preview' macro provided by SwiftUI fails to build and probably other system-provided macros.
 # Enabled by default for Swift 5.10+ on macOS.
-SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX = "swift._disable_swift_sandbox"
+SWIFT_FEATURE_DISABLE_SWIFT_SANDBOX = "swift.disable_swift_sandbox"
 
 # Pass -warnings-as-errors to the compiler.
 SWIFT_FEATURE_TREAT_WARNINGS_AS_ERRORS = "swift.treat_warnings_as_errors"
--- a/swift/internal/xcode_swift_toolchain.bzl
+++ b/swift/internal/xcode_swift_toolchain.bzl
@@ -50,6 +50,7 @@ load(
     "SWIFT_FEATURE_SUPPORTS_SYSTEM_MODULE_FLAG",
     "SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE",
     "SWIFT_FEATURE_USE_RESPONSE_FILES",
+    "SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX",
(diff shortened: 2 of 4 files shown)


swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -569,6 +569,10 @@ def _xcode_swift_toolchain_impl(ctx):
     else:
         swiftcopts.extend(ctx.attr._copts[BuildSettingInfo].value)
 
+    # We pass '-disable-sandbox' to allow compiling targets with Swift compiler plugins in Bazel sandbox on Xcode 15.3+
+    if _is_xcode_at_least_version(xcode_config, "15.3"):
+        swiftcopts.append("-disable-sandbox")
```
luispadron, line 574: Should this only be added to the `swift_compiler_plugin`?

luispadron (approved):

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -569,6 +569,10 @@ def _xcode_swift_toolchain_impl(ctx):
     else:
         swiftcopts.extend(ctx.attr._copts[BuildSettingInfo].value)
 
+    # We pass '-disable-sandbox' to allow compiling targets with Swift compiler plugins in Bazel sandbox on Xcode 15.3+
```
luispadron, line 572: Maybe some more info on the issue it causes / links to the swift issues, from JP:

> There were issues with Swift compiler plugins (incl. macros) and nested sandboxes on macOS with Swift 5.9 that have been fixed with 5.10:
> swift: https://github.com/apple/swift/pull/70079
> swift-driver: https://github.com/apple/swift-driver/pull/1493
> swift-package-manager: https://github.com/apple/swift-package-manager/pull/7167

swift/internal/xcode_swift_toolchain.bzl
brentleyjones, line 574: I think instead of setting the foot directly here, we should instead implicitly enable a feature that sets it (there should be other examples of that in the toolchain as well). That way people can opt out of it if they want for some reason. Then we can use a configurator to add the flag in the right order (right now this is being applied after user defined swiftcopts, which is wrong).

brentleyjones (changes requested):

luispadron:
Is it possible to add a test case as well for this, would catch regressions in the future

swift/internal/feature_names.bzl
```diff
@@ -341,6 +341,10 @@ SWIFT_FEATURE__FORCE_ALWAYSLINK_TRUE = "swift._force_alwayslink_true"
 # feature.
 SWIFT_FEATURE__SUPPORTS_MACROS = "swift._supports_macros"
 
+# Disables Swift sandbox which prevents issues with nested sandboxing when Swift code contains system-provided macros.
+# If enabled '#Preview' macro provided by SwiftUI fails to build and probably other system-provided macros.
```
brentleyjones, line 345:
```suggestion
# Disables Swift sandbox which prevents issues with nested sandboxing when Swift code contains system-provided macros.
# If enabled '#Preview' macro provided by SwiftUI fails to build and probably other system-provided macros.
# Enabled by default for Swift 5.10+ on macOS.
```

swift/internal/feature_names.bzl
```diff
@@ -341,6 +341,10 @@ SWIFT_FEATURE__FORCE_ALWAYSLINK_TRUE = "swift._force_alwayslink_true"
...
+SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX = "swift._disable_swift_sandbox"
```
brentleyjones, line 346:
```suggestion
SWIFT_FEATURE_DISABLE_SWIFT_SANDBOX = "swift.disable_swift_sandbox"
```

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -50,6 +50,7 @@ load(
     "SWIFT_FEATURE_SUPPORTS_SYSTEM_MODULE_FLAG",
     "SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE",
     "SWIFT_FEATURE_USE_RESPONSE_FILES",
+    "SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX",
```
brentleyjones, line 53: Once renamed to remove the `_`, please move it up.

swift/internal/xcode_swift_toolchain.bzl
```diff
@@ -640,6 +641,9 @@ def _xcode_swift_toolchain_impl(ctx):
         requested_features.append(SWIFT_FEATURE__SUPPORTS_MACROS)
         requested_features.append(SWIFT_FEATURE__SUPPORTS_CONST_VALUE_EXTRACTION)
 
+    if _is_xcode_at_least_version(xcode_config, "15.3"):
+        requested_features.append(SWIFT_FEATURE__DISABLE_SWIFT_SANDBOX)
```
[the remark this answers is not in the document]
brentleyjones: Yes. They would do `--features=-swift.disable_swift_sandbox`.

brentleyjones (changes requested):

luispadron (approved):
Thanks for fixing up and addressing comments!

keith:
Is this fixed in 15.4?

luispadron:
@keith tested in the reproducer @adincebic provided but looks like its still broken unless we disable_sandbox:


```
<unknown>:0: warning: compiler plugin not loaded: '/Applications/Xcode-15.4.0.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin/swift-plugin-server; failed to initialize
app/source/ContentView.swift:15:1: error: external macro implementation type 'PreviewsMacros.SwiftUIView' could not be found for macro 'Preview(_:body:)'
#Preview {
^
SwiftUI.Preview:2:41: note: 'Preview(_:body:)' declared here
@freestanding(declaration) public macro Preview(_ name: String? = nil, body: @escaping @MainActor () -> any View) = #externalMacro(module: "PreviewsMacros", type: "SwiftUIView")
    
```

brentleyjones (approved):

PhilCai1993:
We use -load-plugin-library instead of executable, it works well, without the sandboxing issue

PR #1207, opened 2024-05-22
Infer emit-bc feature if -lto is passed

mjburghard:
When passing `-lto=llvm-thin` or `-lto=llvm-full` to swiftc, it switches from emitting object files to bitcode files.
To include `llvm-bc` instead of `object` in the generated output-file-map, the feature `SWIFT_FEATURE_EMIT_BC` (`swift.emit_bc`) must be inferred. 
This PR implements this similar to `SWIFT_FEATURE__WMO_IN_SWIFTCOPTS`.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -132,6 +132,13 @@ _WMO_FLAGS = {
     "-wmo": True,
 }
 
+# Swift command line flags that enable LTO. (This
+# dictionary is used as a set for quick lookup; the values are irrelevant.)
+_LTO_FLAGS = {
+    "-lto=llvm-thin": True,
+    "-lto=llvm-full": True,
+}
+
 def compile_action_configs(
         *,
         os = None,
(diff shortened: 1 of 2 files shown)


swift/internal/compiling.bzl
```diff
@@ -132,6 +132,13 @@ _WMO_FLAGS = {
     "-wmo": True,
 }
 
+# Swift command line flags that enable LTO. (This
+# dictionary is used as a set for quick lookup; the values are irrelevant.)
+_LTO_FLAGS = {
+    "-lto=llvm-thin": True,
+    "-lto=llvm-full": True,
```
brentleyjones, line 139:
```suggestion
    "-lto=llvm-thin": None,
    "-lto=llvm-full": None,
```

mjburghard: @brentleyjones Should I update `_WMO_FLAGS` above as well?

brentleyjones: Ideally we would, but then we have a chance at merge conflicts. Let's ignore my comment for now since we have an existing pattern of using non-`None`.

keith:
i think probably instead of doing this we should add new features for LTO which add these flags automatically, as well as the required feature?

mjburghard:
@keith Something like `swift.thin_lto` and `swift.full_lto`? I found the feature for the [unix_cc_toolchain](https://github.com/bazelbuild/bazel/blob/373ea03ff072315cf558d6c7b478fb5495b03fe1/tools/cpp/unix_cc_toolchain_config.bzl#L1286). I may need a couple of pointers, but can look into this.

keith:
yea that's what im thinking. generally we use new features for this stuff vs checking in on the command line args. at least one reason why that's nicer is command line arguments can be negated / overwritten and we'd probably need to handle that

mjburghard:
First draft is in https://github.com/bazelbuild/rules_swift/pull/1208

PR #1208, opened 2024-05-22
Add features to enable LTO

mjburghard:
Add features to enable LTO (thin and full).
When passing `-lto=`, `swiftc` switches to emit bc files and the generated output-file-map must contain a mapping for `llvm-bc` instead of `object`.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -262,11 +262,7 @@ def compile_action_configs(
             configurators = [
                 swift_toolchain_config.add_arg("-emit-bc"),
             ],
-            features = [
-                [SWIFT_FEATURE_EMIT_BC],
-                [SWIFT_FEATURE_FULL_LTO],
-                [SWIFT_FEATURE_THIN_LTO],
-            ],
+            features = [SWIFT_FEATURE_EMIT_BC],
         ),
 
         # Add the single object file or object file map, whichever is needed.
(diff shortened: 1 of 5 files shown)


swift/internal/compiling.bzl
```diff
@@ -259,7 +261,11 @@ def compile_action_configs(
             configurators = [
                 swift_toolchain_config.add_arg("-emit-bc"),
             ],
-            features = [SWIFT_FEATURE_EMIT_BC],
+            features = [
+                [SWIFT_FEATURE_EMIT_BC],
+                [SWIFT_FEATURE_FULL_LTO],
+                [SWIFT_FEATURE_THIN_LTO],
+            ],
```
mjburghard, line 268: Not sure if this is needed, since the compiler internally changes the behavior and adding the additional features here does not simplify anything.

swift/internal/compiling.bzl
```diff
@@ -399,6 +405,24 @@ def compile_action_configs(
...
         ),
+
+        # Link Time Optimization (LTO)
+        swift_toolchain_config.action_config(
+            actions = [swift_action_names.COMPILE],
+            configurators = [
+                swift_toolchain_config.add_arg("-lto=llvm-thin"),
+            ],
+            features = [SWIFT_FEATURE_THIN_LTO],
+            not_features = [SWIFT_FEATURE_FULL_LTO],
```
keith, line 416: I'm not sure what I would expect to happen if you passed `--features=swift.full_lto --features=swift.thin_lto` but I would probably expect the last one to win as opposed to neither to happen. So I think I'd drop this from both

swift/internal/compiling.bzl
keith: if you don't need to pass -emit-bc to the compiler I would leave this out then

keith (commented):
nice!

keith (approved):

PR #1212, opened 2024-06-11
Add feature to propagate the generated module map

brentleyjones:
Allows `objc_library` to modularly import `swift_library` targets without having to generate a similar module map.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -71,6 +71,7 @@ load(
     "SWIFT_FEATURE_OPT",
     "SWIFT_FEATURE_OPT_USES_OSIZE",
     "SWIFT_FEATURE_OPT_USES_WMO",
+    "SWIFT_FEATURE_PROPAGATE_GENERATED_MODULE_MAP",
     "SWIFT_FEATURE_REWRITE_GENERATED_HEADER",
     "SWIFT_FEATURE_SPLIT_DERIVED_FILES_GENERATION",
     "SWIFT_FEATURE_SUPPORTS_BARE_SLASH_REGEX",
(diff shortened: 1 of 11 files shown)


mattrobmattrob (commented):
Can you add a simple test that exercises this feature and shows what it enables vs. the default?

brentleyjones:
@mattrobmattrob I added an example which fails on CI when the new feature isn't passed.

swift/internal/compiling.bzl
```diff
@@ -2714,6 +2715,17 @@ to use swift_common.compile(include_dev_srch_paths = ...) instead.\
         transitive_modules = transitive_modules,
     )
 
+    includes = []
+    public_hdrs = []
+    if compile_outputs.generated_header_file:
+        public_hdrs = [compile_outputs.generated_header_file]
+        if compile_outputs.generated_module_map_file and is_feature_enabled(
```
mattrobmattrob, line 2722: This is intentionally nested because the module won't be generated without the generated public headed, correct?

swift/internal/feature_names.bzl
```diff
@@ -142,6 +142,11 @@ SWIFT_FEATURE_NO_ASAN_VERSION_CHECK = "swift.no_asan_version_check"
 # the target is not generating a header.
 SWIFT_FEATURE_NO_GENERATED_MODULE_MAP = "swift.no_generated_module_map"
 
+# If enabled, the generates module map is named `module.modulemap` and
+# the parent directory is added as to `CcInfo.compliation_context.includes`.
+# This allows `objc_library` to import the Swift module.
```
mattrobmattrob, line 147: This feature just controls the propagation to the `compilation_context` and not the generation. Should that part be removed from the comment?

If so, some typos:
```suggestion
# If enabled, the parent directory of the generated module map is added to
# `CcInfo.compilation_context.includes`. This allows `objc_library` to import
# the Swift module.
```

If not, some typos:
```suggestion
# If enabled, the generated module map is named `module.modulemap` and
# the parent directory is added as to `CcInfo.compilation_context.includes`.
# This allows `objc_library` to import the Swift module.
```

mattrobmattrob (approved):

swift/internal/feature_names.bzl
brentleyjones: Yes. It was as per the comment before, now it's just the propagation now.

swift/internal/compiling.bzl
brentleyjones: Not needed, I'll un-nest it.

PR #1215, opened 2024-06-17
Enable `swift.add_target_name_to_output` by default

brentleyjones:
Since we are releasing 2.0 with breaking changes, we should enable this by default now, and then in 3.0 we can remove the feature.

### DIFF
--- a/test/module_interface_tests.bzl
+++ b/test/module_interface_tests.bzl
@@ -31,10 +31,28 @@ explicit_swift_module_map_test = make_action_command_line_test_rule(
     },
 )
 
+explicit_swift_module_map_without_target_name_test = make_action_command_line_test_rule(
+    config_settings = {
+        "//command_line_option:features": [
+            "swift.use_explicit_swift_module_map",
+            "-swift.add_target_name_to_output",
+        ],
+    },
+)
+
 vfsoverlay_test = make_action_command_line_test_rule(
     config_settings = {
         "//command_line_option:features": [
             "swift.vfsoverlay",
+        ],
+    },
+)
+
+vfsoverlay_without_target_name_test = make_action_command_line_test_rule(
+    config_settings = {
+        "//command_line_option:features": [
+            "swift.vfsoverlay",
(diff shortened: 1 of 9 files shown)


mattrobmattrob (approved):

luispadron (approved):

keith:
Even though we can flip this I don't think this ends up being the right default for folks. It would definitely be nice to remove the implementation complexity but I don't think it's worth deviating from the standard output layout. Especially since this is fixing for something that doesn't appear to be the common case

brentleyjones:
> but I don't think it's worth deviating from the standard output layout

Most rules are written such that the target name is part of the outputs, to prevent this from being an issue. I think the new way is correct while the old way is incorrect.

test/module_interface_tests.bzl
```diff
@@ -69,9 +87,38 @@ def module_interface_test_suite(name):
...
+        ],
+        mnemonic = "SwiftCompileModuleInterface",
+        target_under_test = "@build_bazel_rules_swift//test/fixtures/module_interface:toy_module",
+    )
+
     vfsoverlay_test(
         name = "{}_vfsoverlay_test".format(name),
         tags = [name],
+        expected_argv = [
+            "-vfsoverlay$(BIN_DIR)/test/fixtures/module_interface/ToyModule/ToyModule.vfsoverlay.yaml",
```
brentleyjones, line 107: This should be `-vfsoverlay$(BIN_DIR)/test/fixtures/module_interface/toy_module/ToyModule.vfsoverlay.yaml`. I'll fix it in another PR.

brentleyjones: Similarly, `_explicit_swift_module_map_test` should include `toy_module/`, but it doesn't.

brentleyjones:
```
examples/apple/objc_interop/main.m:20:9: fatal error: 'examples/apple/objc_interop/generated_header/Printer-Swift.h' file not found
#import "examples/apple/objc_interop/generated_header/Printer-Swift.h"
```

Yuck. Not flipping.

PR #1217, opened 2024-06-17
Apply `swift.add_target_name_to_output` only to outputs that need it

brentleyjones:
In particular, only outputs that need to have the module name as part of the filename.

### DIFF
--- a/test/module_interface_tests.bzl
+++ b/test/module_interface_tests.bzl
@@ -60,7 +60,7 @@ def module_interface_test_suite(name):
         name = "{}_explicit_swift_module_map_test".format(name),
         tags = [name],
         expected_argv = [
-            "-explicit-swift-module-map-file $(BIN_DIR)/test/fixtures/module_interface/ToyModule.swift-explicit-module-map.json",
+            "-explicit-swift-module-map-file $(BIN_DIR)/test/fixtures/module_interface/toy_module.swift-explicit-module-map.json",
         ],
         not_expected_argv = [
             "-Xfrontend",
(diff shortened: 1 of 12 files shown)


test/module_interface_tests.bzl
```diff
@@ -60,7 +60,7 @@ def module_interface_test_suite(name):
         name = "{}_explicit_swift_module_map_test".format(name),
         tags = [name],
         expected_argv = [
-            "-explicit-swift-module-map-file $(BIN_DIR)/test/fixtures/module_interface/ToyModule.swift-explicit-module-map.json",
+            "-explicit-swift-module-map-file $(BIN_DIR)/test/fixtures/module_interface/toy_module.swift-explicit-module-map.json",
```
brentleyjones, line 63: One of the two changes when not using the feature.

test/module_interface_tests.bzl
```diff
@@ -73,7 +73,7 @@ def module_interface_test_suite(name):
         name = "{}_vfsoverlay_test".format(name),
         tags = [name],
         expected_argv = [
-            "-vfsoverlay$(BIN_DIR)/test/fixtures/module_interface/ToyModule.vfsoverlay.yaml",
+            "-vfsoverlay$(BIN_DIR)/test/fixtures/module_interface/toy_module.vfsoverlay.yaml",
```
brentleyjones, line 76: One of the two changes when not using the feature.

brentleyjones (commented):
VFS overlay and explicit module map changed the name to be based on the target (since they don't need to be named after the module). The rest of the changes made the flag apply to fewer output paths, since they already included the target name in them.

luispadron (approved):

PR #1219, opened 2024-06-18
Pass `compilation_contexts` and `swift_infos` instead of `deps` to `swift_common.compile`

brentleyjones:
The choice to pass `compilation_contexts` separately was a trade-off. The compilation contexts nested inside the Clang modules aren't easily suitable for use here. We can't rely on only the direct modules' CCs because information that should still get propagated—like defines and include paths—would get lost when passing through any `objc_library` that doesn't export a module. And we don't want to process the full set of transitive modules' CCs for each compile action, as this would be a significant performance hit. The only other option considered was to propagate a copy of the CC at the top level of `SwiftInfo`, even for no-module `objc_library` targets, but there were concerns about performance here as well, and whether that context would drift from the real one.

Even with the extra argument, this change makes the compilation API more flexible by removing the assumption that compilation requirements must be expressed as `Target`s. For example, removing the assumption that dependencies propagate `SwiftInfo` directly allows the API to be used inside aspects where the `SwiftInfo` providers may need to be wrapped by a custom provider in order to avoid collisions.

PiperOrigin-RevId: 424979602
(cherry picked from commit bfcae19af93924651911e6bdc669608e9453a774)

---

Cherry-pick notes: In `swift_common.compile` we take in `cc_infos` and `objc_infos` instead of `compilation_contexts` since we still need merged linking information to support  `-disable-autolink-framework`. If we ever remove that feature, we should change `compile` to match `upstream`.

### DIFF
--- a/proto/swift_proto_utils.bzl
+++ b/proto/swift_proto_utils.bzl
@@ -49,7 +49,7 @@ load(
 )
 
 def proto_path(proto_src, proto_info):
-    """Derives the string used to import the proto. 
+    """Derives the string used to import the proto.
 
     This is the proto source path within its repository,
     adjusted by `import_prefix` and `strip_import_prefix`.
@@ -270,14 +270,16 @@ def compile_swift_protos_for_target(
     include_dev_srch_paths = include_developer_search_paths(attr)
     module_context, cc_compilation_outputs, other_compilation_outputs = swift_common.compile(
         actions = ctx.actions,
+        cc_infos = get_providers(compiler_deps, CcInfo),
         copts = ["-parse-as-library"],
-        deps = compiler_deps,
         feature_configuration = feature_configuration,
         include_dev_srch_paths = include_dev_srch_paths,
         module_name = module_name,
+        objc_infos = get_providers(compiler_deps, apple_common.Objc),
         package_name = None,
         srcs = generated_swift_srcs,
         swift_toolchain = swift_toolchain,
+        swift_infos = get_providers(compiler_deps, SwiftInfo),
         target_name = target_label.name,
         workspace_name = ctx.workspace_name,
     )
--- a/doc/api.md
+++ b/doc/api.md
@@ -86,10 +86,10 @@ A new attribute dictionary that can be added to the attributes of a
 ## swift_common.compile
 
 <pre>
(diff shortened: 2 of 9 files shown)


proto/swift_proto_utils.bzl
```diff
@@ -270,14 +270,16 @@ def compile_swift_protos_for_target(
...
         copts = ["-parse-as-library"],
-        deps = compiler_deps,
         feature_configuration = feature_configuration,
         include_dev_srch_paths = include_dev_srch_paths,
         module_name = module_name,
+        objc_infos = get_providers(compiler_deps, apple_common.Objc),
         package_name = None,
         srcs = generated_swift_srcs,
         swift_toolchain = swift_toolchain,
+        swift_infos = get_providers(compiler_deps, SwiftInfo),
```
brentleyjones, line 282: @AttilaTheFun `upstream` this is
```
    compile_deps = deps + ctx.attr._proto_support
    compile_deps_swift_infos = get_providers(compile_deps, SwiftInfo)

    module_name = swift_common.derive_module_name(ctx.label)

    module_context, compilation_outputs = swift_common.compile(
        actions = ctx.actions,
        compilation_contexts = get_compilation_contexts(ctx.attr.deps),
        copts = ["-parse-as-library"],
        module_name = module_name,
        srcs = generated_files,
        swift_infos = compile_deps_swift_infos,
        swift_toolchain = swift_toolchain,
        target_name = ctx.label.name,
    )
```

just making sure we are doing the right thing here for the new fields.

AttilaTheFun: @brentleyjones not sure I entirely understand the implications here. What happens if compiler deps includes non-swift dependencies? 

The user can pass additional dependencies to the compiler, typically for the proto runtime libraries like SwiftProtobuf and grpc-swift.

Those libraries may not be Swift, though. They could be C/C++/Objective-C libraries instead which are required by the generated protobuf code.

For example, at Snap, we're working on adopting these rules. We'll use SwiftProtobuf out of the box, but we'll need to fork grpc-swift to use our c++ networking service.

brentleyjones: The old code would extra these providers from `deps + private_deps`, so the change I made is at least the same as what was there. Just calling out that `upstream` seems to split it slightly.

luispadron (approved):

PR #1253, opened 2024-06-21
Make it an error to specify both `swiftinterface` and `swiftmodule` in `swift_import`

luispadron:
In rules_swift 1.x specifying both `swiftinterface` and `swiftmodule` would cause the `swiftinterface` attr to be ignored. This was made into a warning then and is being upgraded to a failure now for rules_swift 2.x

### DIFF
--- a/swift/swift_import.bzl
+++ b/swift/swift_import.bzl
@@ -46,8 +46,12 @@ def _swift_import_impl(ctx):
     )
 
     if not (swiftinterface or swiftmodule):
-        fail("One or both of 'swiftinterface' and 'swiftmodule' must be " +
+        fail("One of 'swiftinterface' or 'swiftmodule' must be " +
              "specified.")
+
+    if swiftinterface and swiftmodule:
+        fail("'swiftinterface' may not be specified when " +
+             "'swiftmodule' is specified.")
 
     swift_toolchain = swift_common.get_toolchain(ctx)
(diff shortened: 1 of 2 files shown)


swift/swift_import.bzl
```diff
@@ -97,9 +101,6 @@ def _swift_import_impl(ctx):
             module_context.swift.swiftmodule,
         ] + compact([module_context.swift.swiftdoc])
     else:
-        # TODO: make this a failure in version 2.x
-        if swiftinterface:
-            print("WARNING: Provided `swiftinterface` attribute will be ignored because `swiftmodule` was provided. This will be an error in a future version of rules_swift.")  # buildifier: disable=print
```
brentleyjones, line 102: Maybe we need `if not swiftmodule:` and a proper `fail()` so the message is clearer?

luispadron: Not sure I understand, i moved the fail up to the top, so these two cases should catch when none or both are specified:

```starlark
if not (swiftinterface or swiftmodule):
   fail("One of 'swiftinterface' or 'swiftmodule' must be specified.")

if swiftinterface and swiftmodule:
  fail("'swiftinterface' may not be specified when 'swiftmodule' is specified.")

```

brentleyjones: Sorry, missed that :+1:

brentleyjones (approved):

PR #1277, opened 2024-06-25
Upgrade swift_argument_parser to 1.3.1.1

brentleyjones:
It has its `max_compatibility_level` increased to support rules_swift 2.0.


brentleyjones:
Already applied here: https://github.com/bazelbuild/bazel-central-registry/pull/2308

PR #1284, opened 2024-07-02
Add features to enable strict concurrency checking

mjburghard:
Add features for the three accepted combinations of `-strict-concurrency=<value>`.

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -351,3 +351,12 @@ SWIFT_FEATURE_THIN_LTO = "swift.thin_lto"
 
 # Enable full LTO and update output-file-map correctly
 SWIFT_FEATURE_FULL_LTO = "swift.full_lto"
+
+# Enable minimal strict concurrency checking (most 'Sendable' checking is disabled)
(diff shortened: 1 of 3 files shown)


brentleyjones:
I think there might be something related to this in `upstream`. And if so, we should cherry-pick it.

mjburghard:
I had a look and do neither see a feature nor any usage of `-strict-concurrency` on `upstream`.

mjburghard:
The two commits containing strict concurrency related code are https://github.com/bazelbuild/rules_swift/commit/86f356cc3ab35d02edd3efe972a76e1d0a6a7261 and https://github.com/bazelbuild/rules_swift/commit/f1765e6e65ab742e332704717ec7e0129a9b60af.

As far as I can see, there is infrastructure in place on `upstream` to handle upcoming an experimental features. I have to take deeper look to fully understand this. I probably post my understandings on the corresponding issue. 
@brentleyjones should we close this for the time being?

brentleyjones:
Yeah, let's close it for now. We will get those changes as we cherry-pick more from upstream. I'll be sure we finish that before Xcode 16 releases, and hopefully a lot sooner.

PR #1286, opened 2024-07-09
Update extract symbol graphs rule to include swiftdoc

luispadron:
This updates the symbol graph extract configurator to additionally include the `.swiftdoc` files created by the the Swift module compilation. This is required to produce documentation comments in the `.symbolgraphs` when using the symbol graph extract rule.

### DIFF
--- a/swift/toolchains/config/compile_config.bzl
+++ b/swift/toolchains/config/compile_config.bzl
@@ -43,6 +43,7 @@ load(
     "SWIFT_FEATURE_DISABLE_SYSTEM_INDEX",
     "SWIFT_FEATURE_EMIT_BC",
     "SWIFT_FEATURE_EMIT_PRIVATE_SWIFTINTERFACE",
+    "SWIFT_FEATURE_EMIT_SWIFTDOC",
     "SWIFT_FEATURE_EMIT_SWIFTINTERFACE",
     "SWIFT_FEATURE_ENABLE_BATCH_MODE",
     "SWIFT_FEATURE_ENABLE_LIBRARY_EVOLUTION",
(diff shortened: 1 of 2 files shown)


swift/toolchains/config/compile_config.bzl
```diff
@@ -817,7 +818,8 @@ def compile_action_configs(
         # maps.
         ActionConfigInfo(
             actions = [SWIFT_ACTION_SYMBOL_GRAPH_EXTRACT],
-            configurators = [_dependencies_swiftmodules_configurator],
+            configurators = [_dependencies_swiftmodules_and_swiftdocs_configurator],
+            features = [SWIFT_FEATURE_EMIT_SWIFTDOC],
```
brentleyjones, line 822: Do we need another configuration for when it's not emitting them?

luispadron: Do we have an example of this somewhere else? Im not super familiar with how the configurator stuff works

brentleyjones: Would be the same as this, just with `not_features`.

luispadron: added this, let me know if that makes sense:


```starlark
        ActionConfigInfo(
            actions = [SWIFT_ACTION_SYMBOL_GRAPH_EXTRACT],
            configurators = [_dependencies_swiftmodules_configurator],
            not_features = [SWIFT_FEATURE_EMIT_SWIFTDOC],
        )
```

brentleyjones (approved):

luispadron:
Going to see if I can optimize this a bit before merging, we should only need the direct `.swiftdoc` not the transitive ones

luispadron:
This works:


```starlark
            if module.name == module_name and swift_module.swiftdoc:
                direct_swiftdocs.append(swift_module.swiftdoc)
```

And makes the inputs more scoped so think we should go with that

PR #1293, opened 2024-07-16
Add `mixed_langauge_library`

brentleyjones:
Replaces `experimental_mixed_langauge_library` in rules_apple. Meant to support SE-0403.


swift/internal/module_maps.bzl
```diff
@@ -148,6 +151,13 @@ def write_module_map(
     content.add_all(dependent_module_names, format_each = '    use "%s"')
     content.add("}")
 
+    if swift_generated_header:
+        content.add("")
+        content.add(module_name, format = 'module "%s".Swift {')
+        _add_headers(headers = [swift_generated_header], kind = "header")
+        content.add("    requires objc")
+        content.add("}")
```
brentleyjones, line 159: I can probably revert this one, since we are modifying whatever existing module map is used. I'll wait until I'm 100% sure though.

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        direct_public_headers = outputs,
+        includes = depset(includes),
+    )
+
+    return [
+        DefaultInfo(files = outputs_depset),
+        CcInfo(compilation_context = compilation_context),
+    ]
+
+_mixed_language_library_umbrella_header = rule(
```
brentleyjones, line 94: Not sure on if we should place this somewhere else in order to allow it to have a non-underscored name. It is an implementation detail, so I'm not sure how much that matters.

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        module_map_file = module_map,
+        module_name = ctx.attr.module_name,
+        umbrella_header = ctx.file.umbrella_header,
+        public_headers = ctx.files.hdrs,
+        public_textual_headers = ctx.files.textual_hdrs,
+    )
+
+    return [DefaultInfo(files = depset([module_map]))]
+
+_mixed_language_library_internal_module_map = rule(
```
brentleyjones, line 125: Not sure on if we should place this somewhere else in order to allow it to have a non-underscored name. It is an implementation detail, so I'm not sure how much that matters.

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        # migrated to use `CcLinkingContext`.
+        apple_common.new_objc_provider(
+            providers = get_providers(
+                [clang_target, swift_target],
+                apple_common.Objc,
+            ),
+        ),
+    ]
+
+_mixed_language_library = rule(
```
brentleyjones, line 318: I feel like this one more than the others should probably have a public name, even if it's not publicly exposed?

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+    attrs = dicts.add(
+        swift_toolchain_attrs(),
+        {
+            "clang_target": attr.label(
+                mandatory = True,
+                providers = [CcInfo],
+            ),
```
brentleyjones, line 325: This and the `swift_target` attribute make sense for this rule, but do break some existing aspects. I could see using `private_deps` with a implicit contract that the first one is `swift_target` and the second is `clang_target`, since there is a higher chance that aspects will propagate correctly over it?

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+    # target to import the module map for the `_clang` target.
+    internal_swift_interop_name = name + "_swift_interop"
+    swift_interop_hint(
+        name = internal_swift_interop_name,
+        module_map = module_map,
+        module_name = module_name,
+    )
+    headers_library_name = name + "_headers"
+    native.objc_library(
+        name = headers_library_name,
```
brentleyjones, line 535: Having up to 4 additional targets (`_modulemap`, `_umbrella_header`, `_swift_interop`, and `_headers`) is annoying, but there isn't an easy way to replicate what `objc_library` does for `includes`. This way gets us consistent compilation.

Once `objc_library` is fully starlarkified, and we can do whatever it does, or even reuse most of its code, we can probably roll all of this into a special `swift_library` implementation.

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+
+output="$1"
+input="$2"
+module_name="$3"
+generated_header="$4"
+
+# FIXME: Actually parse the map and fix the paths relative to the original
+# location, since it might not be the generated one. We can't use VFS overlay
+# like SE-0403, so rewriting the modulemap is the only way.
+sed 's|header "|header "../../|g' "$input" > "$output"
```
brentleyjones, line 168: I really really really don't like this, but I don't see another way. If we didn't support custom module maps, then we would just generate a new module map here and be done with it. But if we didn't support custom module maps there would be a class of SwiftPM packages that we couldn't support.

swift/internal/feature_names.bzl
```diff
@@ -132,7 +132,9 @@ SWIFT_FEATURE_NO_GENERATED_MODULE_MAP = "swift.no_generated_module_map"
 
 # If enabled, the parent directory of the generated module map is added to
 # `CcInfo.compilation_context.includes`. This allows `objc_library` to import
-# the Swift module.
+# the Swift module. If you swap this faeture between enabled and disabled, and
```
luispadron, line 135:
```suggestion
# the Swift module. If you swap this feature between enabled and disabled, and
```

swift/mixed_language_library.bzl
luispadron: Maybe we can split this up into separate files to make it easier to work with but keep them internal in like: `./swift/internal/mixed_language/`.

Another idea: move these (including the public API) to a `./mixed_language` package. Would make it clear that it's not just Swift based on directory layout and with loads like: `load("@rules_swift//mixed_language:mixed_language_library.bzl", "mixed_language_library")`. If we do this we might even make want to make things like `extended_module_map`, `mixed_language_umbrella_header` public API?

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        "{name}/{module_name}/{module_name}.h".format(
+            module_name = ctx.attr.module_name,
+            name = ctx.attr.name,
+        ),
+    )
+    _write_umbrella_header(
+        actions = actions,
+        headers = ctx.files.hdrs,
+        umbrella_header = umbrella_header,
+    )
```
luispadron, line 76: nit: not sure making this its own function gains us much since its only used here, thoughts on just keeping it in the implementation function.

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+
+    return [
+        DefaultInfo(files = outputs_depset),
+        CcInfo(compilation_context = compilation_context),
+    ]
+
+_mixed_language_library_umbrella_header = rule(
+    attrs = {
+        "actual_name": attr.string(mandatory = True),
```
luispadron, line 96: nit: naming is hard, maybe `umbrella_header_name` or `header_name`?

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        includes = depset(includes),
+    )
+
+    return [
+        DefaultInfo(files = outputs_depset),
+        CcInfo(compilation_context = compilation_context),
+    ]
+
+_mixed_language_library_umbrella_header = rule(
+    attrs = {
```
luispadron, line 95: Could we add docs to the `attrs` :pray:

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        "actual_name": attr.string(mandatory = True),
+        "includes": attr.string_list(),
```
luispadron, line 97: Is this used anywhere? It seems we only ever set it here:

```starlark
includes = [paths.dirname(umbrella_header.dirname)]
```

swift/mixed_language_library.bzl
luispadron: I prefer the each rule to be in it's own `.bzl` and we can just add them to `internal` somewhere so its clear its not public API

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        umbrella_header = ctx.file.umbrella_header,
+        public_headers = ctx.files.hdrs,
+        public_textual_headers = ctx.files.textual_hdrs,
+    )
+
+    return [DefaultInfo(files = depset([module_map]))]
+
+_mixed_language_library_internal_module_map = rule(
+    attrs = {
+        "actual_name": attr.string(mandatory = True),
```
luispadron, line 127: Same here about this attr being more clear on what this name is for and docs 🙏🏼

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+)
+
+# `_mixed_language_library_internal_module_map`
+
+def _mixed_language_library_internal_module_map_impl(ctx):
+    actions = ctx.actions
+
+    module_map = actions.declare_file(ctx.attr.actual_name + ".modulemap")
+
+    # TODO: Set `dependent_module_names`
```
luispadron, line 112: - [ ] Set `dependent_module_names`

```suggestion
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+    doc = """\
+Creates the module map file that is used internally by the clang and swift
+library targets in the `mixed_language_library` macro.
+""",
+    implementation = _mixed_language_library_internal_module_map_impl,
+)
+
+# `_mixed_language_library`
+
+def _write_extended_module_map(
```
luispadron, line 142: Thoughts on moving this to `module_maps.bzl` and dropping the underscore?

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+set -eu
+
+output="$1"
+input="$2"
+module_name="$3"
+generated_header="$4"
+
+# FIXME: Actually parse the map and fix the paths relative to the original
+# location, since it might not be the generated one. We can't use VFS overlay
+# like SE-0403, so rewriting the modulemap is the only way.
```
luispadron, line 167: Could we elaborate on why we can't use VFS overlay here like SE-0403

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+
+module "$module_name".Swift {
+    header "$generated_header"
+
+    requires objc
+}
+EOF
+""",
+        inputs = [original_module_map],
+        mnemonic = "ExtendMixedLanguageModuleMap",
```
luispadron, line 180: If we move this to `module_map.bzl`:

```suggestion
        mnemonic = "ExtendedModuleMap",
```

swift/mixed_language_library.bzl
luispadron: I like `private_deps` and we just add some `fails` that ensure the contract is upheld.

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        module_name = None,
+        private_deps = [],
+        swift_copts = [],
+        swift_defines = [],
+        swift_srcs,
+        textual_hdrs = [],
+        umbrella_header = None,
+        deps = [],
+        **kwargs):
+    """Creates a mixed language library from a clang and swift library target \
```
luispadron, line 375:
```suggestion
    """Creates a mixed language library from a clang and Swift library target \
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        name: The name of the target.
+        clang_copts: The compiler flags for the clang library. These will only
+            be used for the clang library. If you want them to affect the swift
+            library as well, you need to pass them with `-Xcc` in `swift_copts`.
+        clang_defines: Extra clang `-D` flags to pass to the compiler. They
+            should be in the form `KEY=VALUE` or simply `KEY` and are passed not
+            only to the compiler for this target (as `clang_copts` are) but also
+            to all dependers of this target. Subject to "Make variable"
+            substitution and Bourne shell tokenization.
+        clang_srcs: The sources for the clang library.
```
luispadron, line 388: To match `hdrs`:

```suggestion
        clang_srcs: The list of C, C++, Objective-C, or Objective-C++ sources for the clang library.
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        enable_modules: Enables clang module support (via `-fmodules`). Setting
+            this to `True`  will allow you to `@import` system headers and other
+            targets: `@import UIKit;` `@import path_to_package_target;`.
+        hdrs: The list of C, C++, Objective-C, or Objective-C++ header files
+            published by this library to be included by sources in dependent
+            rules. This can't include `umbrella_header`.
+        includes: List of `#include`/`#import` search paths to add to this
```
luispadron, line 395: Do these `includes` get added to the Swift target as well? I didn't see them forwarded in the `swift_library` call below, if thats correct this should specify it's only for the clang target

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+            this and downstream targets.
+        deps: A list of targets that are dependencies of the target being built.
+        **kwargs: Additional arguments to pass to the underlying clang and swift
+            library targets.
+    """
+    features = kwargs.pop("features", [])
+    tags = kwargs.pop("tags", [])
+    visibility = kwargs.pop("visibility", None)
+
+    internal_tags = (["manual"] + tags) if "manual" not in tags else []
```
luispadron, line 451: Is the intention to always forward the `tags` to the internal targets? if so, i think we want:

```suggestion
    internal_tags = (["manual"] + tags) if "manual" not in tags else tags
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        )
+    if kwargs.pop("copts", None):
+        fail(
+            "Use 'clang_copts' and/or 'swift_copts' to set copts.",
+            attr = "copts",
+        )
+    if kwargs.pop("implementation_deps", None):
+        fail(
+            """\
+'implementation_deps' is not supported. Use 'private_deps' to set \
```
luispadron, line 468:
```suggestion
'implementation_deps' is not supported for `mixed_language_library`. Use 'private_deps' to set \
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+    if not kwargs.pop("generates_header", True):
+        fail(
+            """\
+Mixed language libraries must generate an Objective-C interop header.\
+""",
+            attr = "generates_header",
+        )
+    if kwargs.pop("copts", None):
+        fail(
+            "Use 'clang_copts' and/or 'swift_copts' to set copts.",
```
luispadron, line 462:
```suggestion
            "`copts` is not supported for `mixed_language_library`. Use 'clang_copts' and/or 'swift_copts' to set copts.",
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+    features = kwargs.pop("features", [])
+    tags = kwargs.pop("tags", [])
+    visibility = kwargs.pop("visibility", None)
+
+    internal_tags = (["manual"] + tags) if "manual" not in tags else []
+
+    if not kwargs.pop("generates_header", True):
+        fail(
+            """\
+Mixed language libraries must generate an Objective-C interop header.\
```
luispadron, line 456:
```suggestion
`mixed_language_library` must generate a clang interop header for the Swift library, set `generates_header` to `True` (default).
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        fail(
+            """\
+'implementation_deps' is not supported. Use 'private_deps' to set \
+implementation deps.\
+""",
+            attr = "implementation_deps",
+        )
+
+    if not clang_srcs and not hdrs:
+        fail("'clang_srcs' or 'hdrs' must not be empty.", attr = "clang_srcs")
```
luispadron, line 475:
```suggestion
        fail("`mixed_language_library` requires either 'clang_srcs' and/or 'hdrs' to be non-empty. If this is not a mixed language Swift library, use `swift_library` instead.", attr = "clang_srcs")
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+    if not swift_srcs:
+        fail("'swift_srcs' must not be empty.", attr = "swift_srcs")
```
luispadron, line 477:
```suggestion
        fail("`mixed_language_library` requires 'swift_srcs' to be non-empty. If this is not a mixed language Swift library, use a clang only library rule like `cc_library` or `objc_library` instead.", attr = "swift_srcs")
```

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        private_deps = private_deps + [umbrella_header]
+
+    # The umbrella header is a public header
+    adjusted_hdrs = [umbrella_header] + hdrs
+
+    # `_headers` is the public interface of the `_clang` target, to be consumed
+    # by the `_swift` target. It won't be propagated to dependers of the mixed
+    # language library. The `_swift_interop` aspect hint allows the `_swift`
+    # target to import the module map for the `_clang` target.
+    internal_swift_interop_name = name + "_swift_interop"
```
luispadron, line 527: In rules_apple these internal targets were named `.internal.<rule>` should we encourage that patten or do we like underscores better here?

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        tags = internal_tags,
+        textual_hdrs = textual_hdrs,
+        **kwargs
+    )
+
+    swift_library_name = name + "_swift"
+    swift_library(
+        name = swift_library_name,
+        srcs = swift_srcs,
+        copts = ["-import-underlying-module"] + swift_copts,
```
luispadron, line 550: Should we check that this `copt` isn't already in `swift_copts` before adding it here

swift/mixed_language_library.bzl
```diff
@@ -0,0 +1,616 @@
...
+        # and extended module map.
+        private_deps = [":" + headers_library_name] + private_deps,
+        tags = internal_tags,
+        deps = deps,
+        **kwargs
+    )
+
+    # TODO: Should this support `cc_library`, and if so, how do we choose?
+    # We can't look at the `clang_srcs` attribute because it might be a
+    # `select`.
```
luispadron, line 574: I think we should, but also I don't know of any example where C++ is mixed with Swift so it's probably not super important unless we get some real world uses of it.

One easy way to support might be to be explicit with the attr names and drop `clang_`, so: `objc_srcs` instead of `clang_srcs`. Which would let us have a `cc_srcs` attr. Then we can `fail` if both `objc_srcs` and `clang_srcs` are provided, etc.

luispadron (commented):
Thank you so much for starting the move over to **rules_swift,** this is looking great!

swift/mixed_language_library.bzl
brentleyjones: I didn't since they were internal, but I can for devs, yes.

swift/mixed_language_library.bzl
brentleyjones: Not anymore, no.

swift/mixed_language_library.bzl
brentleyjones: On the flip side: our aspects depend on only adding information from the `swift_library`.

swift/mixed_language_library.bzl
brentleyjones: They do, via the `_headers` target.

swift/mixed_language_library.bzl
brentleyjones: These are a lot more visible, especially since they are on public rule kinds, so I decided to go with the "clearer" names.

swift/mixed_language_library.bzl
brentleyjones: We can, but I would probably want to `fail` then instead of not adding it. If the user is passing that flag, they shouldn't be using this macro.

swift/mixed_language_library.bzl
brentleyjones: There is also using this from Linux, which would require the `cc_library` target.

swift/mixed_language_library.bzl
brentleyjones: Actually, `swift_copts` might be a `select`, so we can't check it here.

luispadron: ah right this is fine then!

swift/mixed_language_library.bzl
brentleyjones: I wonder if we should just have a `bool` attribute to determine this?

swift/mixed_language_library.bzl
brentleyjones: I split them up, but left all of the rules as internal for now, just the macro is public.

swift/mixed_language_library.bzl
luispadron: I'm cool with keeping it as a TODO until we get some requests for it and have more time to think about it

swift/mixed_language_library.bzl
brentleyjones: There is a bug with `private_deps` that prevents this from working. I'll land that fix in another PR first.

brentleyjones: And maybe that is how it should be. Thankfully using `deps` instead works. I added a comment explaining it.

mixed_language/mixed_language_library.bzl
```diff
@@ -0,0 +1,361 @@
...
+        module_name = module_name,
+        swift_target = ":" + swift_library_name,
+        tags = tags,
+        testonly = testonly,
+        umbrella_header = umbrella_header,
+        visibility = visibility,
+        # We collect the same deps as the `_swift` target in order to propagate
+        # transitive `SwiftInfo`. It's inefficent to try to do that from the
+        # `SwiftInfo` of the `_swift` target.
+        deps = deps
```
brentleyjones, line 360: We need to add `private_deps` here as well.

swift/mixed_language_library.bzl
brentleyjones: I think I'm going to leave this as a todo for now. Not currently blocking anything, but we really should be doing it.

swift/mixed_language_library.bzl
brentleyjones: Going to leave for now.

swift/mixed_language_library.bzl
brentleyjones: Implied since the origin of the error will be `mixed_language_library`?

swift/mixed_language_library.bzl
brentleyjones: Implied since the origin of the error will be `mixed_language_library`?

brentleyjones: Adjusted anyway.

swift/mixed_language_library.bzl
brentleyjones: Adjusted anyway.

brentleyjones:
There is a request to allow the macro to use non-standard rules for the clang and swift targets. So instead of `swift_library` one could use their own `custom_swift_library`. I think this can be done in a followup change, to prevent blocking the core functionality here.

luispadron (approved):

PR #1294, opened 2024-07-17
Adding target name to output path for plugin executable

tymurmustafaiev:
Adding target_name to output path of swift plugin executable. Feature covered by `SWIFT_FEATURE_ADD_TARGET_NAME_TO_OUTPUT` flag.

Without this change and `SWIFT_FEATURE_ADD_TARGET_NAME_TO_OUTPUT` enabled, we have naming collisions as bazel makes executable file on root folder, and in this root folder we make our `target_name` folder with all other files like: swiftinterface, headers, etc.
By adding this change, we put executable in same target_name folder with all other generated files


brentleyjones (approved):

brentleyjones:
Thanks!

PR #1302, opened 2024-08-24
Expose additional utilities through swift_common for custom swift libraries

AttilaTheFun:
3rd party consumers of rules_swift may wish to create custom rules which behave similar to swift_library from the perspective of other rules and tools.

Right now this is a fairly complex process as you would need to reimplement or fork some logic which is currently internal to rules_swift. It is easy to make mistakes, or this logic could diverge from the rules_swift implementation over time.

I want to provide, through swift_common, an API which is basically just "give us a list of swift source files and their dependencies, and we'll compile them for you and create the necessary providers".

This PR does exactly that, and provides an example demonstrating how easy it is to create such a rule using these utilities from swift_common.

One question you might ask is, "Why not just create a macro which runs your custom rule to create swift source file(s) as outputs, and pass those to a real swift_library target?"

This can work for some simpler use cases, but when your rule does complex logic involving aspects and other custom rules, things break down quickly. It also creates additional targets in the build graph which could be confusing for consumers.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -156,6 +156,38 @@ A `struct` with the following fields:
           macros).
 
 
+<a id="swift_common.compile_and_create_linking_context"></a>
+
+## swift_common.compile_and_create_linking_context
+
+<pre>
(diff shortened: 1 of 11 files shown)


AttilaTheFun:
Closing this for now as I managed to use an output attribute + swift library though this would still be nice to have.

PR #1317, opened 2024-09-30
Add execution groups for actions that don't have to use the same toolchain used when building the target

brentleyjones:
Once we migrate to platforms/toolchains, this will avoid the default behavior where a toolchain's `exec_compatible_with` determines the action routing for all actions under that rule.

PiperOrigin-RevId: 501277236
(cherry picked from commit 11e5b45a3cff3ec240bb9fbbabb90d6572c86e54)

### DIFF
--- a/swift/swift_test.bzl
+++ b/swift/swift_test.bzl
@@ -60,6 +60,9 @@ load(
     "create_swift_module_context",
 )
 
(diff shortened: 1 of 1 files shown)


luispadron (approved):

brentleyjones:
This requires Bazel 7. Will hold until we drop Bazel 6.

PR #1319, opened 2024-09-30
Switch the Swift rules over to toolchains

brentleyjones:
PiperOrigin-RevId: 503982850
(cherry picked from commit c42aad205d457f6fede1da7ab245800d3ec76641)

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -10,13 +10,15 @@ x_defaults:
     test_targets:
       - "//examples/..."
       - "//test/..."
-  linux_common: &linux_common
+  linux_environment: &linux_environment
     platform: ubuntu2004
     environment:
       CC: clang
       SWIFT_VERSION: "5.9.2"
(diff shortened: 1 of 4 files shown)


luispadron (approved):

brentleyjones:
I need to add toolchains as well. Not sure if this is a breaking change. I know for rules_apple to be able to use this they need to update a couple call sites. But since we are still sticking the toolchain in an attribute, I believe the old ways continue to work. 

Maybe it can get odd in mixed non-toolchain/toolchain scenarios? I guess wondering if we need to bump the compatibility version again for this change (I hope not).

brentleyjones:
Without changing anything in rules_apple, this change + #1321 + registering the toolchains still works for me locally.

brentleyjones:
Will merge #1323 first.

brentleyjones:
This change passes our CI, so it's probably safe without bumping our compatibility version.

PR #1323, opened 2024-10-01
Fix non-Apple toolchains

brentleyjones:


brentleyjones:
https://buildkite.com/bazel/rules-swift-swift/builds/5591

luispadron (approved):

PR #1329, opened 2024-10-01
Add the `swift_module_mapping` rule

brentleyjones:
This change only adds the new rule and its provider. The build flag (mentioned in documentation) and compilation support will be added in subsequent changes.

PiperOrigin-RevId: 486235761
(cherry picked from commit 6b56b0114674e7d88ab6f2e72d6c3fb5fb646093)

### DIFF
--- a/swift/BUILD
+++ b/swift/BUILD
@@ -224,6 +224,14 @@ bzl_library(
 )
 
 bzl_library(
+    name = "swift_module_mapping",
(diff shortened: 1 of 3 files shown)


luispadron (approved):
Should this be added to https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md

brentleyjones:
Probably, when it starts to do something.

PR #1335, opened 2024-10-03
Add a Boolean `library_evolution` attribute to `swift_library`

brentleyjones:
Historically, library evolution has been supported by a global flag that Apple framework rules transition on in order to build all dependencies with library evolution enabled. In retrospect, this is wrong; we shouldn't be automatically applying library evolution to the whole subgraph in this way. Instead, the owner of the framework should explicitly specify library evolution on the client-facing modules that make up their SDK.

The goal here is to ensure that framework owners carefully audit their public `deps` vs. implementation-only `private_deps` so that the `.swiftinterface` doesn't attempt to import anything that it shouldn't, especially with Swift 5.7 validating emitted interfaces after compilation by default.

This change only adds the new attribute; the flag is still honored while we have clients and rules_apple depending on it.

PiperOrigin-RevId: 486750877
(cherry picked from commit b9175e66bb1d0e0f8eda6afdb413c685ce7943d7)

Cherry-pick notes: With this change the private swiftinterface is enabled by `ctx.attr.library_evolution` as well. It can be disabled with `features = ["-swift.emit_private_swiftinterface”]`.


luispadron (approved):

tumata:
Hi,

I've recently started using Bazel to automate the generation of `XCFrameworks` for distributing our internal frameworks. While reviewing the documentation on the master branch, I encountered the error: `no such attribute 'library_evolution' in 'swift_library' rule` when attempting to use the `library_evolution` parameter.

Now finding this PR from only 4 days ago, it that explains my issue!

From your PR description, it seems that library evolution is enabled by default in the current version `2.1.1`. Could you please confirm if that's the case?

Thank you for your help!

luispadron:
@tumata this commit hasn't made it into a release yet. You can enable it via `--features="swift.enable_library_evolution"`

tumata:
Thanks for your help @luispadron. Where do I add this feature you're mentioning? 

Should I add it to my `bazel build` command? 

```
bazel build //:my_xcframework --features='swift.enable_library_evolution'
```

Please forgive my ignorance, I'm very fresh with Bazel :)

luispadron:
Yes, it's a flag, you can read more on it: https://bazel.build/reference/command-line-reference#flag--features

You can also add it to a `.bazelrc` file so it's always applied, or use the `features` attribute for specific targets

PR #1346, opened 2024-10-04
Update example macro based on latest swift-syntax APIs on main

brentleyjones:
PiperOrigin-RevId: 547489523
(cherry picked from commit 8c439d08bbc49486e0fde4140411d72c1715e930)


luispadron (approved):

brentleyjones:
Maybe we should change this to use rules_swift_package_manager instead 🤔.

luispadron:
> Maybe we should change this to use rules_swift_package_manager instead 🤔.

Would be nice to share the dep in the same dependency graph but I don't think we can until https://github.com/cgrindel/rules_swift_package_manager/issues/769

PR #1348, opened 2024-10-07
Remove `compilation_attrs` and `library_rule_attrs` from `swift_common`

brentleyjones:
No custom rules are using these today. If they need to, they can just declare whatever attributes they need on their own.

PiperOrigin-RevId: 495328434
(cherry picked from commit 28a27e1f8d9263cfba2210feb6f3192024822e80)

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -73,59 +73,6 @@ A C++ `FeatureConfiguration` value (see
   for more information).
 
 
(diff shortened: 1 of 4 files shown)


mattrobmattrob (approved):

brentleyjones:
Reverted in #1414, to be re-applied as part of major version bump.

PR #1349, opened 2024-10-07
Remove the `swift_explicit_module` output group from `swift_clang_module_aspect`

brentleyjones:
This is no longer used anywhere.

PiperOrigin-RevId: 495593700
(cherry picked from commit 3cb3d5ba8390a9e8de2d78197694059fd0bff093)

### DIFF
--- a/swift/swift_clang_module_aspect.bzl
+++ b/swift/swift_clang_module_aspect.bzl
@@ -534,12 +534,9 @@ def _handle_module(
         ),
     ]
 
-    if precompiled_module:
-        output_groups = {"swift_explicit_module": depset([precompiled_module])}
-        if indexstore_directory:
-            output_groups["indexstore"] = depset([indexstore_directory])
+    if indexstore_directory:
         providers.append(
-            OutputGroupInfo(**output_groups),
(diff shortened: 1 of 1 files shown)


swift/swift_clang_module_aspect.bzl
```diff
@@ -534,12 +534,9 @@ def _handle_module(
         ),
     ]
 
-    if precompiled_module:
-        output_groups = {"swift_explicit_module": depset([precompiled_module])}
-        if indexstore_directory:
-            output_groups["indexstore"] = depset([indexstore_directory])
+    if indexstore_directory:
```
mattrobmattrob, line 537: And I suppose the removal of the `precompiled_module` clause is acceptable?
```suggestion
    if precompiled_module and indexstore_directory:
```

mattrobmattrob (approved):

swift/swift_clang_module_aspect.bzl
brentleyjones: Yeah, it's what upstream has. Plus it probably should have always been this way.

PR #1386, opened 2024-10-14
Remove two uses of the `swift.objc_link_flag` feature

brentleyjones:
The new toolchain has been around long enough.

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -387,22 +387,6 @@ def create_linking_context_from_compilation_outputs(
             ),
         )
 
-    if is_feature_enabled(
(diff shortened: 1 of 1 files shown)


thii (approved):

brentleyjones:
I think we need c58d81a5fedcee78997dd085185d5744ad7d9e26 first.

brentleyjones:
Looks like we need one use of `-ObjC` still since we don't set `language = objc` on `cc_common.link`. Doing that would have the **apple_support** toolchain provide this value for us. We currently can't set that though since it requires more values.

brentleyjones:
Actually, we probably need this forever otherwise `cc_binary` won't link correctly.

PR #1390, opened 2024-10-14
Remove `swift_module_alias`

brentleyjones:
Was originally intended to be like `alias` and just used for migration, but is easily misused to merge/reexport targets, which in turns causes problems for indexing, automation, etc. that try to map symbols back to targets.

Since it generates a new files, if dependent on multiple things to merge, that ends up being a choke point that slows downs builds when someone really only needs a subset of the dependencies.

So just remove it so it isn't something folks will try to reach for with many down sides.

If someone really needs the rexport steps for a migration, they can manually make a targets/srcs to do it, and hopefully thus be aware of the costs so there will migrate off it as soon as possible.

PiperOrigin-RevId: 528836559
(cherry picked from commit 4f27bd0d4217b96838190dfe9e3b99b32e420274)


thii (approved):

manivel-deliveroo:
@brentleyjones I've used `swift_module_alias` in my project and that is breaking now with this change. I'm sorry I'm not fully familiar with the rules. But what is the alternative I should be using? Is it `swift_library`?

brentleyjones:
This is only removed in the upcoming 3.0 release. The 2.x releases (which are the only ones supported right now) still has it. You can copy the old implementation, make a macro that generates a `swift_library` that has `@exports` in it, etc. We will have more guidance once we cut the 3.0 release.

PR #1407, opened 2024-10-15
Upgrade stardoc to 0.7.1

brentleyjones:
Needed to support Bazel HEAD.

This changes the generated docs to include load statements with `@rules_swift`. As part of rules_swift 3.0 we can change all references to the new form as well.


luispadron (approved):

aaronsky:
I made #1462 to try and get the rest of the way there and unblock us, but I may have gone overboard on the implementation. Hope it was helpful for me to try.

PR #1425, opened 2024-10-22
Fix `build_file` labels

brentleyjones:
Regressed for WORKSPACE users in 7fafc9bc47e80ca9fb08affeb44d135818c8c8db.


brentleyjones:
We would have to have another workspace that loads this one, since it the labels do resolve correctly from within the repo. I think that's too much hassle for something that shouldn't really break again. And yeah, I'm not 100% sure we will support WORKSPACE in **rules_swift** 3.0.

PR #1433, opened 2024-10-22
Warn instead of fail when `swiftc` not found on Linux

brentleyjones:


brentleyjones:
Not sure if we should even warn here. Doing it for now though.

cerisier (approved):

PR #1445, opened 2024-10-25
Remove unnecessary `SwiftBinaryInfo` required provider from `plugins` attribute

brentleyjones:
I incorrectly added this as part of the cherry-pick in 5f6ac77099eda1838c36c7e3c7b4e88309caad1d.


AttilaTheFun (approved):
Thanks! This will make it a lot easier to use pre-built swift compiler plugins.

PR #1452, opened 2024-11-07
Stop using config.string(allow_multiple = True)

EdSchouten:
According to the Bazel documentation, this option is deprecated. config.string_list(repeatable = True) should be used instead. This PR is similar to this one, which I sent out for rules_rust:

https://github.com/bazelbuild/rules_rust/pull/2983


EdSchouten:
CI failures seem to be unrelated to the change at hand?

mattrobmattrob (approved):

PR #1454, opened 2024-11-18
Adds using Swift from C++ example

jiawen:


adincebic:
We should probably rebase on top of main.

PR #1458, opened 2024-11-22
Add support for `package_name` in `mixed_language_library`

luispadron:
This attr was missing from the `swift_library`.


aaronsky (approved):

brentleyjones (approved):

luispadron:
@brentleyjones mind admin merging, believe the ci failure is that bazel 8 issue we need to fix

PR #1462, opened 2024-11-30
wip: upgrade stardoc and protobuf to restore Bazel HEAD CI

aaronsky:
Trying my hand at this, to improve personal competency. This updates the minimum protobuf to 29.0, removes the dependency on rules_proto, and updates stardoc to 0.7.2. Some or all of this may be an overreach, but I'm unfamiliar with the constraints.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -13,8 +13,10 @@ module.
 ## create_swift_interop_info
 
 <pre>
-create_swift_interop_info(<a href="#create_swift_interop_info-exclude_headers">exclude_headers</a>, <a href="#create_swift_interop_info-module_map">module_map</a>, <a href="#create_swift_interop_info-module_name">module_name</a>, <a href="#create_swift_interop_info-requested_features">requested_features</a>, <a href="#create_swift_interop_info-suppressed">suppressed</a>,
-                          <a href="#create_swift_interop_info-swift_infos">swift_infos</a>, <a href="#create_swift_interop_info-unsupported_features">unsupported_features</a>)
+load("@rules_swift//doc:doc.bzl", "create_swift_interop_info")
+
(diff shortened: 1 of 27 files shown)


aaronsky:
alternative approach to bumping up protobuf so much would be to pin to [rules_proto 7.0.2](https://github.com/bazelbuild/rules_proto/releases/tag/7.0.2)

doc/api.md
```diff
@@ -74,7 +76,9 @@ A provider whose type/layout is an implementation detail and should not
 ## derive_swift_module_name
 
 <pre>
-derive_swift_module_name(<a href="#derive_swift_module_name-args">args</a>)
+load("@rules_swift//doc:doc.bzl", "derive_swift_module_name")
```
brentleyjones, line 79: Ugh, this is annoying.

brentleyjones:
> alternative approach to bumping up protobuf so much would be to pin to [rules_proto 7.0.2](https://github.com/bazelbuild/rules_proto/releases/tag/7.0.2)

Pros and cons of these approaches? Could the protobuf bump break people that depend on us?

doc/api.md
brentleyjones: The main annoying thing is `doc.bzl` versus where it actually is 😕.

aaronsky:
I'm not super-clued into the goings-on around rules_proto/protobuf, but I saw after writing that comment that stardoc 0.7.2 declares a dep on protobuf 29.0. So I think it might be a moot point.

doc/api.md
aaronsky: is this something we should bring up in the stardoc repo as a bug? or maybe there's another way we're supposed to be organizing these?

brentleyjones: Could probably use a full issue. I saw it mentioned in a comment here: https://github.com/bazelbuild/stardoc/issues/94#issuecomment-2492218404

I think we need to go the route of concatenating the output instead.

keith:
we could probably pull in a patch to revert https://github.com/bazelbuild/stardoc/pull/216 to avoid the load issues

PR #1466, opened 2024-12-05
Run stardoc with older bazel

keith:
Just to make CI green for now


keith:
this will be annoying locally once 8.x is released (unless we pin the whole repo back locally) but probably not worth blocking anything on this

aaronsky (approved):

PR #1481, opened 2025-01-24
Mark non_module_deps extension as reproducible

cerisier:
As per https://bazelbuild.slack.com/archives/C04DFUBQGSU/p1737756347358239

I was curious about the "reproducibility" of `swift_autoconfiguration` but from a `MODULE.bazel.lock` it seems it is.

I had to bump `bazel_features` to `1.9.0` to check the availability of the `reproducible` field on `extension_metadata`.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,7 +6,7 @@ module(
     repo_name = "build_bazel_rules_swift",
 )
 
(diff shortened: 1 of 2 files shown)


keith (approved):

keith:
some lint failure

luispadron (approved):

Issue #1199, opened 2024-05-07
swift_rules_extra_dependencies() fails with rules_proto 6.0.0


jiawen:
The recent [rules_proto 6.0.0 release](https://github.com/bazelbuild/rules_proto/releases/tag/6.0.0) migrates `rules_proto_toolchains()` from `repositories.bzl` to `toolchains.bzl`. We need a migration path for [swift/extras.bzl](https://github.com/bazelbuild/rules_swift/blob/1aec64c218fc057c2a836e67bd55bc514e0ef8bb/swift/extras.bzl).

Current:
```
load(
    "@rules_proto//proto:repositories.bzl",
    "rules_proto_dependencies",
    "rules_proto_toolchains",
)
```

Should be:
```
load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
load("@rules_proto//proto:toolchains.bzl", "rules_proto_toolchains")
```

PR #1482, opened 2025-02-04
Support rules proto 6.0.0

NilsIrl:
Fix https://github.com/bazelbuild/rules_swift/issues/1199


NilsIrl:
Yep, I think MODULE.bazel would need to be updated too

mattrobmattrob:
Typically, rules_swift doesn't update dependencies unless it needs some portion of, for example, rules_proto 6.X.X in its implementation. I'd recommend applying this change via something like `single_version_override.patches` until rules_proto is updated.

NilsIrl:
I do not have the bandwidth at the moment

PR #1488, opened 2025-02-06
Always create all toolchains

brentleyjones:
Now that rules_swift is using the toolchains API, we can always create all of our toolchains. This enables cross-platform RBE builds.

### DIFF
--- a/swift/internal/swift_autoconfiguration.bzl
+++ b/swift/internal/swift_autoconfiguration.bzl
@@ -212,14 +212,10 @@ def _create_linux_toolchain(repository_ctx):
 No 'swiftc' executable found in $PATH. Not auto-generating a Linux Swift \
 toolchain.
 """)  # buildifier: disable=print
-        repository_ctx.file(
-            "BUILD",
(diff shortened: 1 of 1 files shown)


swift/internal/swift_autoconfiguration.bzl
```diff
@@ -290,22 +273,9 @@ def _create_xcode_toolchain(repository_ctx):
...
-    "@build_bazel_rules_swift//swift/toolchains:xcode_swift_toolchain.bzl",
-    "xcode_swift_toolchain",
-)
-
-package(default_visibility = ["//visibility:public"])
-
+    return """\
 xcode_swift_toolchain(
-    name = "toolchain",
+    name = "xcode-toolchain",
```
brentleyjones, line 278: This was only possible because we removed `swift_toolchain_attrs`. We can't cherry-pick this change into `2.x` because downstream rulesets might still use `swift_toolchain_attrs` and this change would break them.

aaronsky (approved):

PR #1492, opened 2025-02-11
:cherries: add `swift.suppress_warnings` feature

aaronsky:
cherry pick of #1489

### DIFF
--- a/swift/internal/feature_names.bzl
+++ b/swift/internal/feature_names.bzl
@@ -384,6 +384,9 @@ SWIFT_FEATURE__SUPPORTS_V6 = "swift._supports_v6"
 SWIFT_FEATURE_DISABLE_SWIFT_SANDBOX = "swift.disable_swift_sandbox"
 
 # Pass -warnings-as-errors to the compiler.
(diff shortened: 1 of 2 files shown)


brentleyjones (approved):

swift/internal/feature_names.bzl
```diff
@@ -383,6 +383,9 @@ SWIFT_FEATURE__SUPPORTS_V6 = "swift._supports_v6"
 # Enabled by default for Swift 5.10+ on macOS.
 SWIFT_FEATURE_DISABLE_SWIFT_SANDBOX = "swift.disable_swift_sandbox"
 
+# Pass -warnings-as-errors to the compiler.
```
qyang-nj, line 386: @aaronsky I think this comment is not correct. It should be `-suppress-warnings`, right?

aaronsky: ah you’re right, silly copy-paste job. the comment is wrong, the flag it sets is right. i’ll fix this when i get a chance.

mattrobmattrob: Got it: https://github.com/bazelbuild/rules_swift/pull/1499

PR #1494, opened 2025-02-18
Stop using `plugins` from transitive dependencies 

qyang-nj:
Currently, if we have the following setup,
```
swift_library (
    name = "LibA"
    plugins = ["//:Macros"],
)

swift_library (
    name = "LibB"
    plugins = ["//:Macros"],
    deps = [":LibA"],
)
```
We would see `-load-plugin-executable /path/to/Macros#Macros` twice while compiling LibB. With a more complicated dependency graph, we are seeing hundreds of  `-load-plugin-executable` of the same plugin on some modules.

Actually, we don't need the plugins from the transitive dependencies. This is already explained the code comment, but the actual code behaves differently. This PR fixes it.

### DIFF
--- a/examples/xplatform/macros/BUILD
+++ b/examples/xplatform/macros/BUILD
@@ -42,12 +42,14 @@ universal_swift_compiler_plugin(
 swift_binary(
     name = "stringify_client",
     srcs = ["StringifyClient.swift"],
+    plugins = [":stringify_macro"],
     deps = [":stringify"],
 )
 
 swift_binary(
     name = "stringify_universal_client",
     srcs = ["StringifyUniversalClient.swift"],
+    plugins = [":stringify_macro_universal"],
     deps = [":stringify_universal"],
 )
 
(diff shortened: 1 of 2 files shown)


swift/internal/compiling.bzl
```diff
@@ -603,9 +603,6 @@ def compile(
     # passed; the compiler does not attempt to load them when deserializing
     # modules.
     used_plugins = list(plugins)
-    for module_context in transitive_modules:
-        if module_context.swift and module_context.swift.plugins:
-            used_plugins.extend(module_context.swift.plugins)
```
qyang-nj, line 608: The comment above says "Plugins that are only used by transitive dependencies do *not* need to be passed".

examples/xplatform/macros/BUILD
```diff
@@ -42,12 +42,14 @@ universal_swift_compiler_plugin(
 swift_binary(
     name = "stringify_client",
     srcs = ["StringifyClient.swift"],
+    plugins = [":stringify_macro"],
```
qyang-nj, line 45: Since `StringifyClient.swift` is actually using `#stringify`, I think it's reasonable to be explicit about this?

keith:
i think the point of this is that the only library that should need to declare the plugin dep is the one that defines the shim for using it?

would deduping the args solve your case instead? that would not affect behavior

qyang-nj:
I think that makes sense. I'll close this one and create new PR.

PR #1495, opened 2025-02-19
Deduplicate plugin arguments

qyang-nj:
Currently, if we have the following setup,
```
swift_library (
    name = "LibA"
    plugins = ["//:Macros"],
)

swift_library (
    name = "LibB"
    plugins = ["//:Macros"],
    deps = [":LibA"],
)
```
We would see `-load-plugin-executable /path/to/Macros#Macros` twice while compiling LibB. With a more complicated dependency graph, we are seeing hundreds of -load-plugin-executable of the same plugin on some modules.

This PR deduplicates the plugins arguments.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -14,6 +14,7 @@
 
 """Implementation of compilation logic for Swift."""
 
+load("@bazel_skylib//lib:collections.bzl", "collections")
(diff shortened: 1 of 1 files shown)


keith (approved):

qyang-nj:
@keith Do you mind merging this? 🙏

keith:
thanks!  sorry was just giving it some time in case others wanted to comment

PR #1504, opened 2025-04-01
Support index-import on all Xcode 16.x versions

luispadron:
This PR updates the index-import dependency to include both version 5.8 and 6.1 as the hash algorithm changed in Swift 6.1.
To make this change backwards compatible we switch to the 5.8 version on Xcode 16.2 and under.


keith (approved):

keith:
an apple_support update will fix last_green

PR #1505, opened 2025-04-01
2.x: Support index-import on all Xcode 16.x versions

luispadron:
This PR updates the index-import dependency to include both version 5.8 and 6.1 as the hash algorithm changed in Swift 6.1.
To make this change backwards compatible we switch to the 5.8 version on Xcode 16.2 and under.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -69,10 +69,11 @@ tasks:
     bazel: latest
     <<: *mac_common
 
-  macos_last_green:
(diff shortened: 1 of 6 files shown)


keith (approved):

luispadron:
@keith im actually going to drop the apple_support update from the 2.x branch unless you think its fine to drop support for Bazel 6 like this?

keith:
yea sounds fine, this branch doesn't need to support last_green

PR #1506, opened 2025-04-03
Update apple_support to latest

luispadron:
Fixes CI


mattrobmattrob (approved):
I'm assuming CI will validate that this is the proper way to handle dep upgrades WRT old Bazel versions and a change like this. 👍

keith (approved):

luispadron:
Hmm good question @mattrobmattrob i don't actually see a non-bzlmod test. Let me check

luispadron:
```
USE_BAZEL_VERSION=7.4.0 bazel build //test/... --noenable_bzlmod
```

Works fine

PR #1523, opened 2025-05-28
Upstream: Add swift_common.is_action_enabled as a public API.

luispadron:
This allows callers to verify whether certain actions that are compiler-version-dependent, like `swift-synthesize-interface`, can be called by a rule or aspect.

Cherry-picks:

- https://github.com/bazelbuild/rules_swift/commit/d2d94bfc2a7b0da7a92bbca77d690bcb5008872d
- https://github.com/bazelbuild/rules_swift/commit/59000d5e1932e7ac3fb67f4476516ae75b80551b
- https://github.com/bazelbuild/rules_swift/commit/9e7c7c42a50e1451a43fa645343da25a374d0f39

### DIFF
--- a/swift/providers.bzl
+++ b/swift/providers.bzl
@@ -114,7 +114,35 @@ def _swift_info_init(
     }
 
 SwiftInfo, _create_raw_swift_info = provider(
-    doc = "Contains information about the compiled artifacts of a Swift module.",
+    doc = """\
+Contains information about the compiled artifacts of a Swift module.
+
+This provider has a custom initializer that will merge the transitive modules of
+a list of `SwiftInfo` providers, rather than a separate "merge" function. The
+correct signature when _creating_ a new `SwiftInfo` provider is the following:
+
+```build
+SwiftInfo(
+    direct_swift_infos,
+    modules,
+    swift_infos,
+)
+```
+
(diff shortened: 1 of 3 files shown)


swift/providers.bzl
```diff
@@ -102,35 +114,7 @@ def _swift_info_init(
...
-*   `modules`: A list of values (as returned by `create_swift_module_context`)
-    that represent Clang and/or Swift module artifacts that are direct outputs
-    of the target being built.
-*   `swift_infos`: A list of `SwiftInfo` providers from dependencies whose
-    transitive modules should be merged into the resulting provider.
-
-When reading an existing `SwiftInfo` provider, it has the two fields described
-below.
-""",
+    doc = "Contains information about the compiled artifacts of a Swift module.",
```
aaronsky, line 117: ? - interesting culling of provider docs, do we want to keep this?

aaronsky (approved):
approved with open question

swift/providers.bzl
luispadron: yeah thought so too, maybe redundant with the docs they added for `_swift_info_init` but that doesnt make into the the generated .md. 

I can undo this part

PR #1527, opened 2025-05-29
Upstream: Update the API of `swift_common.compile_module_interface`.

luispadron:
Cherry-pick: https://github.com/bazelbuild/rules_swift/commit/4c4184942dc091c1b43b0243848876d3c5b90c07

This adds the `target_name` argument that is present in other compile APIs so that supplemental outputs can get unique names based on the name of the Bazel target being compiled (will be used in a follow-up change).

Likewise, this changes the return type of the function from just the module context to a `struct` that contains two fields: `module_context` (the original value), and `supplemental_outputs`, which will eventually be used to provide an indexstore (supported starting from Swift 6.2/Xcode 16.3).

PiperOrigin-RevId: 741516048

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -279,12 +279,22 @@ Compiles a Swift module interface.
 
 **RETURNS**
 
-A Swift module context (as returned by `create_swift_module_context`)
-  that contains the Swift (and potentially C/Objective-C) compilation
-  prerequisites of the compiled module. This should typically be
-  propagated by a `SwiftInfo` provider of the calling rule, and the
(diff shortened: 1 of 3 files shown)


luispadron:
Moving to draft until we can take: https://github.com/bazelbuild/rules_swift/commit/81c5b92823c03238cb9ad6f1b360227be4cb2691

Which requires updates to the worker code.

aaronsky (approved):

keith:
https://github.com/bazelbuild/rules_swift/pull/1705

PR #1528, opened 2025-05-29
Fix build failures in bzlmod/no-bzlmod and update CI

luispadron:
Fix issues doing `bazel build //...` in both bzlmod and legacy modes. Also updates CI for Bazel 8 and new dependency upgrades


MODULE.bazel
```diff
@@ -6,13 +6,13 @@ module(
     repo_name = "build_bazel_rules_swift",
 )
 
-bazel_dep(name = "bazel_features", version = "1.9.0")
+bazel_dep(name = "bazel_features", version = "1.10.0")
```
luispadron, line 9: This dependency upgrade is required because of:

```
        File "/private/var/tmp/_bazel_lpadron/0d55de22f58fc38282eca02f6f449eac/external/local_config_apple_cc/cc_toolchain_config.bzl", line 983, column 36, in _impl
                enabled = bazel_features.cc.fixed_dsym_path_quoting,
Error: 'struct' value has no field or method 'fixed_dsym_path_quoting'
```

I only noticed this when running:

```
USE_BAZEL_VERSION=7.6.1 bazel build //examples/... --noenable_bzlmod --enable_workspace
```

This is passing in bzlmod because `bazel_features` is being upgraded for us to a higher version by our other dependencies. We should consider revisiting `--check_direct_dependencies=off` @brentleyjones

aaronsky (approved):

PR #1532, opened 2025-06-04
[repro] module maps created for or on objc_library do not propagate to transitive objc_library

aaronsky:
this is a repro of an issue i've identified while trying to move to `mixed_language_library` from our custom macro. Regardless of how `module_map` is hinted in `swift_interop_info`, generated, or otherwise attached to an `objc_library`, out-of-the-box these module maps are not propagated to downstream `objc_library` in a `objc -> swift -> objc` arrangement. this is especially problematic in scenarios where PrintAsClang, going by its own uncontrollable heuristics (in this case, the addition of a lightweight generic in a public-facing objc type), inserts an `@import` line for the upstream objc library in a generated header.

this is something that needs to change in either `swift_clang_module_aspect` or `objc_library` to discover or otherwise propagate the upstream module maps, as this simple example does not work without customization of the rules or introducing manual dependencies on modulemaps (which isn't possible if relying on the generated modulemaps from the aspect).


jschear:
My understanding is that this is "working as intended" from the point of view of Bazel core -- module maps are only used for importing clang modules into Swift. Even if you had `objc_library(a)` depending on `objc_library(b)` (with a module map provided), modular imports wouldn't work: https://github.com/bazelbuild/bazel/issues/9429

I'm trying to use this intermediate rule (which pulls out the aspect-generated module map and kind of misuses `CcInfo`'s `headers` and `includes`) as a workaround, though it hasn't been battle-tested yet: https://gist.github.com/jschear/721fda69fc64efb5820855def1615bc0

`rules_swift_package_manager` has a similar approach, generating it's own modulemaps and propagating them through a CcInfo.
- https://github.com/cgrindel/rules_swift_package_manager/commit/d3f47657e54e50ce20b0785274f1515036440eaa#diff-d7638fbe2320b6f14b4515f89f0285502db480136fd78ce11c4ca2e65e231e60R56-R59
- https://github.com/cgrindel/rules_swift_package_manager/issues/208

I would support fixing this in `objc_library`,  but I'm not sure if that would be accepted. There was an attempt long ago that stalled, pre-starlarkification. https://github.com/bazelbuild/bazel/pull/6243

jschear:
I guess [rules_swift_package_manager also reimplements module map generation,](https://github.com/cgrindel/rules_swift_package_manager/blob/main/swiftpkg/internal/module_maps.bzl) it isn't the end of the world to duplicate it again 😄

PR #1533, opened 2025-06-05
Ensure undeclared outputs are deleted when not tracked

luispadron:
In rules_swift < 3.x the .swiftsourceinfo files are unconditionally written to the module path. In rules_swift >= 3.x these same files are no longer tracked by Bazel unless explicitly requested. When using non-sandboxed mode, previous builds will contain these files and cause build failures when Swift tries to use them, in order to work around this compatibility issue, we check the module path for the presence of .swiftsourceinfo files and if they are present but not requested, we remove them.

Testing:

- `bazel clean --expunge`
- `git checkout 2.8.2`
- `bazel build //examples/... --spawn_strategy=local` (pass)
- `git checkout master`
- `bazel build //examples/... --spawn_strategy=local` (failure)
- `git checkout <this-branch>`
- `bazel build //examples/... --spawn_strategy=local` (pass)

### DIFF
--- a/swift/toolchains/config/compile_config.bzl
+++ b/swift/toolchains/config/compile_config.bzl
@@ -41,6 +41,7 @@ load(
     "SWIFT_FEATURE_COVERAGE_PREFIX_MAP",
     "SWIFT_FEATURE_DBG",
     "SWIFT_FEATURE_DEBUG_PREFIX_MAP",
+    "SWIFT_FEATURE_DECLARE_SWIFTSOURCEINFO",
     "SWIFT_FEATURE_DISABLE_AVAILABILITY_CHECKING",
     "SWIFT_FEATURE_DISABLE_CLANG_SPI",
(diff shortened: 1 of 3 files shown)


swift/toolchains/config/compile_config.bzl
```diff
@@ -562,6 +563,15 @@ def compile_action_configs(
...
             ],
         ),
+
+        # Ensure that .swiftsourceinfo files are tracked and not deleted by the worker
+        ActionConfigInfo(
+            actions = [
+                SWIFT_ACTION_COMPILE,
+            ],
+            configurators = [add_arg("-Xwrapped-swift=-emit-swiftsourceinfo")],
+            features = [SWIFT_FEATURE_DECLARE_SWIFTSOURCEINFO],
```
brentleyjones, line 573: Was this change expected?

brentleyjones (approved):

PR #1535, opened 2025-06-06
Remove prints when toolchain is skipped

keith:
If you have rules_swift in your dependency tree, but don't use swift
anywhere in your build, this still warns is too noisy. Likely the most
common case for this is because grpc depends on rules_apple.


brentleyjones:
> grpc depends on rules_apple

What an odd depedency.

PR #1547, opened 2025-06-26
Use `ctx.coverage_instrumented` to detect if a target should be instrumented

brentleyjones:


brentleyjones:
Not sure if `configure_features` is the best location for this, but I basically needed a `features_for_build_modes` that applied at the target level.

PR #1550, opened 2025-07-09
Fix .swiftsourceinfo readonly bugs in version 3+

luispadron:
Similar to #1533 but when using `--strategy=worker`

In rules_swift < 3.x the .swiftsourceinfo files are unconditionally written to the module path. In rules_swift >= 3.x these same files are no longer tracked by Bazel unless explicitly requested. When using non-sandboxed mode, previous builds will contain these files and cause build failures when Swift tries to use them, in order to work around this compatibility issue, we check the module path for the presence of .swiftsourceinfo files and if they are present but not requested, we remove them.

Testing:

- `bazel clean --expunge`
- `git checkout 2.8.2`
- `bazel build @com_github_apple_swift_argument_parser//... --strategy=worker,local --worker_sandboxing=false` (pass)
- `git checkout master`
- `bazel build @com_github_apple_swift_argument_parser//... --strategy=worker,local --worker_sandboxing=false` (failure)
- `git checkout <this-branch>`
- `bazel build @com_github_apple_swift_argument_parser//... --strategy=worker,local --worker_sandboxing=false` (pass)

### DIFF
--- a/tools/worker/swift_runner.cc
+++ b/tools/worker/swift_runner.cc
@@ -27,6 +27,14 @@ bool ArgumentEnablesWMO(const std::string &arg) {
          arg == "-force-single-frontend-invocation";
 }
 
+bool StripPrefix(const std::string &prefix, std::string &str) {
+  if (str.find(prefix) != 0) {
+    return false;
+  }
+  str.erase(0, prefix.size());
+  return true;
+}
+
 namespace {
 
 // Creates a temporary file and writes the given arguments to it, one per line.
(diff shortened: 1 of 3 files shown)


brentleyjones (approved):

aaronsky (approved):

jpsim:
This appears to have broken relative path formatting in Periphery: https://github.com/peripheryapp/periphery/blob/master/Sources/PeripheryKit/Results/OutputFormatter.swift#L79-L81

Not sure what project needs to be adjusted. cc @ileitch

Issue #1529, opened 2025-05-30
Generated .dSYM doesn't contain debug info when using swift.full_lto & swift.emit.bc features


sanju-naik:
## Problem

I am trying to enable `swift.full_lto & swift.emit.bc` to experiment around the app size reduction, but looks like when these flags are enabled, the generated `.dSYM` file doesn't contain `debug info`.  So, when crashes are symbolicated using this dSYM, they don't display file names and line numbers in the crash log. 

I have verified that the generated `.dSYM` is missing `__debug_info` & `__debug_line` using `dwarfdump --show-section-sizes //path-to.dSYM`

Here is the dwarfdump output from my local 

```
dwarfdump --show-section-sizes /private/var/tmp/_bazel_sanju/128aa60197c68471680d99b52b191635/execroot/_main/bazel-out/ios_sim_arm64-opt-ios-sim_arm64-min14.0-applebin_ios-ST-216e50c47d69/bin/BazelDemoHelloWorld.app.dSYM
----------------------------------------------------
file: /private/var/tmp/_bazel_sanju/128aa60197c68471680d99b52b191635/execroot/_main/bazel-out/ios_sim_arm64-opt-ios-sim_arm64-min14.0-applebin_ios-ST-216e50c47d69/bin/BazelDemoHelloWorld.app.dSYM/Contents/Resources/DWARF/BazelDemoHelloWorld
----------------------------------------------------
SECTION           SIZE (b)
----------------  --------
__debug_abbrev           1 (0.00%)
__debug_str              1 (0.00%)
__apple_namespac        36 (0.17%)
__apple_names           36 (0.17%)
__apple_types           48 (0.23%)
__apple_objc            36 (0.17%)

 Total Size: 158  (0.77%)
 Total File Size: 20638
----------------------------------------------------
```

## Expected behaviour

The generated .dSYM should contain valid debug info so that crashes can be symbolicated correctly with file names and line numbers. 

This works as expected when it's done on a vanilla Xcode project by passing `-Xfrontend -lto=llvm-full -Xfrontend --emit-bc` in `OTHER_SWIFT_FLAGS` . 

Here is the dwarfdump output of .dSYM generated from Xcode 

```
dwarfdump --show-section-sizes /Users/sanju/Downloads/DerivedData/BazelDemoHelloWorld-fdzzlllmmnlyfofgcchicknpofpz/Build/Products/Release-iphonesimulator/BazelDemoHelloWorld.app.dSYM/
[...]

PR #1552, opened 2025-07-21
[Bugfix] Pass `-object_path_lto` option to linker when LTO flags are enabled

sanju-naik:
This is needed otherwise, build will produce an invalid dSYM when using : 
- `swift.emit.bc` or `swift.full_lto` or `swift_thin_lto` flags. 

 More details in this issue - https://github.com/bazelbuild/rules_swift/issues/1529

closes - https://github.com/bazelbuild/rules_swift/issues/1529

### DIFF
--- a/swift/internal/linking.bzl
+++ b/swift/internal/linking.bzl
@@ -30,6 +30,8 @@ load(
     ":feature_names.bzl",
     "SWIFT_FEATURE_LLD_GC_WORKAROUND",
     "SWIFT_FEATURE_OBJC_LINK_FLAGS",
+    "SWIFT_FEATURE_FULL_LTO",
+    "SWIFT_FEATURE_THIN_LTO",
 )
 load(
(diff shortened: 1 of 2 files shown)


keith:
hmm i think this should probably live in apple_support and be enabled when the `lto` feature is enabled there?

sanju-naik:
I guess we need to make this change in this file - https://github.com/bazelbuild/apple_support/blob/master/crosstool/cc_toolchain_config.bzl? . Can you help point out how and where `rules_swift`  invokes `appple_support`? Or is it done by [bazel](https://github.com/bazelbuild/bazel)? Couldn't find reference to `cc_toolchain_config.bzl` file in both repos, though.

keith:
yea it happens by bazel when calling cc_common.link but the tl;dr is that enabled feature names flags are added automatically when linking

PR #1560, opened 2025-08-23
Add support for compact repo mapping manifest to the runfiles library

cerisier:
Implements https://github.com/bazelbuild/bazel/issues/26262

This implementation follows what others like rules_java have done. Note that the implementation is O(prefix). If performance is an issue, we could probably make that a better with a Trie to replicate NavigableMap.


cerisier:
Anyone for a review now that 8.4.0 is out ?

PR #1576, opened 2025-09-26
Unconditionally pass -file-prefix-map to the compiler to remove nondeterministic paths.

luispadron:
Simplified the `debug_settings` tests since the rules unconditionally pass `-file-prefix-map` now. We may want to remove the support for `-debug-prefix-map` in the future since it's redundant.

Cherry-pick: a440c7ebbc5a0a30acb3886775a63a1ffa9d6c39


luispadron:
Need to rebase this. Only the second commit is relevant

brentleyjones (approved):

luispadron:
@brentleyjones This can merge now

PR #1583, opened 2025-09-27
Explicitly pass .swiftmodule and .swiftinterface files as action inputs

adincebic:
Previously when building `swift_import` targets that depend on other `swift_import` targets with `.swiftinterface` files failed when using Bazel sandbox with error:

```
  cannot open file 'swift-syntax/arm64/SwiftSyntax.private.swiftinterface' (No such file or directory)
```

The `SWIFT_ACTION_COMPILE_MODULE_INTERFACE` action was missing transitive dependency files as inputs so I made sure that they are explicitly listed.

### DIFF
--- a/swift/toolchains/config/compile_config.bzl
+++ b/swift/toolchains/config/compile_config.bzl
@@ -2102,8 +2102,19 @@ def _dependencies_swiftmodules_configurator(prerequisites, args):
         uniquify = True,
     )
 
+    # Include both swiftmodule and swiftinterface files as inputs to ensure
+    # they are available in the sandbox for compilation
+    transitive_inputs = []
+    for module in prerequisites.transitive_modules:
+        swift_module = module.swift
+        if swift_module:
+            if swift_module.swiftmodule:
+                transitive_inputs.append(swift_module.swiftmodule)
+            if swift_module.swiftinterface:
+                transitive_inputs.append(swift_module.swiftinterface)
+
     return ConfigResultInfo(
-        inputs = prerequisites.transitive_swiftmodules,
+        inputs = transitive_inputs,
     )
 
 def _module_aliases_configurator(prerequisites, args):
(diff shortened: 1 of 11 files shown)


swift/toolchains/config/compile_config.bzl
```diff
@@ -2006,8 +2005,19 @@ def _dependencies_swiftmodules_configurator(prerequisites, args):
...
+        swift_module = module.swift
+        if swift_module:
+            if swift_module.swiftmodule:
+                transitive_inputs.append(swift_module.swiftmodule)
+            if swift_module.swiftinterface:
+                transitive_inputs.append(swift_module.swiftinterface)
+
     return ConfigResultInfo(
-        inputs = prerequisites.transitive_swiftmodules,
+        inputs = transitive_inputs,
```
brentleyjones, line 2020: We should make it so `prerequisites.transitive_swiftmodules` has the right things (and probably rename it to `transitive_included_modules`... though maybe we don't do that for now to keep cherry-picks easier). Also, we should only add one or the other. We should add `swift_module.swiftmodule` if it exists, otherwise `swift_module.private_swiftinterface`, otherwise `swift_module.swiftinterface`.

adincebic: @brentleyjones thanks for pushing back. This is resolved now.

brentleyjones (approved):

swift/toolchains/config/compile_config.bzl
```diff
@@ -2077,8 +2087,19 @@ def _dependencies_swiftmodules_vfsoverlay_configurator(prerequisites, args, is_f
         "-I{}".format(prerequisites.vfsoverlay_search_path),
     )
 
+    # Include both swiftmodule and swiftinterface files as inputs to ensure
+    # they are available in the sandbox for compilation
+    transitive_inputs = [prerequisites.vfsoverlay_file]
```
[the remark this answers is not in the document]
adincebic: Yes, covered that case too.

PR #1585, opened 2025-10-07
Enable support for AEGs

brentleyjones:
(cherry picked from commit 8a4310e85b1f9e01658b871ac96c5a29a5c40dce)

Resolves https://github.com/bazelbuild/rules_swift/issues/1365.

### DIFF
--- a/swift/swift_binary.bzl
+++ b/swift/swift_binary.bzl
@@ -14,7 +14,6 @@
 
 """Implementation of the `swift_binary` rule."""
 
-load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
 load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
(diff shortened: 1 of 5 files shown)


brentleyjones:
We will need to apply https://github.com/bazelbuild/rules_apple/commit/3da7b31bcb8c4190f13adce9ef289ee0ff923979 in **rules_apple** before landing this.

brentleyjones:
Also looks like we need to set toolchains for the proto stuff. Upstream defines `swift_proto_toolchain`, and I think we have to do something similar. cc: @AttilaTheFun.

AttilaTheFun:
@brentleyjones I'm not familiar with Automatic Extension Groups but you're welcome to modify the swift_proto_library rule to support them.

However, longer term I was thinking it would be more maintainable to refactor them into a macro with a source-generating rule which passes the generated sources into a regular swift_library.

This is the approach I took at Snap for our new Swift dependency injection system and it has been working well for us so far.

There are a couple of tradeoffs to this approach, though, and it would be a source-breaking change so we would have to retain the old behavior with some kind of legacy_mode flag.

luispadron (approved):

PR #1586, opened 2025-10-07
Move the C++ toolchain into its own execution group so that it does not have to be bound by the same constraints as the execution platform selected for the Swift toolchain.

brentleyjones:
PiperOrigin-RevId: 745988579
(cherry picked from commit bb0f38be62004cfd5e6f90958ccc58d2b77af233)


brentleyjones:
Current failures are unrelated to PR, but I don't feel safe merging this while they are there.

PR #1589, opened 2025-10-12
Add rpaths for the toolchain's compatibility back-deployment librarie

adincebic:
This is necessary to ensure that these binaries run correctly when using APIs like `Span` that are back-deployed on older OSes. Since the `swift_*` rules don't do bundling of their own, we simply point to the libraries in the toolchain itself (using the `xcode-select`-dependent symlinks, as we were already doing for tests). To actually distribute such a binary, it's the responsibility of the user to include the required dylibs with it (or, better, use a rule from rules_apple to perform the bundling).

I've explicitly ignored the swift-5.0 and swift-5.5 directory since we have no plans to deploy anything from these rules to such old OSes.

Cherry pick: [2f5aa6e6a3dae09293c9b35ff3cc5795fef279e7](https://github.com/bazelbuild/rules_swift/commit/2f5aa6e6a3dae09293c9b35ff3cc5795fef279e7)

### DIFF
--- a/swift/toolchains/xcode_swift_toolchain.bzl
+++ b/swift/toolchains/xcode_swift_toolchain.bzl
@@ -108,7 +108,6 @@ load(
 )
 load("//swift/toolchains/config:tool_config.bzl", "ToolConfigInfo")
 
-
 _CPP_TOOLCHAIN_TYPE = Label("@bazel_tools//tools/cpp:toolchain_type")
 
 visibility("public")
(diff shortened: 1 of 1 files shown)


luispadron (approved):

luispadron:
@adincebic thanks for updating, mind fixing the conflict and merging?

qyang-nj:
It looks like this change unconditionally adds the following rpaths to all binaries, including iOS binaries:
```
/private/var/select/developer_dir/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-6.2/iphoneos
/var/db/xcode_select_link/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-6.2/iphoneos
```
These paths refer to the build machine’s Xcode installation and do not exist on iOS devices. I'm wondering if it's possible to omit these rpaths for iOS binaries.

keith:
@qyang-nj can you test with https://github.com/bazelbuild/rules_swift/pull/1864?

PR #1624, opened 2025-12-03
use a perl script for mixed_language_module_map_extender

alex-ameen:
Optimizes existing `mixed_language_module_map_extender` to avoid repeatedly spinning up `perl` runtimes by simply making it a `perl` script.

We experienced performance issues with this script in our repos and are hoping this will speed things up.

Bazel profiles showed a small improvement on ~2,000 invocations of the tool, 258s to 230s. Nothing too dramatic, but a nice ~10% improvement.


alex-ameen:
@luispadron run it bro

tools/mixed_language_module_map_extender/mixed_language_module_map_extender.pl
```diff
@@ -0,0 +1,90 @@
...
+              "<swift_generated_header>\n";
+    if ( @ARGV == 1 )
+      {
+        print STDERR $msg;
+        exit 0;
+      }
+    else
+      {
+        die $msg;
+      }
```
keith:
```suggestion
    die $msg;
```

probably better to be careful and not ever exit 0 with incorrect usage?

keith (approved):

PR #1645, opened 2026-01-16
Add support for `minimum_os_version` in Swift rules

luispadron:


swift/toolchains/xcode_swift_toolchain.bzl
```diff
@@ -448,8 +496,10 @@ def _all_action_configs(
                 SWIFT_ACTION_SYNTHESIZE_INTERFACE,
             ],
             configurators = [
-                add_arg("-target", target_triples.str(target_triple)),
-                add_arg("-sdk", apple_toolchain.sdk_dir()),
+                _make_target_triple_configurator(
+                    base_target_triple = target_triple,
+                    sdk_dir = apple_toolchain.sdk_dir(),
```
keith, line 501: looks like `-sdk` shouldn't be part of this one? it's just passed through?

PR #1670, opened 2026-03-29
Avoid emitting `-LNone` in `developer_dirs_linkopts` on Linux

bc-lee:
When `developer_dirs` is empty, `swift_developer_lib_dir(...)` returns `None`. `developer_dirs_linkopts(...)` was still formatting that value as a library search path, which produced `-LNone` in the final link command.

Fix this by only adding the `-L...` flag when a real Swift developer library directory exists, and add unit tests covering empty, non-platform, and platform developer dir cases.

Fixes #1669


keith (approved):

keith:
thanks!

PR #1679, opened 2026-04-13
3.x: Stop testing bazel 7 on BCR and enable proto toolchain resolution

adincebic:
Unblocking 3.x releases.

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -5,7 +5,6 @@ shell_commands: &shell_commands
 
 matrix:
   bazel:
-  - 7.x
   - 8.x
   # TODO: Re-enable once >10.0.0-pre.20251105.2 is available https://github.com/bazelbuild/bazel/issues/27749
   # - rolling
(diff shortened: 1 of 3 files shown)


.bcr/presubmit.yml
```diff
@@ -5,7 +5,6 @@ shell_commands: &shell_commands
 
 matrix:
   bazel:
-  - 7.x
   - 8.x
   # TODO: Re-enable once >10.0.0-pre.20251105.2 is available https://github.com/bazelbuild/bazel/issues/27749
   # - rolling
```
aaronsky: can we also add 9.x + restore rolling testing

adincebic: Sure. Need to check if 3.x builds on rolling first.

luispadron (approved):

.bcr/presubmit.yml
adincebic: Done.

aaronsky (approved):

PR #1681, opened 2026-04-17
Remove _cc_toolchain attr

keith:
Shouldn't be necessary with our uses of `use_cpp_toolchain`

Cherry picked from 319d37f06f9f1e2dc1debf572fa7a9d72cbad36e


aaronsky (approved):
mind cherry-picking this to 3.x too?

keith:
does it have a meaningful behavior difference?

aaronsky:
i guess not

PR #1686, opened 2026-04-20
Add swift.static_stdlib feature to link Swift runtime statically

mauriciogg:
TODO(#8) in swift/toolchains/swift_toolchain.bzl.

Introduces SWIFT_FEATURE_STATIC_STDLIB, which, when enabled resolves the standard library and runtime (including swiftrt.o) from the toolchain's `lib/swift_static/{os}` tree instead of the shared `lib/swift/{os}` tree and drops the shared-library rpath


keith:
I'm definitely for this feature, it would be great if you could add a sh_test in the end that verified that some binary was linked statically

jhodsdon:
@keith is it possible to have this in the 4.0.0 release? I have some immediate uses for this and would love to start using it :) and thank you @mauriciogg for pushing this up

keith:
yea should be fine. seems simple enough. we're changing a ton of toolchain stuff this release so it'll be a bit easier to test once that is all done

PR #1687, opened 2026-04-20
Add explicit module support

### LINKED ISSUE #1475: [Feature Request] Explicit clang modules
**Context** 

[Explicit module for swift](https://developer.apple.com/documentation/xcode/building-your-project-with-explicit-module-dependencies) helps with build speed by reducing the duplicated pcm file generation (via compilation). Implicit clang modules are used for Swift / Obj-C interop most commonly. 

Build speed improvement comes from 1) not having to generate duplicate pcm files and 2) pcm files become explicit outputs of bazel compile actions making Bazel's action scheduling more efficient. 

In Snapchat, we used to use clang modules for all objective-c library targets by default for Swift inter-op. But we realized the overhead from implicit clang modules, and changed to creating clang modules only when objective-c library is depended on by swift target. As a result, we were able to see ~50% improvement in clean full application build time (~40 --> ~20min).

keith:
This adds a v1 of explicit module support. The primary focus for this
version is getting the rules wired up to discover the modules in the
SDK and then generate pcm files for the clang modules. Notably this
does not automatically surface the prebuilt `swiftmodule` files from the
SDK, it's still up to the compiler to find those automatically.

This generally works by:

- For each Xcode version
  - For each SDK
    - Generate BUILD file with the entire module dependency tree (based on `swiftc -scan-dependencies`)
  - Alias those back to select the right SDK for a given module
- Alias those back to select the right Xcode version based on `--xcode_version`
- Add the relevant modules to the `deps` of all your targets-
- Pass `--features=swift.use_c_modules --features=swift.emit_c_module --host_features=swift.use_c_modules --host_features=swift.emit_c_module`

Fixes https://github.com/bazelbuild/rules_swift/issues/1475


keith:
todos before merging (beside testing in the real world):

- [x] pass everything by default, don't require users add their own
- [x] finalize naming (happy to take review suggestions)
- [x] add a module tag for limiting to specific sdks to save generation time and file size
- [x] make sure pcms are hermetic enough
- [x] test how they work with objc_library
- [x] passing `-application-extension` to some targets makes them incompatible with the pcms
- [x] pass `-disable-building-interface`
- [x] figure out absolute paths in Testing.swiftmodule that is produced
- [x] test how they work with lldb
- [x] potentially fix slow analysis by reducing aspect calls
- [x] add target for users to run the tool with a given xcode path so they can generate their own configs
- [x] verify the apple_sdk_clang_module rule is good enough for v1 (this was pulled back up from swift_c_modules with no changes)
- [x] docs
- [ ] ~I have to enable the feature swift.headers_always_action_inputs  because the .swiftinterface  of an apple_static_xcframework_import target isn't propagated down.~ need to figure out in rspm
- [ ] ~MiSnap cross-xcframework deps: MiSnap.rspm is a binary xcframework whose headers include headers from MiSnapCore, MiSnapCamera, and MiSnapScience. Without declaring these as deps, PCM precompilation can’t resolve the transitive headers. Requires the post-BUILD patch fix since these deps must be added to the generated BUILD file. Fix: Add MiSnapCore.rspm, MiSnapCamera.rspm, and MiSnapScience.rspm to the deps of MiSnap.rspm. (I still don't understand why it's only necessary for explicit module build)~ this requires additional deps added by rspm
- [ ] ~what linux support do we need for v1~ not blocking on this, need to do some work to the scanner for linux
- [ ] ~dep graphs changing on `-D` flags does not work for sure, no clear path to that~
- [ ] ~figure out rspm deps~ needs to be done, not blocking on that for this v1
- [ ] ~As expected, precompile_clang_module fails in RBE for me. One issue is the different xcode location. I still need to figure out if there are other issues and test out providing own BUILD file to apple_sdk . I also need to test it out on Linux machine.~
- [ ] ~add explicit deps to things we automatically add for users, like test_observer~, bootstrapping problems, have to merge and add to some transitive deps first
- [x] tools/explicit_modules/scan.py  also needs to scan C library modules in usr/include/ (e.g., CommonCrypto, zlib, SQLite3) and modules declared via extern module in the top-level modulemap were missed.
- [x] Racing in PCM hermetic runner: actions targeting different SDKs (e.g., iPhoneSimulator + MacOSX for swift compiler plugin) raced over the same symlink names, and the remove-then-create pattern had a TOCTOU race between parallel actions for the same SDK.
- [x] PCM precompilation for mixed_language_library : need to update mixed_language_library to support PCM precompilation, and also pass precompiled modules to objc_library and mixed_language_library  in mixed_language/mixed_language_library.bzl
- [x] When swift.add_default_precompiled_modules is on, we still need to pass precompiled modules to the swift_compiler_plugin.
- [x] something wrong with stack protector mismatches https://bazelbuild.slack.com/archives/C09TMF7SK5M/p1777322541039219
- [x] Make sure CommonCrypto works on all supported platforms, something about it was missing on iOS I think
- [x] need swiftmodule debugging options on, and a test validating they don't contain absolute paths
- [x] silencing these warnings: 

```./__bazel_pcm_developer_dir/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.4.sdk/System/Library/Frameworks/BrowserEngineKit.framework/Headers/BEDownloads.h:15:1: warning: imported declaration 'BEDownloadMonitorLocation' could not be mapped to 'BEDownloadMonitor.Location' [#ClangDeclarationImport]
13 | NS_ASSUME_NONNULL_BEGIN
14 |
15 | NS_SWIFT_NAME(BEDownloadMonitor.Location)
   | |- warning: imported declaration 'BEDownloadMonitorLocation' could not be mapped to 'BEDownloadMonitor.Location' [#ClangDeclarationImport]
   | `- note: please report this issue to the owners of 'BrowserEngineKit'
16 | NS_SWIFT_SENDABLE
17 | API_AVAILABLE(ios(18.2))

[#ClangDeclarationImport]: <https://docs.swift.org/compiler/documentation/diagnostics/clang-declaration-import>
```

- [x] #if SWIFT_PACKAGE not propagated to PCM: SPM packages use #if SWIFT_PACKAGE guards to choose between bare imports and framework-style imports. SWIFT_PACKAGE is defined via copts (not defines), and copts are NOT propagated through CcInfo.compilation_context to the PCM action. Headers fall through to framework-style imports that fail. Fix: patches the package with command like "find . -name '*.h' -exec sed -i '' 's/#if SWIFT_PACKAGE/#if 1/g' {} +"
- [x] copts are not propagated to PCM precompilation. For example, the swift package [Aardvark](https://github.com/square/Aardvark/blob/6f0d61048902dc39e9a8272bbeccb7d6bf79ba1b/Sources/CoreAardvarkSwift/Logging.swift#L19) checks #if SWIFT_PACKAGE .  rules_swift_package_manager [adds -DSWIFT_PACKAGE to copts](https://github.com/cgrindel/rules_swift_package_manager/blob/b5257d3dced5f259092c3831a463da68382da62b/swiftpkg/internal/swiftpkg_build_files.bzl#L112-L118) and is able to build correctly in implicit module build, but it fails in explicit module build.
- [x] swift-nio C macro visibility: In explicit module builds, C macros from system headers (IP_RECVPKTINFO, IP_PKTINFO) are not visible through user module namespaces. NIOPosix accesses CNIODarwin.IP_RECVPKTINFO which fails because these are preprocessor macros, not constants.
Fix: it needs to patch for a bunch of C marcos like "grep -q 'CNIODarwin_IP_RECVPKTINFO' Sources/CNIODarwin/include/CNIODarwin.h || sed -i '' '/extern const int CNIODarwin_IPV6_PKTINFO;/a\\\nextern const int CNIODarwin_IP_RECVPKTINFO;\\\nextern const int CNIODarwin_IP_PKTINFO;\n' Sources/CNIODarwin/include/CNIODarwin.h" 
- [x] swift-perception and swift-composable-architecture use @available(iOS, obsoleted: ...) on backport types (e.g., Bindable). With Xcode 26 SDK, obsoleted: makes these types completely unavailable, causing compile errors in any code that references them. Fix: patch the package with commands like "sed -i '' 's/obsoleted:/deprecated:/g' Sources/PerceptionCore/SwiftUI/Bindable.swift". (Not sure why it starts failing for explicit module build)

keith:
for simple testing you can add this to the root BUILD file:

```bzl
load("//swift:swift_binary.bzl", "swift_binary")

swift_binary(
    name = "foo",
    srcs = ["main.swift"],
    deps = [
        "@apple_sdk//:Foundation",
        "@apple_sdk//:SwiftShims",
    ],
)
```

```
// main.swift
import Foundation

let a: NSString = "real foundation api"
print("hi", a)
```

and then run:

```
bazel run :foo --features=swift.use_c_modules --features=swift.emit_c_module --features=swift.use_explicit_swift_module_map
```

swift/internal/apple_sdk_clang_module.bzl
```diff
@@ -0,0 +1,191 @@
...
+"""Implementation of the `apple_sdk_clang_module` rule."""
+
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
+load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@build_bazel_rules_swift//swift:providers.bzl", "SwiftInfo", "create_clang_module_inputs", "create_swift_module_context")
+load("@build_bazel_rules_swift//swift:swift_common.bzl", "swift_common")
+load("@build_bazel_rules_swift//swift/internal:compiling.bzl", "precompile_clang_module")
+load("@build_bazel_rules_swift//swift/internal:feature_names.bzl", "SWIFT_FEATURE_SYSTEM_MODULE")
+load("@build_bazel_rules_swift//swift/internal:toolchain_utils.bzl", "SWIFT_TOOLCHAIN_TYPE")
+load("@build_bazel_rules_swift//swift/internal:utils.bzl", "merge_runfiles")
```
aaronsky:
```suggestion
load("//swift:providers.bzl", "SwiftInfo", "create_clang_module_inputs", "create_swift_module_context")
load("//swift:swift_common.bzl", "swift_common")
load("//swift/internal:compiling.bzl", "precompile_clang_module")
load(":feature_names.bzl", "SWIFT_FEATURE_SYSTEM_MODULE")
load(":toolchain_utils.bzl", "SWIFT_TOOLCHAIN_TYPE")
load(":utils.bzl", "merge_runfiles")
```

tools/worker/pcm_hermetic_runner.cc
```diff
@@ -0,0 +1,213 @@
...
+
+std::string GetEnv(const char* name) {
+  const char* value = std::getenv(name);
+  return (value != nullptr && value[0] != '\0') ? std::string(value)
+                                                : std::string();
+}
+
+void ReplaceAll(std::string& arg, const std::string& needle,
+                const std::string& replacement) {
+  if (needle.empty()) return false;
```
congt: ReplaceAll is a void function, and should not return false

keith: yea sorry i pushed while still iterating, fixed

doc/explicit_modules.md
```diff
@@ -0,0 +1,248 @@
...
+ --developer-dir \
+ /Applications/Xcode-26.4.0-RC1.app/Contents/Developer \
+ MacOSX iPhoneOS iPhoneSimulator
+```
+
+NOTE: The format of this `BUILD` file isn't considered stable, so if you
+choose to vendor it yourself you should regenerate it whenever you
+update `rules_swift`.
+
+## Handling `-application-extension`
```
aaronsky, line 167: Is this a temporary instruction?

tools/explicit_modules/xcode_explicit_module_repo.bzl
```diff
@@ -0,0 +1,92 @@
...
+            args.extend(["--exclude-module", "{}:{}".format(sdk, module)])
+    return args
+
+def _xcode_explicit_module_repo_impl(rctx):
+    developer_dir = _resolve_developer_dir(rctx)
+    rctx.report_progress("Scanning SDKs for Xcode {}".format(rctx.attr.xcode_version))
+    rctx.watch(rctx.attr._script)
+    result = rctx.execute(
+        [
+            "/usr/bin/python3",
```
aaronsky, line 44: shame it isn't easier to request the configured hermetic toolchain in here

.pre-commit-config.yaml
```diff
@@ -7,3 +7,10 @@ repos:
     hooks:
       - id: buildifier
       - id: buildifier-lint
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
aaronsky, line 11: ah, bless. let's add this to rules_apple too

MODULE.bazel
```diff
@@ -45,6 +45,45 @@ use_repo(apple_cc_configure, "local_config_apple_cc")
 
 register_toolchains("//swift/toolchains:all")
 
+system_sdk = use_extension("//swift:extensions.bzl", "system_sdk")
+system_sdk.configure_sdks(
+    # NOTE: This doesn't apply to downstream repos
```
aaronsky, line 50: doesn't it? or am I mistaken. i thought the `use_extension` needs `dev_dependency = True` in order for this to be true

test/hermetic_pcm/xcode_version_at_least.bzl
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+def _xcode_version_at_least_impl(ctx):
+    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+    version = xcode_config.xcode_version()
+    threshold = apple_common.dotted_version(ctx.attr.minimum_version)
+    greater_or_equal = version != None and version >= threshold
+    return [config_common.FeatureFlagInfo(value = "True" if greater_or_equal else "False")]
+
+xcode_version_at_least = rule(
```
aaronsky, line 12: can we not pull this from apple_support? or is that just a helper function

aaronsky (approved):
some small comments, nothing blocking

doc/explicit_modules.md
keith: I don't think so. unless we get dynamic dependencies in the future i think this will have to stay. unless we find some major issue with this disabling the important behavior of this and need to find another path.

MODULE.bazel
keith: the extension is explicitly checking if it's the root module or not and otherwise ignoring it. we could loosen that in the future but i think that's probably fine for clarity

test/hermetic_pcm/xcode_version_at_least.bzl
keith: yea just a helper function, in this case i need it to end up as a target i can throw in `target_compatible_with` conditions

PR #1691, opened 2026-04-21
Add minimum_os_version attribute to Swift rules

jpsim:
Adds a `minimum_os_version` string attribute on `swift_binary`, `swift_library`, `swift_test`, `swift_compiler_plugin`, `swift_overlay`, and `swift_proto_library`. On Apple/Xcode toolchains, when set, the value is used to construct the OS component of the target triple passed to `swiftc` (for example, `apple-macos13.0`). When unset, the target triple from the selected platform configuration is used unchanged.

Also exposes `minimum_os_version` on `swift_common.compile` and adds a `target_triples.with_os_version` helper for constructing versioned triples.

This is a simpler alternative to #1645, which depended on upstream Bazel work to expose `XcodeSdkVariantInfo` via `xcode_sdk_variant` rules. That work has stalled while the Xcode rules migrate out of Bazel core into apple_support, so this change overrides the OS version directly in the Xcode toolchain without depending on new Bazel providers.

Refs:
* https://github.com/bazelbuild/rules_swift/pull/1645
* https://github.com/cgrindel/rules_swift_package_manager/issues/892
* https://github.com/bazelbuild/bazel/pull/21991
* https://github.com/bazelbuild/bazel/pull/28028


luispadron (approved):

jpsim:
The failures on "last green" ci jobs are unrelated

PR #1694, opened 2026-04-23
Add a new `generated_swiftinterfaces` parameter to

thePsguy:
Update`swift_library` to explicitly declare additional
`.swiftinterface` files as output.

This will allow `private` and `package` swift interfaces to be declared as outs, consequently allowing them to be included in frameworks built with such a library.

This also unblocks usage of @_spi


adincebic:
Producing `.swiftinterface` files is already supported through existing features. When building in library evolution mode, you can pass `--features=swift.enable_library_evolution`. If needed, `.swiftinterface` generation can also be requested explicitly via `--features=swift.emit_swiftinterface`, and similarly `.private.swiftinterface` via `--features=swift.emit_private_swiftinterface`.

Given these options, it seems we already have sufficient mechanisms in place, so adding a separate attribute for the same behavior may not be necessary.

thePsguy:
@adincebic  They're produced but not part of the declared outs despite library evolution being enabled. Is there another way to include them in the build graph?

PR #1695, opened 2026-04-24
Add flag_aliases to MODULE.bazel

keith:


keith:
this doesn't work yet, adding as an example

keith:
we can add if it ever works

PR #1697, opened 2026-04-25
Fix new warning when building modules

keith:
```
INFO: From Compiling Swift module //test/fixtures/module_interface/library:dependent_toy_module_library:
  <unknown>:0: warning: emitting module interface files requires '-language-mode'
```

### DIFF
--- a/swift/toolchains/config/compile_config.bzl
+++ b/swift/toolchains/config/compile_config.bzl
@@ -1314,6 +1314,18 @@ def compile_action_configs(
                 SWIFT_FEATURE__SUPPORTS_UPCOMING_FEATURES,
             ],
         ),
+        # Swift 6.2+ warns building swiftmodule files without any -language-mode
+        ActionConfigInfo(
+            actions = [
+                SWIFT_ACTION_COMPILE,
+                SWIFT_ACTION_DERIVE_FILES,
+            ],
+            configurators = [add_arg("-swift-version", "5")],
+            not_features = [
+                SWIFT_FEATURE_ENABLE_V6,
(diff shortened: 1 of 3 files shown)


swift/toolchains/config/compile_config.bzl
```diff
@@ -1314,6 +1314,18 @@ def compile_action_configs(
                 SWIFT_FEATURE__SUPPORTS_UPCOMING_FEATURES,
             ],
         ),
+        # Swift 6.2+ warns building swiftmodule files without any -language-mode
+        ActionConfigInfo(
+            actions = [
+                SWIFT_ACTION_COMPILE,
+                SWIFT_ACTION_DERIVE_FILES,
+            ],
+            configurators = [add_arg("-swift-version", "5")],
```
aaronsky, line 1323: i think this conflicts with anyone setting it in their copts, yes? or folks still setting the 4.2 language mode

keith: your `--swiftcopt` should win over this default if you're doing it manually

aaronsky: i tend to set this in the rule’s copts, so my macros can offer a `swift_version` attr

keith: yea that should win over this too

adincebic (approved):

swift/toolchains/config/compile_config.bzl
keith: pushed tests covering the order

PR #1698, opened 2026-04-28
Fix is exec config for 9.0.0

keith:
https://github.com/bazelbuild/bazel/issues/29421


keith:
needs https://github.com/bazel-contrib/bazel_features/pull/141 in a bazel_features release

adincebic (approved):

PR #1699, opened 2026-04-29
Add linux support to tools/swift-releases

keith:
Not super important but wasn't too hard

### DIFF
--- a/tools/swift-releases/ToolchainDownloader.swift
+++ b/tools/swift-releases/ToolchainDownloader.swift
@@ -1,6 +1,13 @@
 import ArgumentParser
-import CryptoKit
 import Foundation
+
+#if canImport(CryptoKit)
+  import CryptoKit
+#endif
+
+#if canImport(FoundationNetworking)
+  import FoundationNetworking
+#endif
 
 extension FileHandle: @retroactive TextOutputStream {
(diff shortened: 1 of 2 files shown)


tools/swift-releases/ToolchainDownloader.swift
```diff
@@ -102,7 +109,48 @@ struct ToolchainDownloader {
...
-    return hash.compactMap { String(format: "%02x", $0) }.joined()
+    return try sha256(of: data)
+  }
+
+  private func sha256(of data: Data) throws -> String {
+    #if canImport(CryptoKit)
+      let hash = SHA256.hash(data: data)
+      return hash.compactMap { String(format: "%02x", $0) }.joined()
+    #else
+      // CryptoKit is Apple-only. On Linux, shell out to `sha256sum` (part of
```
aaronsky, line 120: we could also add swift-crypto as a dependency. i wouldn't mind

keith: yea i looked at that, i didn't feel like making it build all of boringssl for this though

aaronsky (approved):
god, `Foundation.Process` is such a clunky API compared to `subprocess.run` or `os/exec`

PR #1701, opened 2026-04-29
Partially fix swiftly usage

keith:
Previously our logic assumed things about the installation structure of
swift on linux that aren't true when installing with swiftly. Now we
fetch the real canonical path to the installation in any layout.

This doesn't work fully because of rules_cc issues fixed by https://github.com/bazelbuild/rules_cc/pull/706


keith:
also depends on https://github.com/bazelbuild/rules_swift/pull/1700

aaronsky (approved):

PR #1703, opened 2026-04-29
Use swiftly on CI

keith:


aaronsky:
should we be using swiftly or exercising the new hermetic toolchain? or both?

keith:
Both I think. I think the other question is if we still test system ones on Linux. I left a job with that here figuring docker containers could benefit

PR #1719, opened 2026-05-05
Make swift_clang_module_aspect grab some copts

keith:
If the underlying cc_library / objc_library target has `-D` flags that
aren't in `defines` (and therefore propagated up), they still might be
necessary if they're used in `#if` conditions in private headers that
are exercised during the pcm compile. We only grab `-D` flags, and
ideally in the future we can use only `local_defines` for this.


keith:
depends on https://github.com/bazelbuild/rules_swift/pull/1720

aaronsky (approved):

keith:
going to have to land with explicit modules

PR #1721, opened 2026-05-05
Update triple for PCM compiles

keith:
Turns out what Xcode does for Swift PCM dependencies is it uses the
version of the SDK as the target triple. This doesn't override the
actual deployment target but is passed through separately to the
compiler.

Importantly this sidesteps minimum OS version transition issues because
this version never changes throughout a single build.

Xcode currently uses your actual deployment target for Objective-C
target PCMs, but we don't support that yet so we can ignore that for
now.


keith:
i think im going to leave this with the explicit module pr since it needs more testing

PR #1727, opened 2026-05-08
Add swift_verify_test rule

keith:
This is currently only used in our tests, it runs `swiftc -frontend
-verify` which is a mode that checks the current file being compiled and
asserts things about its diagnostics. It's not really public API but it
is heavily used in the compiler tests so it feels pretty safe. This lets
us write tests against things that shouldn't compile given some
configuration.


keith:
example of why this is useful: https://github.com/bazelbuild/rules_swift/pull/1728

aaronsky (approved):

PR #1735, opened 2026-05-09
Parse /usr/include/module.modulemap for explicit modules

keith:
This file declares some modules that aren't in the dependency tree any
other place. Parsing only this 1 file shouldn't be too bad for perf.


adincebic (commented):
Does it make sense to apply exclusion filter here? Maybe it can have broken module which user would want to exclude?

aaronsky (approved):

PR #1738, opened 2026-05-11
Add example failing test with canImport

keith:
Currently `canImport` for modules that aren't in your deps returns
`true`, but then things fail because underlying clang modules aren't in
the dep tree.

### DIFF
--- a/test/fixtures/precompiled_modules/BUILD
+++ b/test/fixtures/precompiled_modules/BUILD
@@ -354,6 +354,25 @@ transition_test(
     ],
 )
 
+swift_library(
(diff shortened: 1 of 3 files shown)


keith:
This seems like a lose lose to me we can either:

1. leave this behavior, but the error message is very bad. for example in this case of missing SwiftUI the error that you see ends up being:

```
<unknown>:0: error: cannot load underlying module for 'CoreData'
<unknown>:0: error: cannot load underlying module for 'CoreData'
```

2. we could make the `canImport`  fail, but that could be worse if the user just forgot to add the `dep`

PR #1741, opened 2026-05-11
Remove absolute path to Xcode from rpaths

keith:
Based on the discussion in
https://github.com/bazelbuild/rules_swift/issues/1629 the symlinks that
we use to find Xcode via rpaths weren't available on the machine,
possibly right after an update, but I think we need to treat that as a
bad environment instead since this otherwise makes the binaries non
hermetic.

This reverts e40028884604d06faa9e9276ba05ce712f32d788


keith:
cc @marekcirkos @adincebic from the discussion on https://github.com/bazelbuild/rules_swift/issues/1629 since this does introduce the absolute path to xcode in the rpaths

aaronsky (approved):

adincebic (approved):

PR #1742, opened 2026-05-11
Handle cross import overlays from transitive deps

keith:
Unless you enforce layering_check for targets, you could rely on cross
import overlays from any of your transitive deps.

This is very not ideal, but otherwise this results in subtle failures.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -1296,18 +1296,20 @@ def _cross_imported_overlays(
 
     # Build a "set" containing the module names of direct dependencies so that
     # we can do quicker hash-based lookups below.
-    direct_module_names = {}
+    module_names = {}
     for swift_info in user_swift_infos:
-        for module_context in swift_info.direct_modules:
-            direct_module_names[module_context.name] = True
+        # TODO: Ideally this would only be the direct dependencies, but unless
+        # you enforce layering_check it's easy to rely on this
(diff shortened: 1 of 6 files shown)


keith:
i don't like having to do this so i still wonder if we should merge it. this affects the case where you opt-in to managing @system_sdk deps yourself, but you miss one on a library. But besides odd compiler failures you can't obviously tell what went wrong. You'll see something like "missing module `_Testing_AppKit`"

swift/internal/compiling.bzl
```diff
@@ -1296,18 +1296,20 @@ def _cross_imported_overlays(
...
     # Build a "set" containing the module names of direct dependencies so that
     # we can do quicker hash-based lookups below.
-    direct_module_names = {}
+    module_names = {}
     for swift_info in user_swift_infos:
-        for module_context in swift_info.direct_modules:
-            direct_module_names[module_context.name] = True
+        # TODO: Ideally this would only be the direct dependencies, but unless
+        # you enforce layering_check it's easy to rely on this
+        for module_context in swift_info.transitive_modules.to_list():
```
keith, line 1303: doing this is the bad part

aaronsky (approved):

swift/internal/compiling.bzl
keith: maybe this shouldn't feel too bad because we already do a lot of flattening of this for the json file and other places

PR #1747, opened 2026-05-14
Use developer dir symlink for swiftinterfaces

keith:
Previously every swiftinterface being built would race for the same
symlink dirs. Now we use the same approach as PCMs to avoid that race
(which only happens with sandboxing disabled).

### DIFF
--- a/swift/internal/system_swiftinterface.bzl
+++ b/swift/internal/system_swiftinterface.bzl
@@ -39,6 +39,10 @@ def _direct_clang_module_for_name(deps, module_name):
 
 def _system_swiftinterface_impl(ctx):
     swift_toolchain = swift_common.get_toolchain(ctx, toolchain_type = SWIFT_SDK_TOOLCHAIN_TYPE)
+    if ctx.attr.system_swiftinterface and ctx.file.swiftinterface:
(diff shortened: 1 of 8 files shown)


swift/internal/system_swiftinterface.bzl
```diff
@@ -39,6 +39,10 @@ def _direct_clang_module_for_name(deps, module_name):
 
 def _system_swiftinterface_impl(ctx):
     swift_toolchain = swift_common.get_toolchain(ctx, toolchain_type = SWIFT_SDK_TOOLCHAIN_TYPE)
+    if ctx.attr.system_swiftinterface and ctx.file.swiftinterface:
+        fail("exactly one of system_swiftinterface or swiftinterface must be set")
+
+    swiftinterface_file = ctx.file.swiftinterface or ctx.attr.system_swiftinterface
```
aaronsky, line 45: it's kinda weird that these aren't the same attr

keith: yea tough because it's files vs strings, the file case is really just for tests here, i could write a separate test rule but it feels risky for drift

adincebic (approved):

PR #1748, opened 2026-05-14
Update pre-commit hooks

keith:
Specifically clang-format has a lot of new fixes


keith:
needs a big clang-format before merge

adincebic (approved):

PR #1749, opened 2026-05-14
Add test for objc_library modules interop

keith:


keith:
https://github.com/bazelbuild/rules_swift/pull/1532

PR #1750, opened 2026-05-14
Handle -v in the worker

keith:
swiftc handles -v and prints the frontend invocation, but it can also be
useful to see the swiftc invocation, which is what this does.


adincebic (approved):
Nice.

PR #1751, opened 2026-05-14
Test stats collection

keith:


keith:
for use with https://github.com/swiftlang/swift/blob/main/docs/CompilerPerformance.md#post-processing-tools-for-diagnostics

PR #1753, opened 2026-05-18
Stop using repo_name

adincebic:
This should no longer be needed.


swift/toolchains/BUILD
```diff
@@ -1,11 +1,100 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
-load(
-    "@build_bazel_apple_support//configs:platforms.bzl",
-    "APPLE_PLATFORMS_CONSTRAINTS",
-)
 
 licenses(["notice"])
 
+APPLE_PLATFORMS_CONSTRAINTS = {
```
aaronsky: intentional?

adincebic: Oh no, thanks for flagging it.

adincebic: Proper fix here https://github.com/bazelbuild/apple_support/pull/592

luispadron (approved):

PR #1754, opened 2026-05-21
Fix test observer warnings-as-errors build

AttilaTheFun:
## Summary

- Explicitly discard the result of `Set.insert` inside the test observer’s `withLock` mutation.
- Avoid the confusable trailing-closure pattern when checking discovered Swift Testing suites.

## Testing

- `bazel build //tools/test_observer:test_observer --features=swift.treat_warnings_as_errors`


katre:
Question for reviewers: is it worth adding `--features=swift.treat_warnings_as_errors` to the presubmit to make sure this stays clean?

AttilaTheFun:
@katre I'd love that -- I don't know what else in rules swift might fail though and how involved it would be to fix all warnings in the repo.

adincebic (approved):

PR #1755, opened 2026-05-21
docs: add standalone Swift toolchain guide

ma-oli:
## Summary

The `swift.toolchain` Bzlmod extension lets users register a hermetic Swift
toolchain downloaded from swift.org, but until now it was undocumented —
new users had to read `MODULE.bazel` and the extension source to figure
out the API.

This PR adds:

- `doc/standalone_toolchain.md` covering the `swift.toolchain` tag
  attributes, supported platform keys, the targets generated per platform
  (`cc_toolchain_embedded_*`, `swift_toolchain_embedded_*`,
  `swift_toolchain_exec_*`), how to register checksums for unbundled
  versions via `bazel run @build_bazel_rules_swift//tools/swift-releases`,
  snapshot toolchain identifiers, and the `dev_dependency = True` pattern
  for non-root modules.
- A short pointer in the top-level `README.md`, placed before the existing
  "Building with Custom Toolchains" section so the hermetic option is
  surfaced alongside the `--action_env=TOOLCHAINS=…` flow.

No code changes.

Also while we're at it, add 6.3.1 and 6.3.2 as supported toolchain versions.


aaronsky (approved):
thanks!

PR #1760, opened 2026-05-27
Fix the test observer's handling of suites in more recent version of swift-testing.

keith:
The `listTests` request stopped reporting suites that don't use the `@Suite` attribute (https://github.com/swiftlang/swift-testing/issues/1099). This feels like a bug, but in the meantime, we can work around it by checking whether the test ID ends in a `)` or not. Functions will, and suites won't.

PiperOrigin-RevId: 753615105
(cherry picked from commit 4ca0c9bd749a519e5a57e4a5f60f3a84d1d59948)


keith:
upstream issue is fixed but this code is still in the upstream branch

aaronsky (approved):

PR #1768, opened 2026-05-27
Enable Swift v6 mode in some examples

keith:
Cherry picks 3a5851148921954c0be0980dc8821e4d956ed1bf


keith:
eh nvm this was reverted on their branch https://github.com/bazelbuild/rules_swift/commit/5ec9a944665bc51863af631eb49c57223b8b8b8f

we can just do this for whichever we actually care about

PR #1776, opened 2026-05-28
Add swift.emit_localized_strings feature for compiler string extraction

jpsim:
## Summary

Adds an opt-in `swift.emit_localized_strings` feature that exposes the
Swift compiler's localized-string data (`.stringsdata`) as a declared
directory output. This is the Bazel equivalent of Xcode's
`SWIFT_EMIT_LOC_STRINGS` ("Use Compiler to Extract Swift Strings") build
setting, and produces input suitable for `xcstringstool sync`.

The implementation follows the existing `swift.index_while_building` /
`indexstore_directory` supplemental-output pattern:

* When the feature is enabled, the compile action passes
  `-emit-localized-strings` and `-emit-localized-strings-path <dir>`.
* Because the compiler emits one `.stringsdata` file per source file and
  that set is not known at analysis time, the output is a declared
  directory (like the index store).
* The directory is surfaced via a new `swift_localized_strings` output
  group.
* Emission is registered only on the compile action, since it does not
  run under `-typecheck`.

If a user both enables the feature and passes their own
`-emit-localized-strings-path` non-hermetically in copts, the build
fails loudly (the declared directory is never written into) rather than
silently honoring the user's path. Hitting this requires a deliberate,
contradictory configuration.

The feature is off by default; existing builds are unaffected.

## Test plan

Adds a `localized_strings_tests` suite covering:

* flag emission when enabled, and the absence of the flag by default;
* the `swift_localized_strings` output group for single-file,
  multi-file, and whole-module-optimization builds;
* split derived-files generation (emission stays on the compile action,
  not the derive-files action).

```
bazel test //test:localized_strings //doc:test_api
```

All pass locally. `doc/api.md` is regenerated for the new `compile`
output field.

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -41,6 +41,7 @@ load(
     "SWIFT_FEATURE_DECLARE_SWIFTSOURCEINFO",
     "SWIFT_FEATURE_EMIT_BC",
     "SWIFT_FEATURE_EMIT_C_MODULE",
+    "SWIFT_FEATURE_EMIT_LOCALIZED_STRINGS",
     "SWIFT_FEATURE_EMIT_PRIVATE_SWIFTINTERFACE",
     "SWIFT_FEATURE_EMIT_SWIFTDOC",
     "SWIFT_FEATURE_EMIT_SWIFTINTERFACE",
@@ -563,6 +564,12 @@ def compile(
                 the indexstore output files created when the feature
                 `swift.index_while_building` is enabled.
 
+            *   `localized_strings_directory`: A directory-type `File` that
+                represents the location where the Swift compiler's
+                `.stringsdata` localized-string files were written (one per
+                source file), created when the feature
+                `swift.emit_localized_strings` is enabled.
+
             *   `macro_expansion_directory`: A directory-type `File` that
                 represents the location where macro expansion files were written
                 (only in debug/fastbuild and only when the toolchain supports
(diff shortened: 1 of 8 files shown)


aaronsky (approved):

swift/internal/compiling.bzl
```diff
@@ -1854,6 +1882,24 @@ def _is_index_store_path_overridden(copts):
             return True
     return False
 
+def _is_localized_strings_path_overridden(copts):
```
keith: i kinda think we should just not handle this so it causes failures, any major issue with that? user still has to enable this feature + have this non-hermetically in copts for it to be an issue?

swift/toolchains/config/compile_config.bzl
```diff
@@ -2250,6 +2260,25 @@ def _index_while_building_configurator(prerequisites, args):
         args.add("-Xcc", "-index-unit-output-path")
         args.add("-Xcc", index_output_path)
 
+def _emit_localized_strings_configurator(prerequisites, args):
+    """Adds flags for localized-string extraction to the command line."""
+
+    # `localized_strings_directory` is `None` when the user supplied their own
```
keith: above comment would mean removing this logic too

test/fixtures/multiple_files/BUILD
```diff
@@ -16,3 +16,17 @@ swift_library(
...
+swift_library(
+    name = "multiple_files_localized_strings_path_override",
+    srcs = [
+        "Empty.swift",
+        "Empty2.swift",
+    ],
+    copts = [
+        "-emit-localized-strings-path",
+        "/tmp/user/override",
+    ],
```
keith: above comment would mean tossing this too

swift/internal/compiling.bzl
jpsim: I modeled this based on the existing similar index store path feature, but your suggestion to simplify is fine with me.

keith: oh yea weird. traced back to https://github.com/bazelbuild/rules_swift/pull/297

i think we might as well start w/o it and can add it if needed

keith (approved):

PR #1781, opened 2026-05-28
Remove swift.vfs_overlay

keith:
This has been superseded by `swift.use_explicit_swift_module_map` which
has the same perf improvement and is used by Xcode as well. If you find
any issues where this isn't equivalent perf, please file them.


aaronsky (approved):

keith:
we can revert if something comes up, good time to do this with the major version bump

PR #1782, opened 2026-05-28
Add the feature swift.werror.<diagnostic-name> to selectively allow certain warnings from the Swift compiler to be treated as errors.

keith:
The feature can be specified multiple times, with different diagnostic
names.

The prefix `swift.werror...` is meant to evoke the `-Werror` Clang flag;
`swift.warning_as_error...` and other variations felt too verbose.

This works by using the `-debug-diagnostic-names` flag to print the
symbolic name of the diagnostic at the end of the message, in square
brackets. Then the worker process intercepts stderr and looks for
diagnostics with matching names, and modifies the message and exit code
if there's a match.

Cherry picks:

- 66a01374d9ddd087e16f7c81515ee31c031fb7af
- 70d9215e105684666ff9c04e63a2b06387e01978


aaronsky (approved):

allevato:
So, when I implemented this originally, it was shortly before they added actual warning controls to the compiler, but I had finally given into pressure from folks who wanted better warning control without using `-warnings-as-errors`. I immediately wished I had just waited a little bit longer, because now we have adoption of this version of the feature that I'd eventually like to migrate to the _real_ warning control flags, but there are still diagnostic IDs that some folks are `werror`ing that don't have new diagnostic groups assigned yet, so we're in this like purgatory state.

I would _**strongly recommend**_ that you not take the ID-based version of the feature that required scraping and munging the compiler's stderr and just map `swift.werror.*` to the appropriate `-Werror` flag in modern compilers. Save yourselves the pain I've dealt myself.

keith:
I wondered about that when I was looking at this, I didn't realize that was already in the release

PR #1783, opened 2026-05-28
Use -swift-module-file over -I

keith:
When you have many -I flags the behavior previously was O(N^2) which
lead to slow compiles. We have the VFS overlay or explicit module map
features for this but sourcekit ignores those flags right now. This is a
different attempt at doing this in a way that works fine for Xcode
but doesn't have perf issues. This is also compatible with disabling
implicit modules

### DIFF
--- a/swift/internal/compiling.bzl
+++ b/swift/internal/compiling.bzl
@@ -134,8 +134,9 @@ def _explicit_swift_module_map_info(
         feature_configuration = feature_configuration,
         feature_name = SWIFT_FEATURE_USE_C_MODULES,
     ):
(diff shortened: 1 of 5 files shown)


keith:
related https://github.com/bazelbuild/rules_swift/pull/1787

keith:
https://github.com/bazelbuild/rules_swift/commit/4359b95b41749eefaf46a9dfe780c08e02362bb4

keith:
we could probably use this instead of `-I` but it doesn't work for explicit modules so i don't care as much. we can revisit if there's a perf case where we can't use the json file but this would fix it

PR #1785, opened 2026-05-29
Make layering check validate unused deps

keith:
This now errors if you have deps on something that you don't import.
Removing a dep might actually lead to behavior changes if there are
implicit Swift behaviors, but in that case the user might prefer to
explicitly import anyways.


adincebic:
This would be incredibly useful.

PR #1789, opened 2026-06-01
Revert "Use -swift-module-file for explicit modules (#1787)"

keith:
This uncovered a new case of differing linker behavior with system
modules. I think we just have to use the json file for system stuff.

https://bazelbuild.slack.com/archives/C09TMF7SK5M/p1780345143085759

This reverts commit c9897bcd0cbd455bd2d2b45f70b0475424845051.


keith:
the second commit here adds the regression test that we found

adincebic (approved):

PR #1793, opened 2026-06-02
Test Remove unused configurator

keith:
Google only logic that complicates reading this


keith:
https://github.com/bazelbuild/rules_swift/pull/1792 is the real pr

PR #1798, opened 2026-06-03
Add nodeps dependency on rules_swift_package_manager

keith:
Lots of stuff changed here, if a user updates to the 4.x rules_swift
release, we want them to also get this version with fixes.


keith:
given https://github.com/bazelbuild/rules_swift/pull/1801 i think we can just not, if you want to manually manage deps you can update it yourself

PR #1810, opened 2026-06-15
Mark explicit modules module extension as reproducible

keith:
This removes lockfile entries that can vary per machine.


tools/explicit_modules/extensions.bzl
```diff
@@ -173,6 +173,8 @@ def _sdk_extension_impl(module_ctx):
         sdks, exclude_modules = _collect_sdk_config(module_ctx)
         _generate_local_repos(module_ctx, sdks, exclude_modules)
 
+    return module_ctx.extension_metadata(reproducible = True)
```
aaronsky, line 176: `extension_metadata`'s `reproducible` param was added in bazel 8.x and 7.1+, so this is okay

aaronsky (approved):

PR #1813, opened 2026-06-16
Add Static Linux Swift SDK support for cross-compilation

adincebic:
Adding a new swift.static_linux_sdk(...) module extension tag, which downloads the Static Linux SDK artifact bundle from swift.org, generates matching Swift and C/C++ toolchains for x86_64-swift-linux-musl and aarch64-swift-linux-musl, and
  wires those toolchains into the existing standalone-toolchain flow.

I am also adding a dedicated //swift/toolchains:static_linux constraint to distinguish Static Linux targets from generic Linux targets. Further more, I tried documenting Static Linux SDK setup and toolchain registration

  One important detail is toolchain registration order: Static Linux SDK toolchains must be registered before generic Linux Swift toolchains, since both are compatible with @platforms//os:linux and Bazel prefers earlier registered toolchains. Open to discussing if there is a better way to do this.


swift/toolchains/swift_toolchain.bzl
```diff
@@ -457,6 +512,13 @@ def _parse_target_system_name(*, arch, os, target_system_name):
 
     if os == "linux":
         return "%s-unknown-linux-gnu" % arch
+    elif os == "windows":
+        # The MSVC cc toolchain reports a `target_gnu_system_name` of "local",
+        # so synthesize the triple. The `-msvc` environment is required: while
+        # `swiftc` defaults to it, `swift-symbolgraph-extract` (used for XCTest
+        # test discovery) matches the module layout's triple exactly and fails
+        # to load modules built for `*-windows-msvc` if it is omitted.
+        return "%s-unknown-windows-msvc" % arch
```
cerisier, line 513: What are the other windows target triples supported by swiftc ?

doc/standalone_toolchain.md
```diff
@@ -146,6 +146,178 @@ bazel run @rules_swift//tools/swift-releases -- list \
     main-snapshot-2024-08-01 --platform xcode --platform ubuntu22.04
 ```
 
+## Cross-compiling with Swift SDKs
+
+swift.org publishes "Swift SDK" artifact bundles (the bundles consumed by
+`swift sdk install`) that let the host compiler cross-compile for platforms
+it cannot target by itself. The `swift` extension can download these and
+define matching Swift and C/C++ toolchains, so that plain `swift_library`
```
cerisier: Should you make this opt-out to allow c/c++ toolchains to be registered by a different module ?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -0,0 +1,694 @@
...
+load("@rules_cc//cc/toolchains:tool_map.bzl", "cc_tool_map")
+load("@rules_cc//cc/toolchains:toolchain.bzl", "cc_toolchain")
+load("@rules_swift//swift/toolchains:swift_toolchain.bzl", "swift_toolchain")
+load("@rules_swift//swift/toolchains:swift_tools.bzl", "swift_tools")
+
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "sdk_files",
+    srcs = glob(["{bundle_dir}/**"]),
```
cerisier: Should you use source directories ?

swift/toolchains/BUILD
```diff
@@ -6,6 +6,17 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
 licenses(["notice"])
 
+constraint_setting(
+    name = "swift_sdk",
+    visibility = ["//visibility:public"],
+)
+
+constraint_value(
+    name = "static_linux",
```
cerisier, line 20: If `static_linux` really implies `musl`, I would model that using `libc` semantic. That way everything will align when https://github.com/bazel-contrib/platforms_contrib/pull/5 lands.

swift/internal/extensions/swift_sdk_releases.bzl
```diff
@@ -0,0 +1,112 @@
...
+        "android": "939e933549d12d28f2e0bf71019d734d309859e9773c572657ce565a81f85d68",
+        "static_linux": {
+            "sha256": "3fd798bef6f4408f1ea5a6f94ce4d4052830c4326ab85ebc04f983f01b3da407",
+            "version": "0.1.0",
+        },
+        "wasm": "a61f0584c93283589f8b2f42db05c1f9a182b506c2957271402992655591dd7c",
+    },
+}
+
+# The Android Swift SDK bundle ships without an NDK sysroot (its
```
dzbarsky: should we reuse https://github.com/keith/hermetic_android_toolchains cc @keith

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -0,0 +1,694 @@
...
+cross-compiling Swift to platforms that the host toolchain cannot target by
+itself (currently WebAssembly, Android, and Static Linux); they are the
+bundles that `swift sdk install` consumes.
+
+Each repository created by these rules pairs one Swift SDK with one standalone
+host toolchain repository (created by `standalone_toolchain`) and defines:
+
+  * a `swift_toolchain` that compiles against the SDK's sysroot and Swift
+    resource directory, and links against its static Swift runtime; and
+  * a rules_cc `cc_toolchain` that drives the matching clang for the target
```
dzbarsky: is it possible to reuse existing cc toolchain like hermetic-llvm? I feel like otherwise there will be a lot of duplicated `clang` binaries in the output base and more maintenance burden overall. We already support wasm, happy to look at what's missing to make android work better

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -0,0 +1,694 @@
...
+# Files in the host toolchain that compile actions need: the driver/frontend
+# binaries, their libraries, and clang's builtin headers (used by the clang
+# importer when the Swift SDK's resource directory does not bundle them).
+_HOST_COMPILER_INPUTS = "swift_sdk_compiler_inputs"
+
+# Files in the host toolchain that link actions driven by its clang need.
+_HOST_LINKER_INPUTS = "swift_sdk_linker_inputs"
+
+_CC_TOOLCHAIN_TEMPLATE = """
+cc_tool(
```
dzbarsky, line 83: i think a lot of this could be cleaner if it was an actual `.bzl` file that had a macro to stamp out these targets and this `.bzl` emitted a load to it. would make LSPs less confused

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -0,0 +1,694 @@
...
+        # resource directory does not include.
+        link_args = _build_list([
+            "-resource-dir",
+            resource_dir + "/clang",
+        ]),
+        suffix = "wasm32",
+        triple = "wasm32-unknown-wasip1",
+    )
+
+    repository_ctx.file("BUILD.bazel", build_content)
```
dzbarsky, line 499: please make sure you mark all repo rules and module extensions `reproducible = True` as appropriate!

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -0,0 +1,694 @@
...
+
+swift_android_sdk_repository = repository_rule(
+    attrs = _common_attrs() | {
+        "api_level": attr.int(
+            doc = "The Android API level to target.",
+            mandatory = True,
+        ),
+        "host_swiftc": attr.label(
```
dzbarsky: do you actually need to pass this or can you just construct the label yourself? (repo rule can fabricate labels). That's how `rules_rs` deals with host_cargo

swift/internal/extensions/swift_sdk_releases.bzl
adincebic: @AttilaTheFun

swift/internal/extensions/toolchains.bzl
```diff
@@ -43,24 +43,179 @@ toolchain(
...
+            target = "android_" + arch,
+            target_compatible_with = [
+                "@platforms//os:android",
+                "@platforms//cpu:" + arch,
+            ],
+            target_suffix = arch,
+        )
+    return content
+
+def static_linux_sdk_toolchains_for_platform(platform, sdk_repository, archs):
```
dzbarsky, line 150: does this mean you are duplicating all the SDK bytes just to change the target to musl? Or is it reusing the tools but only emitting new `toolchain` targets? (Finding it a bit hard to trace what is materialized where, the way this is written)

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -269,15 +285,27 @@ xcode_swift_toolchain(
         ]),
     )
 
+def _python_executable_works(repository_ctx, python_bin):
```
dzbarsky, line 253: shouldn't we be using the python toolchain? this seems pretty hacky (though I see its partially preexisting)

swift/toolchains/BUILD
dzbarsky: +1. and if you want you can reuse @llvm's libc/musl constraint for now, that is what rules_rs does as well. Once platforms_contrib ships, we will make ours a forwarding alias

swift/toolchains/swift_toolchain.bzl
```diff
@@ -457,6 +512,13 @@ def _parse_target_system_name(*, arch, os, target_system_name):
 
     if os == "linux":
         return "%s-unknown-linux-gnu" % arch
+    elif os == "windows":
+        # The MSVC cc toolchain reports a `target_gnu_system_name` of "local",
```
dzbarsky, line 511: what is the "MSVC cc toolchain", do you mean the default one created by rules_cc? What about hermetic ones like @llvm? this feels like it's baking in a lot of assumptions about other toolchains?

MODULE.bazel
```diff
@@ -110,6 +117,52 @@ register_toolchains(
...
+    "@swift_toolchain//:swift_toolchain_wasm32_ubuntu22.04",
+    "@swift_toolchain//:swift_toolchain_wasm32_ubuntu22.04-aarch64",
+    "@swift_toolchain//:swift_toolchain_wasm32_xcode",
+    dev_dependency = True,
+)
+
+# Keep broad built-in Linux toolchains after Static Linux SDK toolchains:
+# Static Linux platforms are also `@platforms//os:linux`, and Bazel chooses the
+# first compatible toolchain.
+register_toolchains("//swift/toolchains:all")
```
dzbarsky, line 225: can we stop auto-registering these? :)

swift/internal/extensions/swift_sdks.bzl
cerisier: So the problem I think is that unless those are used for exec tools, the LLVM version used by Apple Clang is that of their own and mismatch means either failures or no LTO

keith:
I'll review this one once logan's stuff is in!

MODULE.bazel
keith: punting this convo to https://github.com/bazelbuild/rules_swift/pull/1815 since nothing new is being registered in this PR, only dev toolchains

test/swift_toolchain_tests.bzl
```diff
@@ -11,6 +11,15 @@ toolchain_macos_arm64_with_sdkroot_test = make_action_command_line_test_rule(
...
 )
 
+toolchain_static_linux_x86_64_test = make_action_command_line_test_rule(
+    config_settings = {
+        "//command_line_option:extra_toolchains": [
+            "//test/fixtures/toolchains:cc_toolchain_static_linux_x86_64",
+            "//test/fixtures/toolchains:toolchain_static_linux_x86_64",
+        ],
+    },
+)
```
keith, line 20: can we use the real toolchain we're registering for this instead of creating some fake ones? this doesn't prove much since we're just testing the behavior of the copy pasted fixture AFAIUI?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -207,6 +207,97 @@ def _download_sdk_bundle(repository_ctx):
...
+        ))
+    return value
+
+def _static_linux_resource_path(target_settings, triple):
+    resource_path = target_settings.get("swiftStaticResourcesPath")
+    field = "swiftStaticResourcesPath"
+    if not resource_path:
+        resource_path = target_settings.get("swiftResourcesPath")
+        field = "swiftResourcesPath"
+    return _relative_metadata_path(resource_path, field, triple)
```
keith: when doesn't it include `swiftStaticResourcesPath`?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -207,6 +207,97 @@ def _download_sdk_bundle(repository_ctx):
...
+            else:
+                linkopts.extend(["-Xlinker", arg])
+            linker_arg_expected = False
+        elif arg == "-Xlinker":
+            linker_arg_expected = True
+        else:
+            linkopts.append(arg)
+
+    if linker_arg_expected:
+        linkopts.append("-Xlinker")
```
keith: why are we processing these? why not take them as they are in the file?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -207,6 +207,97 @@ def _download_sdk_bundle(repository_ctx):
...
+
+def _static_linux_linkopts_from_sdk(
+        repository_ctx,
+        *,
+        arch,
+        linux_static_dir,
+        linux_static_dir_relative,
+        repo_root):
+    swiftrt_relative = "{}/{}/swiftrt.o".format(linux_static_dir_relative, arch)
+    args_relative = linux_static_dir_relative + "/static-executable-args.lnk"
```
keith, line 265: what is the difference between this one and the static-stdlib-args.lnk? looks like the args aren't 1:1, so maybe we need some combo? fwiw for android and wasm we didn't fetch from this file we hardcoded the args instead. this is probably a bit nicer for when new ones are introduced, assuming that all these args make sense in bazel.

swift/toolchains/BUILD
keith: it does imply musl, but maybe the problem is you "can't" just use any old musl cc toolchain, since apple's clang fork that ships with swift is tied to who knows what implementation details. so that probably still means we need some custom constraint like this? (and maybe should also add the musl constraint for random select() statements)

examples/cross_compilation/static_linux/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
...
+load("//swift:swift.bzl", "swift_binary", "swift_library")
+
+platform(
+    name = "static-linux-x86_64",
+    constraint_values = [
+        "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
+        "//swift/toolchains:static_linux",
+    ],
+)
```
keith, line 11: if this is something we end up requiring, we should move this somewhere that users can reference it too

keith (commented):
still reviewing but flushing some comments

swift/internal/extensions/swift_sdks.bzl
adincebic: Because we can't just take the contents as is:

```
-static
-lswiftCore
-lswift_RegexParser
-Xlinker
-undefined=pthread_self
-Xlinker
-undefined=pthread_once
-Xlinker
-undefined=pthread_key_create
-ldispatch
-lBlocksRuntime
-lpthread
-ldl
-lc++
-lm
```

purely because of `-Xlinker` flags.

swift/internal/extensions/swift_sdks.bzl
adincebic: `static-stdlib-args.lnk` is for the case where you want to statically link Swift standard library `swiftc -static-stdlib` which we don't need here, at least not yet.

examples/cross_compilation/static_linux/BUILD.bazel
adincebic: Probably at the root?

swift/internal/extensions/swift_sdks.bzl
adincebic: Also those flags that come from  static-executable-args.lnk seem to make sense with Bazel unless I am missing something. I wouldn't like us to hardcode them if we don't have to.

swift/internal/extensions/swift_sdks.bzl
adincebic: Well it actually includes it every time at least starting from Swift 6.3.2. However since both `swiftStaticResourcesPath` and `swiftResourcesPath` are present in `swift-sdk.json` I first check for the `swiftStaticResourcesPath` and fall back to the `swiftResourcesPath`

```
{
  "schemaVersion": "4.0",
  "targetTriples": {
    "x86_64-swift-linux-musl": {
      "toolsetPaths": [
        "toolset.json"
      ],
      "sdkRootPath": "musl-1.2.5.sdk/x86_64",
      "swiftResourcesPath": "musl-1.2.5.sdk/x86_64/usr/lib/swift_static",
      "swiftStaticResourcesPath": "musl-1.2.5.sdk/x86_64/usr/lib/swift_static"
    },
    "aarch64-swift-linux-musl": {
      "toolsetPaths": [
        "toolset.json"
      ],
      "sdkRootPath": "musl-1.2.5.sdk/aarch64",
      "swiftResourcesPath": "musl-1.2.5.sdk/aarch64/usr/lib/swift_static",
      "swiftStaticResourcesPath": "musl-1.2.5.sdk/aarch64/usr/lib/swift_static"
    }
  }
}
```

adincebic: Ok I downloaded some of the older ones and all of them have the same shape, so I am dropping the `swiftResourcesPath` since it has the same value as `swiftStaticResourcesPath` anyway.

test/swift_toolchain_tests.bzl
adincebic: Yes.

swift/internal/extensions/swift_sdks.bzl
keith: you can remove this after https://github.com/bazelbuild/rules_swift/pull/1839

adincebic: Perfect. Removed.

adincebic:
@keith when you find some time, we should try to wrap this up. I am fine with closing the PR if you think that it needs complete redesign and I can start from scratch.

PR #1815, opened 2026-06-22
Stop registering Swift toolchains transitively

dzbarsky:
## Summary

It's bad practice for rules to autoregister toolchains, as it takes away control from end users and can cause undesired fetches. Especially when the toolchain in question triggers non-hermetic repo rule executions.

Mark `//swift/toolchains:all` registration in `rules_swift`'s `MODULE.bazel` as a development dependency. Dependent Bazel modules no longer inherit the autoconfigured Swift toolchain registration and must explicitly register `@rules_swift//swift/toolchains:all`.

Update `README.md` and `.github/generate-notes.sh` so release snippets include the required explicit registration. `rules_swift` development continues to register `//swift/toolchains:all` because `rules_swift` is the root module during development.

### DIFF
--- a/.github/generate-notes.sh
+++ b/.github/generate-notes.sh
@@ -15,5 +15,7 @@ This release is compatible with: TODO
 
 \`\`\`bzl
 bazel_dep(name = "rules_swift", version = "$new_version")
+
+register_toolchains("@rules_swift//swift/toolchains:all")
 \`\`\`
 EOF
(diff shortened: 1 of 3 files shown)


keith:
which case is this eagerly fetched now vs your override (if you need it at all) taking over?

dzbarsky:
My primary interest is making sure the pattern fixed in https://github.com/bazelbuild/rules_swift/pull/1651 isn't accidentally reintroduced. But also in Adins PR there are repeated references to needing to register toolchains in the right order so they don't override each other. I think it's simplest to just let the consumer issue the 1-line call and have full control :)

PR #1817, opened 2026-06-23
Support cross-compiling to WebAssembly with a Swift SDK

AttilaTheFun:
WebAssembly support, part of the #1809 split, on top of the now-merged foundation
PRs #1820 (`swift_binary(linkshared)`) and #1821 (Swift-SDK framework). A sibling
of the Android PR (#1818).

## What this adds

`swift.wasm_sdk`: downloads the swift.org WebAssembly Swift SDK bundle matching a
`swift.toolchain` and, using the shared framework (#1821), defines Swift + C/C++
toolchains targeting `wasm32-unknown-wasip1`. `swift_binary(linkshared = True)`
gains the WebAssembly **reactor** output here (`-mexec-model=reactor`, no `main`,
exported functions) — the wasm-specific bit that keith asked to keep out of the
generic `swift_binary`. `examples/cross_compilation` builds a reactor and a
browser web app that drives it from JS.

## Open questions

- **`-swift.file_prefix_map`** is disabled for the wasm toolchain. With it on, the
  worker's developer-dir remap hard-requires `DEVELOPER_DIR` on Apple hosts. The
  clean fix is to gate that remap on `swift._supports_developer_dir` (the
  analysis-time config already is); it touches the shared Apple worker path, so I
  wanted your read first.


.bazelci/presubmit.yml
```diff
@@ -62,6 +62,21 @@ tasks:
...
+  macos_cross_compilation:
+    name: "Cross-compilation (wasm)"
+    platform: macos_arm64
+    xcode_version: "26.2"
+    bazel: latest
+    # The //examples/cross_compilation targets are tagged `manual` (they fetch
+    # the Swift SDK bundle), so they are excluded from the `//examples/...`
+    # wildcard the other tasks build. List them explicitly here so the Swift-SDK
+    # cross-compilation toolchain is exercised in CI. Build-only: the trivial
+    # reactor exercises the link path, while runtime behavior is covered
```
keith: how bad is this download? i think unless it's really horrible we should probably drop this and remove the manual tags. being able to do `bazel test ...` and have that be more complete for CI is valuable IMO

examples/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,72 @@
...
+        ":Reactor.wasm",
+    ],
+    outs = [
+        "web_app/index.html",
+        "web_app/Reactor.wasm",
+    ],
+    cmd = "cp $(location web/index.html) $(RULEDIR)/web_app/index.html && " +
+          "cp $(location :Reactor.wasm) $(RULEDIR)/web_app/Reactor.wasm",
+    tags = ["manual"],
+)
```
keith: is there any better way for this to be a "test" as is afaiui we wouldn't be 100% sure this works unless someone manually ran this and audited? if for nothing else for CI breakages

examples/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,72 @@
+load("//examples/embedded:transition.bzl", "transition_binary")
+load("//swift:swift.bzl", "swift_binary", "swift_library")
+
+package(default_visibility = ["//visibility:public"])
+
```
keith:
```suggestion
```

examples/cross_compilation/README.md
```diff
@@ -0,0 +1,33 @@
+# Cross-compilation example (WebAssembly)
```
keith: i'd toss  this file and move the wasm stuff to a wasm subdirectory instead, seems like most of this is covered in the BUILD file

keith (commented):
didn't get all the way through yet but leaving what i have so far

swift/internal/extensions/toolchains.bzl
```diff
@@ -55,6 +55,30 @@ toolchain(
...
+    name = "swift_toolchain_{target}_{platform}",
+    exec_compatible_with = {exec_compatible_with},
+    target_compatible_with = {target_compatible_with},
+    toolchain = "@{sdk_repository}//:swift_toolchain_{target_suffix}",
+    toolchain_type = "@rules_swift//toolchains:toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
+toolchain(
+    name = "cc_toolchain_{target}_{platform}",
```
keith, line 76: im a bit suspicious on this one. do we really want to be registering our own here or do we want users to be bringing their own? I guess they don't have to register it and things will still work if they have a reasonable toolchain so maybe it's fine?

examples/cross_compilation/wasm/reactor_test.sh
```diff
@@ -0,0 +1,16 @@
...
+# Executes the Swift reactor under wasmtime and asserts on the value Swift
+# computes: `greeting_length` returns the UTF-8 length of
+# "Hello from Swift, WebAssembly!" (30).
+
+set -euo pipefail
+
+wasmtime="$TEST_SRCDIR/$1"
+reactor="$TEST_SRCDIR/$2"
+
+actual="$("$wasmtime" run --invoke greeting_length "$reactor" 2>/dev/null)"
```
keith:
```suggestion
actual="$("$wasmtime" run --invoke greeting_length "$reactor")"
```

probably helpful for debugging?

examples/cross_compilation/wasm/BUILD.bazel
```diff
@@ -0,0 +1,125 @@
+# A WebAssembly reactor written in Swift, executed under a hermetic wasmtime
+# and embedded in a static web app.
+
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load("//examples/embedded:transition.bzl", "transition_binary")
+load("//swift:swift.bzl", "swift_binary", "swift_library")
+
+package(default_visibility = ["//visibility:public"])
+
```
keith:
```suggestion
```

examples/cross_compilation/wasm/BUILD.bazel
```diff
@@ -0,0 +1,125 @@
+# A WebAssembly reactor written in Swift, executed under a hermetic wasmtime
+# and embedded in a static web app.
+
```
keith:
```suggestion
```

.bazelci/presubmit.yml
AttilaTheFun: The SDK bundle is a 69 MB download (~286 MB extracted) — comparable to the Android SDK bundle we now fetch in the normal jobs anyway. Dropped the manual tags and the dedicated task so the example now builds and tests in the normal //examples/... wildcard on both macOS and Linux. The wasm-only Reactor is target_compatible_with os:wasi (same pattern as the Android example) so host wildcards skip it while the transitioned Reactor.wasm still builds.

examples/cross_compilation/BUILD.bazel
AttilaTheFun: Yes, it's a real executed test now. reactor_test runs the reactor under a hermetic wasmtime (a ~10 MB dev-dependency download per host, pinned in wasm.MODULE.bazel) and asserts on the value Swift computes. So CI executes Swift-compiled wasm, not just links it. I went with wasmtime over Node/rules_nodejs since it's a single static binary and doesn't add a JS toolchain. And we can keep browser demo for humans to try it out.

keith (approved):
lgtm. very minor stuff

swift/internal/extensions/toolchains.bzl
AttilaTheFun: I needed to keep it with the reasoning now documented in a comment: this cc toolchain drives the SDK bundle's own clang (so the pair stays version-matched), it's `target_compatible_with` the wasm platform only so it can never win resolution anywhere else, and since root-module registrations take precedence, anyone who brings their own wasm cc toolchain overrides this one automatically. Without a registered default, every consumer would have to configure a wasm cc toolchain before anything links.

examples/cross_compilation/wasm/reactor_test.sh
AttilaTheFun: Done — dropped the `2>/dev/null`.

examples/cross_compilation/wasm/BUILD.bazel
AttilaTheFun: Done.

examples/cross_compilation/wasm/BUILD.bazel
AttilaTheFun: Done.

AttilaTheFun:
Rebased onto main (post-#1819) and addressed the remaining comments — the example BUILD nits are applied, and the wasm cc-toolchain registration is kept with the rationale documented inline (wasm-platform-only resolution; root-module registrations take precedence for anyone bringing their own).

keith (approved):

PR #1818, opened 2026-06-23
Support cross-compiling to Android with a Swift SDK

AttilaTheFun:
Android support, part of the #1809 split, on top of the now-merged foundation
PRs #1820 (`swift_binary(linkshared)`) and #1821 (Swift-SDK framework). A sibling
of the WebAssembly PR (#1817); it does not depend on it.

## What this adds

`swift.android_sdk` downloads the swift.org Android Swift SDK bundle and, using
the shared framework (#1821), defines a Swift toolchain targeting
`{aarch64,x86_64}-unknown-linux-android`. `swift_binary(linkshared = True)`
produces a JNI `lib<name>.so` whose entry points are written entirely in Swift.
`examples/cross_compilation/android_app` builds one and ships a **complete,
runnable Android app** that loads it and shows the Swift-computed greeting on
screen — with nothing preinstalled (no Android Studio, no device), one command
boots a hermetic emulator, installs, and launches it:

```sh
bazel run //examples/cross_compilation/android_app:run
```

The example app running on an Android emulator, showing &quot;Hello from Swift, Android!&quot;

## The NDK comes from a cc toolchain, not from rules_swift

Per review feedback, rules_swift does **not** fetch or manage the NDK. C/C++
compilation and linking for Android targets go through a separately registered
Android C++ cc toolchain, and the Swift toolchain reads that toolchain's sysroot
at analysis time. So you bring an Android cc toolchain — e.g. `@androidndk//:all`
from [hermetic_android_toolchains](https://github.com/keith/hermetic_android_toolchains)
— and register it alongside the Swift toolchain:

```starlark
register_toolchains(
    "@swift_toolchain//:swift_toolchain_android_aarch64_xcode",
    "@androidndk//:all",
)
```

This keeps rules_swift out of the NDK-management business entirely (no NDK
download, no per-host repos, no bespoke C++ toolchain) and works with any
rules_android_ndk-based Android cc toolchain.

A few NDK-integration details that aren't obvious:

- `CcToolchainInfo.sysroot` from rules_android_ndk's toolchain reports the *clang*
  directory, not the sysroot, so the Swift toolchain derives the sysroot from the
  toolchain's files.
- The Swift binary link runs through `cc_common.link` with that cc toolchain, but
  rules_android_ndk reports its sysroot as the clang dir and exposes an empty
  `dynamic_runtime_lib`, so the toolchain adds `--sysroot` and stages
  `libc++_shared.so` itself. It also links libc++ as the *shared* runtime
  (rules_android_ndk links it statically by default — rules_android_ndk#93).
- To package `libc++_shared.so` into an APK, `select_android_runtime_lib` (in
  `swift/toolchains/android_runtime_lib.bzl`) selects it from the resolved cc
  toolchain.

## The runnable example (`examples/cross_compilation/android_app`)

A real `android_binary` (built with `rules_android` + `rules_kotlin`, the
real-world way Android apps are packaged) completes the chain end-to-end:
`MainActivity.kt → NativeBridge.greetingFromSwift()` (Kotlin, JNI) →
`libSwiftJNI.so` (`swift_binary(linkshared)`, `@_cdecl`) → `Greeter`
(`swift_library`).

The one rules_swift-specific detail: `swift_binary(linkshared = True)` exposes
its `.so` via `DefaultInfo`, not `CcInfo`, so it can't go straight into
`android_binary.deps`. Wrapping it in a `cc_library` (a prebuilt `.so` in `srcs`
becomes a `CcInfo` dynamic library) lets `android_binary`'s per-ABI native split
collect it into `lib/arm64-v8a/`; `libc++_shared.so` goes in the same way via
`select_android_runtime_lib`.

`rules_android` / `rules_kotlin` / `rules_java` / `rules_jvm_external` (and the
hermetic emulator the `:run` target downloads) are all `dev_dependency = True`,
split into `android_app/android.MODULE.bazel` and `include()`d, with a pinned
Maven lock and the hermetic `@androidsdk` — so consumers of `rules_swift` are
unaffected and `protobuf` stays at 34.

## CI

The example builds and tests in the normal macOS `//examples/...` job (the
`@androidndk`/`@androidsdk` toolchains come from hermetic_android_toolchains, dev
dependencies; the arm64 `--android_platforms` + a hermetic JDK are set in
`.bazelrc`). No emulator runs in CI; instead the Android path is guarded by:

- **Analysis tests** (`//test:android`) asserting the action command lines — the
  Swift compile targets `aarch64-linux-android`, and the link runs the NDK clang
  for the Android target against the NDK sysroot, linking the shared libc++ and
  staging `libc++_shared.so`.
- **Artifact tests** (`build_test`) running the NDK `llvm-readelf` on the built
  `.so` (AArch64 ELF, `libc++_shared.so` in `NEEDED`, the `@_cdecl` symbol
  exported in `.dynsym`) and asserting the APK packages both `.so`s under
  `lib/arm64-v8a/` plus the dexed Kotlin.


.bazelci/presubmit.yml
```diff
@@ -62,6 +62,25 @@ tasks:
...
+    # The //examples/cross_compilation targets are tagged `manual` (they fetch
+    # the Swift SDK bundle and the NDK), so they are excluded from the
+    # `//examples/...` wildcard the other tasks build. List them explicitly here
+    # so the Swift-SDK cross-compilation toolchains are exercised in CI.
+    # `--config=android_example` supplies the rules_android flags (arm64
+    # --android_platforms + a hermetic JDK) the APK needs; it's harmless for the
+    # bare .so. The build_test asserts the runnable APK (Kotlin -> Swift JNI ->
+    # Swift library) assembles; runtime behavior is shown by `android_app/`.
+    build_flags:
+      - "--config=android_example"
```
keith: doesn't look like this is needed, we can just set this always

.bazelci/presubmit.yml
```diff
@@ -62,6 +62,25 @@ tasks:
...
+    build_targets:
+      - "//examples/cross_compilation:libSwiftJNI.so"
+      - "//examples/cross_compilation/android_app:app_build_test"
```
keith: i think we should remove the manual tag and just let these run in the normal ci job

examples/cross_compilation/Sources/Greeter/Greeter.swift
```diff
@@ -0,0 +1,14 @@
+/// A plain `swift_library` used as a normal dependency of the Android JNI
+/// entry point. Nothing in here is platform-specific; it is compiled for
+/// whichever platform the depending target is built for.
```
keith:
```suggestion
```

examples/cross_compilation/Sources/SwiftJNI/SwiftJNI.swift
```diff
@@ -0,0 +1,24 @@
...
+// JNI types (`JNIEnv`, `jclass`, `jstring`, ...) from the Android Swift SDK, so
+// no C shim is needed. `@_cdecl` gives the function the exact symbol name JNI
+// looks up: `Java_<package>_<class>_<method>`, with `.`/`_` escaped per the JNI
+// spec. It is the Kotlin-callable native implementation of:
+//
+//   package com.example.swiftjni
+//   class NativeBridge { external fun greetingFromSwift(): String }
+//
+// and it delegates to the `Greeter` `swift_library` (a normal dependency),
+// completing the Kotlin -> Swift (JNI .so) -> Swift library call chain.
```
keith:
```suggestion
```

examples/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+    linkshared = True,
+    tags = ["manual"],
+    deps = [":Greeter"],
+)
+
+transition_binary(
+    name = "libSwiftJNI.so",
+    binary = ":SwiftJNI",
+    platform = ":android-aarch64",
+)
```
keith: can probably toss this and just only let this be built as a dep of the android targets?

examples/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+# The platform covered by the Android Swift SDK toolchain registered by this
+# repository's dev `MODULE.bazel` (via the `swift` extension's `android_sdk`
+# tag). Build the targets below with `--platforms` set to this.
+platform(
+    name = "android-aarch64",
+    constraint_values = [
+        "@platforms//cpu:aarch64",
+        "@platforms//os:android",
+    ],
+)
```
keith: we should use the ones exposed by rules_android instead of our own

examples/cross_compilation/BUILD.bazel
keith, line 1: should we move all of this into the android subdirectory? seems like it might be nicer for them to be side by side

.bazelrc
```diff
@@ -22,6 +22,23 @@ build --worker_sandboxing
 
 build --enable_platform_specific_config
 
+# Accept the Android NDK/SDK licenses for hermetic_android_toolchains' @androidndk
+# (the Android cc toolchain used by //examples/cross_compilation). Only consulted
+# when those repositories are fetched.
```
keith:
```suggestion
```

.bazelrc
```diff
@@ -22,6 +22,23 @@ build --worker_sandboxing
...
+common --repo_env=ACCEPTED_ANDROID_NDK_LICENSE_VERSION=r27c
+common --repo_env=ACCEPTED_ANDROID_SDK_LICENSE_VERSION=35
+
+# Build the runnable APK example (//examples/cross_compilation/android_app) with
+# `--config=android_example`. Scoped to a config (not applied globally) so the
+# rest of the build is untouched: --android_platforms drives rules_android's
+# per-ABI native split (the emulator is arm64), and the hermetic remote JDK runs
+# rules_android's Java actions instead of requiring a host JDK.
```
keith:
```suggestion
```

.bazelrc
```diff
@@ -22,6 +22,23 @@ build --worker_sandboxing
...
+common:android_example --android_platforms=@rules_android//:arm64-v8a
+common:android_example --java_language_version=17
+common:android_example --java_runtime_version=remotejdk_17
+common:android_example --tool_java_language_version=17
+common:android_example --tool_java_runtime_version=remotejdk_17
```
keith:
```suggestion
common --android_platforms=@rules_android//:arm64-v8a
common --java_language_version=17
common --java_runtime_version=remotejdk_17
common --tool_java_language_version=17
common --tool_java_runtime_version=remotejdk_17
```

MODULE.bazel
```diff
@@ -110,6 +113,100 @@ register_toolchains(
...
+    "@rules_jvm_external//:extensions.bzl",
+    "maven",
+    dev_dependency = True,
+)
+maven.install(
+    name = "rules_android_maven",
+    aar_import_bzl_label = "@rules_android//rules:rules.bzl",
+    artifacts = [],
+    known_contributing_modules = ["rules_android"],
+    lock_file = "//:rules_android_maven_install.json",
```
keith: can we move this file out of the root, i guess into the android_app dir? just to avoid the noise in the root

MODULE.bazel
```diff
@@ -110,6 +113,100 @@ register_toolchains(
...
+use_repo(rules_java_toolchains, "remote_java_tools")
+
+# rules_android's own Maven dependencies, pinned to a checked-in lock file:
+#   REPIN=1 bazel run @rules_android_maven//:pin
+maven = use_extension(
+    "@rules_jvm_external//:extensions.bzl",
+    "maven",
+    dev_dependency = True,
+)
+maven.install(
```
keith: can you link to the issue / pr about how we shouldn't have to do this

MODULE.bazel
```diff
@@ -110,6 +113,100 @@ register_toolchains(
...
+
+# A runnable APK example (//examples/cross_compilation/android_app) that loads
+# the Swift JNI .so and shows the Swift greeting on screen. It uses rules_android
+# + rules_kotlin — the real-world way Android apps are built. These are all
+# dev-only (rules_android pulls in a Maven graph and is heavy), so they never
+# affect rules_swift consumers; a consumer brings their own Android rules.
+bazel_dep(name = "rules_android", version = "0.7.3", dev_dependency = True)
+bazel_dep(name = "rules_java", version = "9.3.0", dev_dependency = True)
+bazel_dep(name = "rules_kotlin", version = "2.2.2", dev_dependency = True)
+bazel_dep(name = "rules_jvm_external", version = "7.0", dev_dependency = True)
```
keith: there's enough here that can you actually split out all of this into a `android.MODULE.bazel`, throw it in the android_app subdirectory, and `include()` it here

doc/standalone_toolchain.md
```diff
@@ -146,6 +146,109 @@ bazel run @rules_swift//tools/swift-releases -- list \
...
+    name = "android-aarch64",
+    constraint_values = [
+        "@platforms//cpu:aarch64",
+        "@platforms//os:android",
+    ],
+)
+```
+
+```sh
+bazel build //my:binary --platforms=//:android-aarch64
```
keith: we should reference the ones from rules_android instead of encouraging users to write their own

doc/standalone_toolchain.md
```diff
@@ -146,6 +146,109 @@ bazel run @rules_swift//tools/swift-releases -- list \
...
+
+### Coexistence with `rules_apple`
+
+A common setup cross-compiles to Android *and* builds the same app's Apple
+targets with `rules_apple`. The two resolve together cleanly: this line of
+`rules_swift` is `compatibility_level = 3` (the same as released `rules_swift`
+3.x), so a current `rules_apple` release works alongside it. Bazel keeps the
+higher of each shared transitive dependency (`apple_support`, `rules_cc`), which
+are backward compatible, so no extra pinning is required.
+
```
keith:
```suggestion
```

examples/cross_compilation/android_app/BUILD.bazel
```diff
@@ -0,0 +1,91 @@
...
+# root MODULE.bazel), so consumers of rules_swift are unaffected.
+#
+# Build/run (the flags come from `--config=android_example` in //.bazelrc):
+#
+#   bazel build --config=android_example \
+#     //examples/cross_compilation/android_app:app
+#
+# Everything here is tagged `manual` (like the rest of //examples/cross_compilation,
+# which fetches the Swift SDK bundle + the NDK) so it stays out of the
+# `//examples/...` wildcard the other CI tasks build.
```
keith:
```suggestion
# A runnable Android app that loads runs Swift code through the JNI
```

swift/toolchains/android_runtime_lib.bzl
```diff
@@ -0,0 +1,47 @@
+"""A rule that selects a single Android NDK runtime library from the toolchain.
+
+The Android NDK's shared C++ runtime (`libc++_shared.so`) must be packaged into
```
keith: it doesn't seem like this is true for normal C++ NDK stuff:

```
% ldd lib/x86_64/libapp.so
  0x0000000000000001 (NEEDED)       Shared library: [liblog.so]
  0x0000000000000001 (NEEDED)       Shared library: [libdl.so]
  0x0000000000000001 (NEEDED)       Shared library: [libc.so]
  0x0000000000000001 (NEEDED)       Shared library: [libm.so]
```

is this a swift specific thing or is this incorrect?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -201,3 +134,110 @@ this SDK is paired with.
             mandatory = True,
         ),
     }
+
+# The architectures the Android Swift SDK provides resources for and that
+# `@platforms//cpu` can express. (The SDK also supports armv7, which can be
+# added on demand.)
+ANDROID_ARCHS = ["aarch64", "x86_64"]
```
keith: should we be reading this dynamically somehow?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -201,3 +134,110 @@ this SDK is paired with.
...
+            ]),
+            features = _build_list([
+                "swift.lld_gc_workaround",
+                "swift.module_map_no_private_headers",
+                "swift.use_autolink_extract",
+                "swift.use_module_wrap",
+                # The file prefix map would make the worker resolve the Xcode
+                # developer directory on macOS hosts, which this toolchain
+                # does not depend on.
+                "-swift.file_prefix_map",
```
keith: i think you probably want `swift._supports_developer_dir` here, doing this would likely result in non-hermetic debug info

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -201,3 +134,110 @@ this SDK is paired with.
...
+            # local and `System.loadLibrary` could not bind them. Pass a linker
+            # version script to hide the runtime symbols if desired.
+            linkopts = _build_list([
+                "{}/android/{}/swiftrt.o".format(resource_dir, arch),
+                "-L{}/android".format(resource_dir),
+                "-ldl",
+                "-llog",
+                "-lm",
+                "-lstdc++",
+                "-Wl,-z,max-page-size=16384",
```
keith, line 198: shouldn't all these be coming from c++ toolchain instead? or which ones are swift specific?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -201,3 +134,110 @@ this SDK is paired with.
...
+            sdkroot = "",
+            suffix = arch,
+            swift_version = repository_ctx.attr.swift_version,
+        )
+
+    repository_ctx.file("BUILD.bazel", build_content)
+
+swift_android_sdk_repository = repository_rule(
+    attrs = _common_attrs() | {
+        "host_swiftc": attr.label(
```
keith: naming seems off? i assume this doesn't mean host in the way bazel thinks of the host?

swift/internal/extensions/swift_sdks.bzl
```diff
@@ -201,3 +134,110 @@ this SDK is paired with.
...
+""",
+            mandatory = True,
+        ),
+    },
+    doc = """\
+Downloads the Android Swift SDK artifact bundle and defines Swift toolchains that
+target `{aarch64,x86_64}-unknown-linux-android`. C/C++ compilation and linking go
+through a separately-registered Android C++ cc toolchain (e.g. `@androidndk//:all`
+from `hermetic_android_toolchains`); the Swift toolchain reads that toolchain's
+sysroot at analysis time.
```
keith:
```suggestion
Downloads the Android Swift SDK artifact bundle and defines Swift toolchains that
target Android.
```

swift/internal/extensions/toolchains.bzl
```diff
@@ -83,6 +75,39 @@ def toolchains_for_platform(platform, toolchain_repository):
         toolchain_repository = toolchain_repository,
     )
 
+def android_sdk_toolchains_for_platform(platform, sdk_repository, archs):
+    """Returns Swift `toolchain` declarations for an Android Swift SDK.
+
+    Only the Swift toolchain is declared; C/C++ compilation and linking for
+    Android targets use a separately registered cc toolchain (e.g.
+    `@androidndk//:all`).
+
```
keith:
```suggestion
```

swift/internal/extensions/toolchains.bzl
```diff
@@ -83,6 +75,39 @@ def toolchains_for_platform(platform, toolchain_repository):
...
+    Args:
+        platform: The host platform name (e.g. "xcode" or "ubuntu22.04") whose
+            standalone toolchain the SDK is paired with.
```
keith:
```suggestion
        platform: The platform name (e.g. "xcode" or "ubuntu22.04") whose
            standalone toolchain the SDK is paired with.
```

swift/internal/extensions/toolchains.bzl
```diff
@@ -83,6 +75,39 @@ def toolchains_for_platform(platform, toolchain_repository):
...
+        sdk_repository: The name of the repository created by
+            `swift_android_sdk_repository`.
+        archs: The Android architectures ("aarch64", "x86_64") to declare
+            toolchains for.
```
keith:
```suggestion
        archs: The Android architectures (e.g. aarch64) to declare
            toolchains for.
```

swift/internal/extensions/toolchains.bzl
```diff
@@ -43,9 +43,10 @@ toolchain(
 
 """
 
-# buildifier: disable=unused-variable
+# Just the Swift toolchain: Android's C/C++ toolchain is registered separately
+# (e.g. `@androidndk//:all`).
```
keith:
```suggestion
```

swift/toolchains/swift_toolchain.bzl
```diff
@@ -503,12 +520,17 @@ def _swift_toolchain_impl(ctx):
         target_triples.parse(ctx.var.get("CC_TARGET_TRIPLE") or target_system_name),
     )
 
-    if "clang" not in cc_toolchain.compiler:
-        fail("Swift requires the configured CC toolchain use clang. " +
+    # Swift can only link through an LLVM/clang cc toolchain, not gcc. Accept any
+    # toolchain whose compiler identifies as clang or LLVM (e.g. the Android NDK's
+    # `@androidndk//:all` reports `llvm-c++`).
```
keith:
```suggestion
```

swift/toolchains/swift_toolchain.bzl
```diff
@@ -503,12 +520,17 @@ def _swift_toolchain_impl(ctx):
...
+    if "clang" not in cc_toolchain.compiler and "llvm" not in cc_toolchain.compiler:
+        fail("Swift requires the configured CC toolchain use clang/LLVM. " +
```
keith:
```suggestion
        fail("Swift requires the configured CC toolchain use clang. " +
```

examples/cross_compilation/android_app/BUILD.bazel
```diff
@@ -0,0 +1,91 @@
...
+load("//examples/embedded:transition.bzl", "transition_binary")
+load("//swift/toolchains:android_runtime_lib.bzl", "select_android_runtime_lib")
+
+package(default_visibility = ["//visibility:private"])
+
+# The NDK's shared C++ runtime, selected from the resolved Android cc toolchain
+# (`@androidndk//:all`) and built for arm64 so it bundles into the APK alongside
+# the Swift .so. (The NDK clang links libc++ dynamically, so the app must ship
+# it; rules_swift's `select_android_runtime_lib` locates it in the cc toolchain.)
+select_android_runtime_lib(
```
keith: if this is always required shouldn't we do this implicitly for folks instead of making it happen in BUILD files? (see other comments about how im not sure this should be always required though)

swift/toolchains/swift_toolchain.bzl
```diff
@@ -549,10 +571,24 @@ def _swift_toolchain_impl(ctx):
         # own `-L`/rpath and `swiftrt.o`); for a cross-compiled SDK target they
         # are at best wrong and at worst invalid (wasm-ld rejects `-pie`, and a
         # second `swiftrt.o` would conflict with the SDK's own).
+        linkopts = list(ctx.attr.linkopts)
```
keith: looks to me like this logic should be based on os == android and not this other stuff, and then abstracted into another function

swift/extensions.bzl
```diff
@@ -94,6 +187,42 @@ def _standalone_toolchain_impl(module_ctx):
...
+    doc = """\
+Downloads the Android Swift SDK matching a `toolchain` tag's Swift version and
+defines Swift toolchains targeting `aarch64-unknown-linux-android` and
+`x86_64-unknown-linux-android`.
+
+C/C++ compilation and linking for those targets go through a separately
+registered Android C++ cc toolchain, which also provides the sysroot the Swift
+compiler reads. Register one alongside the Swift toolchains -- for example
+`@androidndk//:all` from
+[hermetic_android_toolchains](https://github.com/keith/hermetic_android_toolchains):
```
keith:
```suggestion
Downloads the Android Swift SDK matching a `toolchain` tag's Swift version and
defines Swift toolchains targeting Android.
```

swift/extensions.bzl
```diff
@@ -94,6 +187,42 @@ def _standalone_toolchain_impl(module_ctx):
...
+
+```starlark
+register_toolchains(
+    "@swift_toolchain//:swift_toolchain_android_aarch64_xcode",
+    "@androidndk//:all",
+)
+```
+
+and build with a platform that has the `@platforms//os:android` and
+`@platforms//cpu:aarch64` (or `x86_64`) constraints.
```
keith:
```suggestion
```

swift/extensions.bzl
```diff
@@ -40,6 +51,62 @@ def _non_module_deps_impl(module_ctx):
...
+            fail("No known Android Swift SDK for version `{}`. Please choose one of {}, or provide the SDK's sha256.".format(
+                swift_version,
+                SWIFT_SDK_RELEASES.keys(),
+            ))
+        sha256 = SWIFT_SDK_RELEASES[swift_version]["android"]
+
+    # The Swift Android toolchain compiles and links C/C++ through a separately
+    # registered Android C++ cc toolchain (e.g. `@androidndk//:all` from
+    # `hermetic_android_toolchains`) and reads that toolchain's sysroot at
+    # analysis time, so rules_swift does not fetch or manage the NDK itself.
```
keith:
```suggestion
```

man the llm is absolutely obsessed with adding this context

AttilaTheFun: Lol yeah I told Claude to clean up the verbose comments but it ignored that here

swift/internal/extensions/swift_sdks.bzl
AttilaTheFun: Renamed to paired_swiftc and the doc comment to "The `swiftc` of the standalone toolchain this SDK is paired with, used to locate the clang builtin headers that match the clang embedded in the Swift compiler"

MODULE.bazel
AttilaTheFun: yeah it's linked now -- the issue is https://github.com/bazelbuild/rules_android/issues/485

swift/internal/extensions/swift_sdks.bzl
AttilaTheFun: The link does go through the cc toolchain — `register_link_binary_action` calls `cc_common.link(cc_toolchain = toolchains.cc, …)` so clang already pulls in the Android defaults (crt, libc++, libm, libdl). What's left here is what it doesn't add:

- Swift-specific: `swiftrt.o` and `-L<resource>/android` (the Swift static stdlib; `libswiftCore` comes in via autolink-extract). These match what `swiftc` adds for a `-static-stdlib` Android link.
- Not Swift-specific, but the NDK toolchain doesn't add them: `-llog` (the NDK never auto-links liblog — your `basic_app` lists it too), `-ldl` (probably redundant now), and `-Wl,-z,max-page-size=16384` (Android 15 16 KB pages).

`-lm`/`-lstdc++` were redundant with the clang driver, so I dropped them. The rest I kept explicit rather than assume a given cc toolchain provides them.

swift/internal/extensions/swift_sdks.bzl
AttilaTheFun: It's used in two places that have to stay in sync: the toolchains-hub BUILD is generated in the module extension *before* the SDK is downloaded, and the per-arch toolchains are generated in the repo rule *after*.

Only the second one could readdir the bundle, but the first needs the arch list up front, so it has to be known statically either way. The SDK's set is small and stable (aarch64 + x86_64; it also ships armv7, which `@platforms//cpu` can't cleanly express), so I kept it as a constant.

swift/toolchains/android_runtime_lib.bzl
AttilaTheFun: You're right about normal NDK C++ — it can static-link libc++ (libc++_static / APP_STL=c++_static), so nothing shows up as a NEEDED, exactly like your ldd. 
It's Swift-specific: the swift.org Android SDK's link recipe (`swift_static-<arch>/android/static-stdlib-args.lnk`) uses `-lstdc++`, i.e. the *shared* libc++, so a Swift JNI `.so` ends up with a real `NEEDED libc++_shared.so` and the APK has to ship it:

```
% llvm-readelf -d libSwiftJNI.so | grep NEEDED
  (NEEDED) Shared library: [libm.so]
  (NEEDED) Shared library: [libdl.so]
  (NEEDED) Shared library: [liblog.so]
  (NEEDED) Shared library: [libc++_shared.so]
  (NEEDED) Shared library: [libc.so]
```

I kept `-lstdc++` (rather than letting it fall back to static libc++) because static-linking libc++ into a JNI library is the thing the NDK warns against once there's more than one `.so` in the app. So the packaging is genuinely required — I just fixed the docstring, which previously implied "or other NDK C++" and gave
the wrong impression that it's a general NDK thing.

keith: i think unless rules_android_ndk does something to avoid the potential duplicate linking of static libstdc++s we probably shouldn't either 🤔 

idk what the right call is here but seems to me like we should act as much as we can like the normal ndk workflows

keith: i assume we could add `-static-libstdc++` to the linker options after whatever swift ones there are  based on https://developer.android.com/ndk/guides/cpp-support

examples/cross_compilation/android_app/android.MODULE.bazel
```diff
@@ -0,0 +1,77 @@
...
+#
+# `android_ndk` (the hermetic_android_toolchains extension) and `@androidndk` are
+# defined in the root module above this include().
+
+bazel_dep(name = "rules_android", version = "0.7.3", dev_dependency = True)
+bazel_dep(name = "rules_java", version = "9.3.0", dev_dependency = True)
+bazel_dep(name = "rules_kotlin", version = "2.2.2", dev_dependency = True)
+bazel_dep(name = "rules_jvm_external", version = "7.0", dev_dependency = True)
+
+# rules_java's hermetic JDK + java tools for rules_android's Java actions.
```
keith:
```suggestion
```

keith (approved):

PR #1819, opened 2026-06-23
Build and test Swift natively on a Windows host

AttilaTheFun:
Windows host toolchain support + miscellaneous fixes, part of the #1809 split, on
top of the now-merged foundation PRs #1820 (`swift_binary(linkshared)`) and #1821
(Swift-SDK framework).

## What this adds

Native Windows host toolchain support on the standalone toolchain path (the
`<arch>-unknown-windows-msvc` triple, MSVC `link.exe` semantics, auto-configuring
a toolchain from an installed `swiftc`), plus independent `test_observer` /
worker fixes (extended-length paths, space-in-path object names, corelibs-xctest
runner) that were bundled into the original combined PR. CI gains a Windows task
building `//tools/...` + the xplatform examples and running the `xctest` example.

## Open questions

- **MSVC vs. hermetic LLVM**: the Windows path assumes an MSVC-compatible linker;
  both `link.exe` and a hermetic `lld-link` accept the `/ALTERNATENAME` /
  `/IGNORE:4217` flags used here. I can key the flags off the linker rather than
  the OS if you want it to compose with `clang-cl` explicitly. (@dzbarsky)
- **Python in `swift_autoconfiguration`** parses the SDK `Info.plist` (a plist,
  not JSON) in a repository rule. The clean fix is to parse `XCTEST_VERSION` in
  Starlark and drop Python; I'd do it in a follow-up CI can validate, since I
  can't exercise the Windows path locally. (@dzbarsky)


swift/internal/compiling.bzl
```diff
@@ -1667,7 +1667,14 @@ def _declare_per_source_output_file(actions, extension, target_name, src):
...
     objs_dir = "{}_objs".format(target_name)
-    owner_rel_path = owner_relative_path(src)
+
+    # Spaces in object file paths break response-file parsing for the Windows
+    # archiver and linker (`lib.exe`/`link.exe`), which treat an unquoted space
+    # as an argument separator. Sanitize them in the derived output path so that
+    # targets whose sources live in directories containing spaces (for example
+    # swift-argument-parser's "Parsable Properties") can be archived and linked
+    # on Windows. Paths without spaces are unaffected.
+    owner_rel_path = owner_relative_path(src).replace(" ", "_")
```
keith:
```suggestion
    # Spaces in object file paths break response-file parsing on Windows
    owner_rel_path = owner_relative_path(src).replace(" ", "_")
```

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -269,15 +285,27 @@ xcode_swift_toolchain(
...
+    """Returns True if `python_bin` is a real, runnable Python interpreter.
+
+    On Windows, `python3.exe`/`python.exe` found on `PATH` are frequently the
+    Microsoft Store "App execution alias" stubs rather than real interpreters:
+    when run non-interactively they print a message pointing at the Store and
+    exit nonzero. Probe the candidate so those stubs are skipped in favor of a
+    working interpreter later on `PATH`.
+    """
+    if not python_bin:
+        return False
```
keith:
```suggestion
```

i think this function should take only real executables and we should check this before calling this

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -269,15 +285,27 @@ xcode_swift_toolchain(
...
         return repository_ctx.os.environ.get("PYTHON_BIN_PATH").strip()
-    out = repository_ctx.which("python3.exe")
-    if out:
-        return out
-    out = repository_ctx.which("python.exe")
-    if out:
-        return out
+    for name in ("python3.exe", "python.exe", "python3", "python"):
+        candidate = repository_ctx.which(name)
+        if _python_executable_works(repository_ctx, candidate):
```
keith:
```suggestion
        candidate = repository_ctx.which(name)
        if candidate and _python_executable_works(repository_ctx, candidate):
```

related to my suggestion above

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -269,15 +285,27 @@ xcode_swift_toolchain(
...
 def _get_python_bin(repository_ctx):
     if "PYTHON_BIN_PATH" in repository_ctx.os.environ:
         return repository_ctx.os.environ.get("PYTHON_BIN_PATH").strip()
-    out = repository_ctx.which("python3.exe")
-    if out:
-        return out
-    out = repository_ctx.which("python.exe")
-    if out:
-        return out
+    for name in ("python3.exe", "python.exe", "python3", "python"):
```
keith:
```suggestion
    for name in ("python3.exe", "python3", "python.exe", "python"):
```

would this make sense? can `python` w/o the 3 ever be python2? 😬

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -305,22 +334,39 @@ Swift toolchain.
...
     version_file, parsed_version = _write_swift_version(repository_ctx, path_to_swiftc)
+
+    # Normalize SDKROOT to forward slashes with no trailing separator: the raw
+    # environment value typically ends in a backslash, which is both invalid at
+    # the end of a Python raw-string literal (used below) and produces doubled
+    # separators when joined.
+    sdkroot = repository_ctx.os.environ["SDKROOT"].replace("\\", "/").rstrip("/")
+
+    # The platform `Info.plist` (which records the bundled XCTest version) sits
+    # three levels above the SDK root, i.e. `<platform>/Info.plist`.
```
keith:
```suggestion
```

swift/internal/swift_autoconfiguration.bzl
```diff
@@ -305,22 +334,39 @@ Swift toolchain.
...
+        "import plistlib; " +
+        "print(plistlib.load(open(r'{}', 'rb'))['DefaultProperties']['XCTEST_VERSION'])".format(info_plist),
     ])
 
     env = {
         "Path": repository_ctx.os.environ["Path"] if "Path" in repository_ctx.os.environ else repository_ctx.os.environ["PATH"],
-        "ProgramData": repository_ctx.os.environ["ProgramData"],
+        # `ProgramData` is normally present in a Windows process environment, but
+        # is not guaranteed to be (e.g. a service-account CI agent), so fall back
+        # to its conventional value rather than failing toolchain configuration.
```
keith:
```suggestion
```

swift/toolchains/swift_toolchain.bzl
```diff
@@ -488,6 +508,13 @@ def _parse_target_system_name(*, arch, os, target_system_name):
 
     if os == "linux":
         return "%s-unknown-linux-gnu" % arch
+    elif os == "windows":
+        # The MSVC cc toolchain reports a `target_gnu_system_name` of "local",
+        # so synthesize the triple. The `-msvc` environment is required: while
+        # `swiftc` defaults to it, `swift-symbolgraph-extract` (used for XCTest
+        # test discovery) matches the module layout's triple exactly and fails
+        # to load modules built for `*-windows-msvc` if it is omitted.
```
keith:
```suggestion
        # The MSVC cc toolchain reports a `target_gnu_system_name` of "local",
        # so synthesize the triple.
```

swift/toolchains/swift_toolchain.bzl
```diff
@@ -503,7 +530,12 @@ def _swift_toolchain_impl(ctx):
         target_triples.parse(ctx.var.get("CC_TARGET_TRIPLE") or target_system_name),
     )
 
-    if "clang" not in cc_toolchain.compiler:
+    # On Windows the Swift toolchain composes with Bazel's MSVC C++ toolchain
+    # (`msvc-cl`): `swiftc` uses its bundled clang for the clang importer and
+    # links via MSVC `link.exe`, so a clang CC toolchain is neither present nor
+    # required. Elsewhere (Linux), Swift drives the configured cc toolchain for
+    # C/C++ interop and linking, which must be clang.
```
keith:
```suggestion
```

i don't think this adds much.

this is a bit suspect to me though. i think in the linux case it's more that only clang successfully links swift and gcc would not, i imagine that could be true for windows if you had a project that was interoping more languages and eventually linking a cc_binary or something, but we can deal with that as it comes up

swift/swift_test.bzl
```diff
@@ -109,6 +110,9 @@ def _generate_test_discovery_srcs(
     Args:
         actions: The context's actions object.
         deps: The list of direct dependencies of the test target.
+        env: Environment variables to set when running the discovery tool. On
```
keith, line 113: shouldn't this come from the toolchain or something instead? wouldn't this be non-hermetic to be a full on path?

swift/toolchains/swift_toolchain.bzl
```diff
@@ -520,7 +544,7 @@ def _swift_toolchain_impl(ctx):
         target_triples.parse(ctx.var.get("CC_TARGET_TRIPLE") or target_system_name),
     )
 
-    if "clang" not in cc_toolchain.compiler and "llvm" not in cc_toolchain.compiler:
+    if ctx.attr.os != "windows" and "clang" not in cc_toolchain.compiler and "llvm" not in cc_toolchain.compiler:
```
keith, line 547: what is the value that we're ignoring here? i think swift still kinda wants to be linked by clang on windows

swift/swift_test.bzl
AttilaTheFun: It does come from the toolchain. The discovery action's env is toolchains.swift.test_configuration.env, which the autoconfigured Windows toolchain populates (a Path with the Swift runtime DLL dir). The absolute path is inherent to autoconfiguring from a host-installed Swift. A future hermetic downloaded Windows toolchain would carry a runfiles-relative path through the same plumbing.

swift/toolchains/swift_toolchain.bzl
AttilaTheFun: The ignored value is msvc-cl which is what Bazel's autoconfigured Windows cc toolchain reports. (With BAZEL_COMPILER=clang-cl it reports clang-cl, which passes the check on its own, and lld-link accepts the same /ALTERNATENAME / /IGNORE:4217 flags, so that composes too.)

Re: "swift wants to be linked by clang", that's true on ELF, where clang mediates autolinking. On COFF the compiler embeds its /DEFAULTLIB directives in the objects' .drectve sections, which link.exe honors natively (also why this branch doesn't enable autolink-extract on Windows). 

The toolchain supplies the rest MSVC-style: swiftrt.obj, the SDK -LIBPATH:s, /ALTERNATENAME:main=..., -IGNORE:4217. Windows CI links a swift_binary, a linkshared DLL, and the XCTest runner through link.exe and runs the test suite. If a mixed-language cc_binary case turns up that needs clang to mediate, it fails loudly at link time and we can deal with it then.

keith (approved):

PR #1821, opened 2026-06-23
Add shared framework for Swift SDK cross-compilation toolchains

AttilaTheFun:
Second foundation PR from the #1809 split (see #1820 for the linkshared half).
This adds the platform-agnostic machinery the per-platform Swift SDK PRs build
on:

```
main ─┬─ #1820 linkshared
      └─ Swift SDK framework (this PR)
            ├─ WebAssembly
            ├─ Android
            └─ Windows host
```

## What this adds

The shared building blocks for downloading a swift.org "Swift SDK" artifact
bundle and turning it into Bazel toolchains:

- **`swift_toolchain`** gains `linkopts` / `linker_inputs` attributes (and
  `_swift_sdk_linkopts_cc_info`) so a toolchain can link against a Swift runtime
  provided by an SDK rather than the host toolchain.
- **`swift_sdks.bzl`**: helpers and BUILD-file templates a Swift SDK repository
  rule uses (bundle download, `cc_tool` / `cc_toolchain` / `swift_toolchain`
  templates).
- **`toolchains.bzl`**: the `_SDK_TOOLCHAIN_PLATFORM` template and a shared
  exec-constraint helper for generating `toolchain()` declarations.
- **`swift_sdk_releases.bzl`**: the SDK download-URL helper and the (initially
  empty) `SWIFT_SDK_RELEASES` checksum registry the platform extensions
  populate.

These are deliberately a foundation: they're exercised end-to-end by the stacked
platform PRs (WebAssembly / Android / Windows). A unit test covers
`swift_sdk_download_url` in `//test:utils`.

### DIFF
--- a/swift/toolchains/swift_toolchain.bzl
+++ b/swift/toolchains/swift_toolchain.bzl
@@ -443,6 +443,37 @@ def _swift_unix_linkopts_cc_info(
         ),
     )
 
+def _swift_sdk_linkopts_cc_info(
+        toolchain_label,
+        linkopts,
+        linker_inputs):
+    """Returns a `CcInfo` with linker flags supplied by the toolchain target.
+
+    Used for toolchains whose Swift runtime libraries come from a Swift SDK
+    rather than from the host toolchain or system; the repository that defines
+    the toolchain provides the exact search paths and runtime objects to link.
+
+    Args:
+        toolchain_label: The label that owns the propagated linker input.
+        linkopts: Linker flags from the toolchain's `linkopts` attribute.
+        linker_inputs: `File`s that link actions using these flags need.
+
+    Returns:
+        A `CcInfo` that propagates the flags to binaries depending on Swift
+        targets.
+    """
+    return CcInfo(
+        linking_context = cc_common.create_linking_context(
+            linker_inputs = depset([
+                cc_common.create_linker_input(
+                    owner = toolchain_label,
+                    user_link_flags = depset(linkopts),
+                    additional_inputs = depset(linker_inputs),
+                ),
+            ]),
+        ),
+    )
+
 def _entry_point_linkopts_provider(*, entry_point_name):
     """Returns linkopts to customize the entry point of a binary."""
     return struct(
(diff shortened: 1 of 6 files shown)


swift/toolchains/swift_toolchain.bzl
```diff
@@ -509,6 +540,12 @@ def _swift_toolchain_impl(ctx):
         )
     elif ctx.attr.os == "none":
         swift_linkopts_cc_info = CcInfo()
+    elif ctx.attr.linkopts or ctx.attr.linker_inputs:
```
keith, line 543: seems to me like this should be in addition to the other things in this if statement instead of in place of? i don't think the naming `linkopts` / `linker_inputs` implies that it will overwrite any default behavior. would that still work? maybe we do need to change the `else` to an `elif` for something unix specific though?

keith (approved):
lgtm, small thing

swift/toolchains/swift_toolchain.bzl
AttilaTheFun: So, additive would actually break the cross-compiled targets. I kept the replace but made the reasoning explicit in a comment + the attr doc.

`_swift_unix_linkopts_cc_info` emits flags for a Linux host toolchain: `-pie`, `-static-libgcc`, `-lrt`, and the host toolchain's own `-L`/rpath and `swiftrt.o`. For an SDK target (WebAssembly/Android) the runtime comes from the SDK, so adding those on top is wrong and in places invalid — wasm-ld rejects `-pie`, `-lrt`/`-static-libgcc` aren't available, and the host `swiftrt.o` would collide with the SDK's own. So `linkopts`/`linker_inputs` have to be the complete runtime set, replacing the computed defaults rather than augmenting them.

To your second point: the SDK case is already its own branch (the `elif`), so it never falls through to the unix `else`. I've added a comment at the branch and rewritten the `linkopts` doc to say plainly that it's a complete replacement and why. Happy to rename the attrs (e.g. `runtime_linkopts`) in one of the subsequent PRs if you think that reads better than the doc.

keith (approved):

swift/toolchains/swift_toolchain.bzl
keith: ok we can always mess with it later as needed

PR #1822, opened 2026-06-25
Add JNI test example

keith:
This is a similar pattern to the android bridge


keith:
with the android example i don't think we really need this, happy to add if folks think otherwise

PR #1827, opened 2026-06-30
Support cross-compiling to Android with a Swift SDK

keith:
Co-authored-by: Logan Shire <logan.shire@gmail.com>


keith:
depends on https://github.com/bazelbuild/rules_swift/pull/1825

keith:
based on https://github.com/bazelbuild/rules_swift/pull/1818

keith:
i think we might end up merging that other one instead, we pushed this stuff there

PR #1829, opened 2026-07-03
Correctly forward external_include_paths to swift compiler

DoDoENT:
This fix (authored by codex) fixes the building of swift code that depends on C++ code that has `external_include_paths` feature enabled.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -50,4 +50,7 @@ build --per_file_copt="external/.*zlib.*/.*.c@-Wno-deprecated-non-prototype"
 build --per_file_copt="external/.*protobuf.*/.*@-Wno-unused-private-field"
 build --host_per_file_copt="external/.*protobuf.*/.*@-Wno-unused-private-field"
 
+# ignore warnings in external CC dependencies
+build --features="external_include_paths"
+
 try-import %workspace%/user.bazelrc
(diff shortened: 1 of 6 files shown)


adincebic (changes requested):
Can you add a test that fails without this change?

DoDoENT:
> Can you add a test that fails without this change?

Added. Also updated `.bazelrc` to enable this `external_include_paths` feature, which ensures that no warnings are reported from external headers and was broken in swift/c++ interop prior to this PR.

.bazelrc
```diff
@@ -44,4 +44,7 @@ build --per_file_copt="external/.*zlib.*/.*.c@-Wno-deprecated-non-prototype"
 build --per_file_copt="external/.*protobuf.*/.*@-Wno-unused-private-field"
 build --host_per_file_copt="external/.*protobuf.*/.*@-Wno-unused-private-field"
 
+# ignore warnings in external CC dependencies
+build --features="external_include_paths"
```
keith:
```suggestion
build --features=external_include_paths --host_features=external_include_paths
```

keith (approved):
nice!

adincebic (approved):
Thanks.

PR #1833, opened 2026-07-08
Remove toolchain_name from sdk module extensions

keith:
Currently the main toolchain() extension fails if you call it more than
once, so there is only ever 1 name. If we ever allow multiple names we
can add something like this back but should potentially use the version
instead to dedup.


keith:
cc @AttilaTheFun

AttilaTheFun:
@keith After dropping the attr everywhere in my stack: full `bazel build //...` (212 targets — macOS/iOS apps, wasm32-wasip1 reactors, Android APKs through the platform transition) completes clean, and every action cache-hits against my baseline (c6cad94).

aaronsky (approved):

PR #1834, opened 2026-07-08
Move swift shas to json files

keith:
This makes generating this file easy for updates over time, now on a
cron


keith:
Currently excludes 6.3.3 so i can test the cron

adincebic (approved):
Nice.

PR #1835, opened 2026-07-08
Move wasm templating to separate BUILD file

keith:
This makes it easier to reason about, there are very few things that are
templated.


keith:
cc @AttilaTheFun

AttilaTheFun:
@keith Tested this against my stack. No issues: drop-in, no consumer-side changes needed. Full `bazel build //...` (212 targets incl. wasm32-wasip1 reactors and Android APKs) completes clean at this PR's head, and all actions cache-hit against my baseline pin (c6cad94).

adincebic (approved):

PR #1843, opened 2026-07-09
Expand target list on CI

keith:
This probably doesn't change much but I noticed that some docs case
fails with `last_green` that doesn't fail in the docs job since it's
using `latest`


keith:
depends on https://github.com/bazelbuild/rules_swift/pull/1844

adincebic (approved):

PR #1851, opened 2026-07-10
Drop additional_rpaths toolchain logic

keith:
These rpaths never seem to be valid for hermetic toolchains. The macOS hermetic
toolchain doesn't work at all right now. I think to fix that we'll need to let bazel
handle these with cc_imports or something instead.


keith:
trying to verify we're not missing some case here, but these rpaths just seem invalid

keith:
this does matter, currently binaries don't use this but tests do. im looking into a wider version of this fix in https://github.com/bazelbuild/rules_swift/pull/1853

PR #1852, opened 2026-07-10
test embedded CI fails

keith:


keith:
i wanted to prove this failed CI on macOS, which it does

PR #1856, opened 2026-07-13
Fix WMO num-threads detection

qyang-nj:
## Summary

- use the shared Swift compiler-option feature detector for the Xcode toolchain
- remove the divergent helper that classified `-num-threads 1` as `-num-threads 0`
- add an action-level regression test for single-threaded WMO
- correct the WMO output-shape documentation

## Root cause

The Xcode toolchain called a duplicate helper that enabled `SWIFT_FEATURE__NUM_THREADS_0_IN_SWIFTCOPTS` when the supplied value was `1`. As a result, `-num-threads 0` was not recognized and rules_swift could inject its default `-num-threads 12`, while also reasoning about the wrong output shape.

### DIFF
--- a/swift/internal/wmo.bzl
+++ b/swift/internal/wmo.bzl
@@ -89,30 +89,6 @@ def is_wmo_manually_requested(user_compile_flags):
             return True
     return False
 
-def wmo_features_from_swiftcopts(swiftcopts):
-    """Returns a list of features to enable based on `--swiftcopt` flags.
-
-    Since `--swiftcopt` flags are hooked into the action configuration when the
-    toolchain is configured, it's not possible for individual actions to query
-    them easily if those flags may determine the nature of outputs (for example,
-    single- vs. multi-threaded WMO). The toolchain can call this function to map
-    those flags to private features that can be queried instead.
-
-    Args:
-        swiftcopts: The list of command line flags that were passed using
-            `--swiftcopt`.
-
-    Returns:
-        A list (possibly empty) of strings denoting feature names that should be
-        enabled on the toolchain.
-    """
-    features = []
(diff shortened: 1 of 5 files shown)


swift/internal/wmo.bzl
```diff
@@ -89,30 +89,6 @@ def is_wmo_manually_requested(user_compile_flags):
...
-            `--swiftcopt`.
-
-    Returns:
-        A list (possibly empty) of strings denoting feature names that should be
-        enabled on the toolchain.
-    """
-    features = []
-    if is_wmo_manually_requested(user_compile_flags = swiftcopts):
-        features.append(SWIFT_FEATURE__WMO_IN_SWIFTCOPTS)
-    if find_num_threads_flag_value(user_compile_flags = swiftcopts) == 1:
```
qyang-nj, line 112: This line should be `if ... == 0`. After the correction, it will be duplication of above `features_from_swiftcopts` function, so we can just delete it.

qyang-nj:
@keith PTAL 🙏🏻

keith (approved):
lgtm, but can you add a default test that asserts the `-num-threads 12` is passed by default?

keith (approved):
there's something weird still about how we handle 0 vs 1, maybe we should be handling 0 or 1 as the same, but w/e we can deal with that as it comes up

qyang-nj:
I believe `-num-threads 0` is the flag for single thread mode (**one `.o` per module**), while `-num-threads 1` enables multiple thread mode (**one `.o` per file**) but with one working thread, so they are very different.

keith:
ah right, i keep forgetting that, great, thanks!

PR #1858, opened 2026-07-14
Android: match the NDK's API level in Swift's target triple (alt to #1857)

AttilaTheFun:
### Alternative to #1857

Same underlying fix as #1857 — the Android NDK cc toolchain reports a level-less `target_gnu_system_name` (`aarch64-linux-android`), so Swift floors availability checking below the SDK's real minimum and rejects APIs the target actually has (e.g. `swift-log`'s `stdout`, gated on Android 23). Both PRs hand `swift_toolchain` a leveled Android triple to fix it.

The difference is **where the API level comes from**, which is exactly the question raised in #1857 ([comment](https://github.com/bazelbuild/rules_swift/pull/1857#issuecomment-4974493636)): *derive it, or match what the user's NDK is set up for?* This PR matches the NDK:

- **`swift.android_sdk(api_level = N)`** selects `<arch>-unknown-linux-android<N>` from the bundle — which ships the whole ladder the Swift release supports (e.g. 28–36) — so Swift compiles for the same level the NDK/`minSdkVersion` drives clang's `--target` to. Set it to your NDK's `api_level`.
- **When `api_level` is omitted**, the bundle's *lowest* supported level is used. That's the safe minimum: the prebuilt runtime won't load below it, and a min-N binary runs on any device ≥ N.
- **`swift_toolchain` never overrides a cc triple that already carries an API level.** The SDK-provided triple only fills in a level the cc toolchain omits, so a toolchain configured with an explicit level wins outright.

So vs. #1857 (which always uses the bundle's floor and always overrides): identical behavior for the common level-less case, plus an `api_level` knob to track the NDK and a don't-override rule so Swift can't silently retarget away from the level the rest of the build uses.

### Changes

- `swift.android_sdk` / `swift_android_sdk_repository` gain an optional `api_level`.
- `_sdk_triple_for_arch` selects the bundle triple for that level, or the deterministic minimum when unset.
- `swift_toolchain` gains `target_system_name`, applied only when the cc triple lacks an API level (`_has_android_api_level`).

### Testing

- `//test:android_swiftcompile_targets_android` updated to assert the leveled `-target aarch64-unknown-linux-android28` (the floor default), and passes.
- `//test:android_link_uses_ndk`, `//test:android_apk_contents` (real APK build), and `//test:const_values_expected_argv` pass.
- Verified end-to-end against a Swift-on-Android app (NIO gRPC client) that previously failed the `swift-log` availability check.

Happy to go with whichever of the two you prefer — closing this or #1857 accordingly.

### DIFF
--- a/swift/extensions.bzl
+++ b/swift/extensions.bzl
@@ -40,7 +40,7 @@ def _known_sdk_versions(swift_sdk_releases, sdk):
         if sdk in swift_sdk_releases[version]
     ]
 
-def _setup_android_sdk(*, tag, toolchain_name, swift_version, platforms, swift_sdk_releases):
+def _setup_android_sdk(*, tag, toolchain_name, swift_version, platforms, swift_sdk_releases, api_level):
     """Creates the repositories for a `swift.android_sdk` tag.
 
     Args:
@@ -49,6 +49,8 @@ def _setup_android_sdk(*, tag, toolchain_name, swift_version, platforms, swift_s
         swift_version: The Swift release version of that toolchain.
         platforms: The host platforms the toolchain was created for.
         swift_sdk_releases: Bundled Swift SDK checksum metadata by release.
+        api_level: The Android API level to target, or empty for the SDK's
+            lowest supported level.
 
     Returns:
         BUILD file content with the `toolchain` declarations to add to the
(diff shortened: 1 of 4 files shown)


keith:
https://github.com/bazelbuild/rules_swift/pull/1859

PR #1859, opened 2026-07-14
Fix incorrect triple for Android

keith:
Previously we were getting the triple which doesn't contain the API
level which gates some API


keith:
this is an alternative to https://github.com/bazelbuild/rules_swift/pull/1858 / https://github.com/bazelbuild/rules_swift/pull/1857

luispadron (approved):

PR #1860, opened 2026-07-16
Add wasm32-unknown-wasip1-threads variant to swift.wasm_sdk

AttilaTheFun:
## What

Adds a `threads` option to the `swift.wasm_sdk` module extension so a consumer
can target `wasm32-unknown-wasip1-threads` (WebAssembly with shared memory,
atomics, and wasi-threads) in addition to the single-threaded
`wasm32-unknown-wasip1`. This unblocks building threaded Swift-wasm reactors
that run with real multicore parallelism (e.g. under WAMR with wasi-threads) via
`bazel build`.

To support the second bundle flavor without a second set of hardcoded flags and
paths, the wasm repository rule now derives the target triple, SDK paths, and
compiler/linker flags from the metadata the bundle itself ships
(`info.json` / `swift-sdk.json` / `toolset.json`) — details below.

```starlark
swift.wasm_sdk(
    toolchain_name = "swift",
    threads = True,
    url = "https://github.com/swiftwasm/swift/releases/download/swift-wasm-6.3-RELEASE/swift-wasm-6.3-RELEASE-wasm32-unknown-wasip1-threads.artifactbundle.zip",
    sha256 = "…",
)
```

## Why an explicit `url` + `sha256`

swift.org does **not** publish a threads bundle — the computed
`https://download.swift.org/.../swift-<v>-RELEASE_wasm.artifactbundle.tar.gz`
URL 404s for the threads variant. The threads SDK comes from
[swiftwasm](https://github.com/swiftwasm/swift/releases) instead, e.g.
`swift-wasm-6.3-RELEASE-wasm32-unknown-wasip1-threads.artifactbundle.zip`. So the
extension gains `url` and `sha256` attrs that override the computed swift.org URL
(and are required when `threads = True`).

The `url`/`sha256` override is also usable on the non-threads path if you want to
pin a mirror.

## The repository rule is driven by the bundle's own metadata

Rather than hardcoding per-variant flags and directory layouts, the repository
rule now parses the metadata every Swift SDK artifact bundle ships (both the
swift.org and swiftwasm bundles follow the same scheme):

1. **`info.json`** names the bundle's artifacts; the default (non-embedded)
   artifact's `swift-sdk.json` path replaces the previous
   bundle-layout computation/scan. (Embedded Swift variants —
   `embedded-swift-sdk.json` — are ignored.)
2. **`swift-sdk.json`** provides the **target triple** (validated against the
   `threads` attribute, so pointing `threads = True` at a single-threaded
   bundle or vice versa fails with a clear message), plus `sdkRootPath` and
   `swiftStaticResourcesPath`, which replace the previously hardcoded
   `WASI.sdk` / `swift.xctoolchain/usr/lib/swift_static` paths.
3. **`toolset.json`** (via `toolsetPaths`, merged in order) provides the
   per-tool `extraCLIOptions`:
   - `cCompiler`/`cxxCompiler` options → the generated `cc_args` (compile);
     for the threads bundle this is `-matomics -mbulk-memory -mthread-model
     posix -pthread -ftls-model=local-exec`.
   - `swiftCompiler` options → `swift_toolchain` copts (for the threads bundle:
     `-static-stdlib` plus the clang flags via `-Xcc`; for the swift.org
     bundle: `-static-stdlib`).
   - `linker` options are raw `wasm-ld` flags (SwiftPM invokes the linker
     directly), so they are wrapped in `-Wl,` for the clang driver the
     generated toolchains link through (threads bundle: `--import-memory
     --export-memory --shared-memory --max-memory=1073741824`).
   - `rootPath` and per-tool executable overrides are intentionally ignored:
     the generated toolchains always drive the paired standalone toolchain's
     own `swiftc`/`clang`.

The parsing helpers (`_relative_metadata_path`, `_swift_sdk_json_path`,
`_swift_sdk_target_settings`, `merged_toolset_options`,
`linker_options_to_clang_args`) are SDK-kind-agnostic, so the Android
repository rule and the Static Linux SDK proposed in #1813 can share them —
`_relative_metadata_path` is taken verbatim from #1813 so whichever lands
second rebases cleanly onto the other's copy.

## Backwards compatibility

The single-threaded path now reads the same metadata. For the swift.org
`wasm32-unknown-wasip1` bundle the resolved paths are identical and the only
flag delta is that the bundle toolset's `-static-stdlib` is now passed to
`swiftc` — inert for compile actions: rebuilding the wasm example produced
bit-identical objects (all downstream actions cache-hit) and
`//examples/cross_compilation/wasm:reactor_test` passes.

## Verification

- **Single-threaded** (swift.org bundle through the metadata path):
  `//examples/cross_compilation/wasm:reactor_test` passes; the recompiled
  objects are bit-identical to before, so everything downstream cache-hits.
- **Threads, end-to-end** on macOS/arm64: a `withTaskGroup` program compiles
  against the `wasm32-unknown-wasip1-threads` swiftmodules, links a
  shared-memory module, and runs under
  `wasmtime -W threads=y,shared-memory=y -S threads=y`. A peak-concurrency
  probe confirms tasks actually run in parallel (8 workers, peak 4 concurrent
  tasks). After the metadata rework the generated toolchains are byte-identical
  to the previously hand-mirrored flags. Full transcript in the PR comments.
- **Unit tests** for `merged_toolset_options` (accumulation across multiple
  toolsets, both bundle shapes) and `linker_options_to_clang_args` in
  `test/utils_tests.bzl`.

## Files changed

- `swift/extensions.bzl` — `threads` (bool), `url` (string) attrs on the
  `wasm_sdk` tag class; `_setup_wasm_sdk` validation + URL/sha256 selection;
  threading `threads`/`url` through to the repository rule.
- `swift/internal/extensions/swift_sdks.bzl` — metadata parsing
  (`info.json`/`swift-sdk.json`/`toolset.json`) shared helpers; triple
  validation against `threads`; toolset-driven template substitutions.
- `swift/internal/extensions/wasmsdk.BUILD` — resource-dir/sysroot and
  compile/link flag placeholders filled from the SDK metadata.
- `test/utils_tests.bzl` — unit tests for the toolset helpers.

### DIFF
--- a/swift/extensions.bzl
+++ b/swift/extensions.bzl
@@ -93,7 +93,23 @@ def _setup_wasm_sdk(*, tag, toolchain_name, swift_version, platforms, swift_sdk_
         BUILD file content with the `toolchain` declarations to add to the
         toolchains hub repository.
     """
+    threads = tag.threads
+    url = tag.url
     sha256 = tag.sha256
+
+    # swift.org does not publish a `wasm32-unknown-wasip1-threads` SDK bundle, so
+    # the threads variant must be pointed at an explicit bundle (for example a
+    # swiftwasm release) via `url` + `sha256`.
+    if threads and not url:
+        fail("`swift.wasm_sdk` with `threads = True` requires an explicit `url` " +
+             "(and `sha256`), because swift.org does not publish a " +
+             "wasm32-unknown-wasip1-threads SDK bundle. Point it at a swiftwasm " +
+             "release, e.g. https://github.com/swiftwasm/swift/releases.")
+    if url and not sha256:
+        fail("`swift.wasm_sdk` with an explicit `url` also requires `sha256`.")
+
+    if not url:
+        url = swift_sdk_download_url(swift_version, "wasm")
     if not sha256:
         if swift_version not in swift_sdk_releases or "wasm" not in swift_sdk_releases[swift_version]:
             fail("No known WebAssembly Swift SDK for version `{}`. Please choose one of {}, or provide the SDK's sha256.".format(
@@ -109,8 +125,9 @@ def _setup_wasm_sdk(*, tag, toolchain_name, swift_version, platforms, swift_sdk_
             name = repository_name,
             sha256 = sha256,
             swift_version = swift_version,
+            threads = threads,
             toolchain_repo = "{}_{}".format(toolchain_name, platform),
-            url = swift_sdk_download_url(swift_version, "wasm"),
+            url = url,
         )
         build_file_content += wasm_sdk_toolchains_for_platform(
             platform = platform,
@@ -238,13 +255,32 @@ _wasm_sdk = tag_class(
         "sha256": attr.string(
             doc = """\
 The expected SHA-256 of the SDK artifact bundle. May be omitted for Swift
-versions known to this version of rules_swift.
+versions known to this version of rules_swift, but is required when `url` or
+`threads` is set.
+""",
+        ),
+        "threads": attr.bool(
+            default = False,
+            doc = """\
+If `True`, target `wasm32-unknown-wasip1-threads` (WebAssembly with shared
+memory, atomics, and wasi-threads) instead of the single-threaded
+`wasm32-unknown-wasip1`. swift.org does not publish a threads SDK bundle, so
+`url` and `sha256` are required in this mode (for example, a swiftwasm release:
+https://github.com/swiftwasm/swift/releases).
+""",
+        ),
+        "url": attr.string(
+            doc = """\
+An explicit download URL for the SDK artifact bundle, overriding the swift.org
+URL computed from the toolchain's Swift version. Required for the `threads`
+variant. When set, `sha256` is also required.
 """,
         ),
     },
     doc = """\
 Downloads the WebAssembly Swift SDK matching a `toolchain` tag's Swift version
-and defines Swift and C++ toolchains targeting `wasm32-unknown-wasip1`.
+and defines Swift and C++ toolchains targeting `wasm32-unknown-wasip1` (or
+`wasm32-unknown-wasip1-threads` when `threads = True`).
 """,
 )
 
--- a/swift/internal/extensions/swift_sdks.bzl
+++ b/swift/internal/extensions/swift_sdks.bzl
@@ -241,6 +241,7 @@ def _swift_wasm_sdk_impl(repository_ctx):
         "{bundle_dir}": bundle_dir,
         "{sdk_dir}": sdk_dir,
         "{swift_version}": repository_ctx.attr.swift_version,
+        "{target_triple}": triple,
         "{toolchain_repo}": repository_ctx.attr.toolchain_repo,
     }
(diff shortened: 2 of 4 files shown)


AttilaTheFun:
### End-to-end verification ✅

Built and ran a Swift Concurrency program through the `threads = True` SDK produced by this PR (with the `206a15e1` triple fix), on macOS/arm64:

```python
# MODULE.bazel
swift.toolchain(name = "swift_toolchain", swift_version = "6.3")   # match the SDK's swiftmodule format
swift.wasm_sdk(
    toolchain_name = "swift_toolchain",
    threads = True,
    url = "…/swift-wasm-6.3-RELEASE-…-wasip1-threads.artifactbundle.tar.gz",
    sha256 = "…",
)
```

```swift
// Hello.swift
@main struct Hello {
  static func main() async {
    await withTaskGroup(of: Int.self) { group in
      for i in 0..<4 { group.addTask { i * i } }
      var total = 0
      for await v in group { total += v }
      print("sum of squares:", total)
    }
  }
}
```

```
$ bazel build //:hello --platforms=//platforms:wasm32
Target //:hello up-to-date: bazel-bin/hello.wasm      # links against …/_Concurrency.swiftmodule/wasm32-unknown-wasip1-threads.swiftmodule

$ wasmtime run -W threads=y,shared-memory=y -S threads=y bazel-bin/hello.wasm
sum of squares: 14
```

The compile resolves the `wasm32-unknown-wasip1-threads` swiftmodules and the linked module is the threads variant (shared `env::memory`), running under wasmtime's threads proposal.

**Note on the triple fix (`206a15e1`):** before it, the generated toolchain compiled with `--target=wasm32-unknown-wasip1` while the SDK only ships `…-threads` modules, so swiftc failed module resolution (`found: wasm32-unknown-wasip1-threads`). Substituting the actual `{target_triple}` into both the clang `--target` and `CC_TARGET_TRIPLE` make-variable fixes it.

**One gotcha worth flagging for users:** the host `swift_version` must share a swiftmodule format with the SDK — pairing a 6.3 SDK with a 6.3.2 host fails with `module compiled with Swift 6.3 cannot be imported by Swift 6.3.2 compiler`. Matching them (6.3 ↔ 6.3-RELEASE SDK) resolves it. Not a defect in this PR, but easy to trip over.

AttilaTheFun:
Pushed 03a6f451, which replaces the hand-mirrored flag lists and the bundle-layout scan with parsing of the bundle's own metadata (`info.json` → `swift-sdk.json` → `toolset.json`) — the PR description is updated to match. Both bundles ship the same metadata scheme, so the single-threaded path goes through it too; the only behavior delta there is the swift.org toolset's `-static-stdlib` now reaching swiftc (inert for compiles — objects came out bit-identical, and `reactor_test` passes). Re-ran the threads end-to-end after the rework: generated toolchains byte-identical to the previous hand-mirrored flags, and the wasmtime peak-concurrency probe still parallelizes.

The parsing helpers are deliberately SDK-kind-agnostic (and `_relative_metadata_path` is lifted verbatim from #1813), so whichever of #1813 / this lands first, the other can rebase onto shared parsing.

PR #1864, opened 2026-07-22
Remove Xcode rpaths for non macOS binaries

keith:
These are for XCTest + compatibility libraries, but when running in a
simulator these paths never exist so they shouldn't be added.

https://github.com/bazelbuild/rules_swift/pull/1589#issuecomment-5051707210

### DIFF
--- a/swift/toolchains/xcode_swift_toolchain.bzl
+++ b/swift/toolchains/xcode_swift_toolchain.bzl
@@ -119,6 +119,12 @@ _DEVELOPER_DIR_SYMLINKS = [
     "/var/db/xcode_select_link",
 ]
 
+def _developer_dir_symlinks_for_target(target_triple):
+    """Returns the developer directory symlinks that are valid for a target."""
+    if target_triples.unversioned_os(target_triple) == "macos":
+        return _DEVELOPER_DIR_SYMLINKS
+    return []
+
 def _platform_developer_framework_dir(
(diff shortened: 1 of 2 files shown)


qyang-nj:
I was about to put up a fix, but you beat me to it! I tested your changes and confirmed that those paths no longer appear in the rpath of iOS binaries. Thanks for the quick fix!

swift/toolchains/xcode_swift_toolchain.bzl
```diff
@@ -119,6 +119,12 @@ _DEVELOPER_DIR_SYMLINKS = [
     "/var/db/xcode_select_link",
 ]
 
+def _developer_dir_symlinks_for_target(target_triple):
+    """Returns the developer directory symlinks that are valid for a target."""
+    if target_triples.unversioned_os(target_triple) == "macos":
```
adincebic, line 124: At the time when I cherry picked this change from upstream I was not able to verify that those paths should be macOS only.

keith: I don't think they can ever be for rpaths since when running in a simulator / on device this Xcode path will never be valid.

adincebic: Makes sense.

adincebic (approved):

PR #1866, opened 2026-07-23
Fix symbol graph extraction when explicit modules is enabled

amorde:
Fix symbol graph extraction when using explicit modules via `swift.use_c_modules`

When `swift.use_c_modules` is enabled, the `SwiftSymbolGraphExtract` action would send unrecognized arguments resulting in errors:

```
<unknown>:0: error: unknown argument: '-Xfrontend'
<unknown>:0: error: unknown argument: '-disable-building-interface'
<unknown>:0: error: unknown argument: '-Xfrontend'
<unknown>:0: error: unknown argument: '-disable-implicit-swift-modules'
```

Other issues when `swift.use_c_modules` is enabled:
* `swift-symbolgraph-extract` does not apply `-clang-target` to its ClangImporter instance, so precompiled modules built for a shared clang target are rejected as a configuration mismatch (`"was compiled for the target 'arm64-apple-iosXX' but the current translation unit is being compiled for target ..."`).
* `swift-symbolgraph-extract` doesn't support explicit module map JSON files. So when this is used in combination with disabling explicit modules via `swift.use_c_modules`, the extractor would be unable to load modules for system modules (`"cannot load underlying module for 'Darwin'"`)

### DIFF
--- a/swift/toolchains/config/compile_config.bzl
+++ b/swift/toolchains/config/compile_config.bzl
@@ -738,12 +738,15 @@ def compile_action_configs(
         ),
 
         # Configure how implicit modules are handled--either using the module
-        # cache, or disabled completely when using explicit modules.
+        # cache, or disabled completely when using explicit modules. Symbol
+        # graph extraction always uses implicit modules because
+        # `swift-symbolgraph-extract` cannot consume explicit modules (see
+        # below), so it gets a module cache configuration even when
+        # `swift.use_c_modules` is enabled.
         ActionConfigInfo(
             actions = all_compile_action_names() + [
                 SWIFT_ACTION_COMPILE_MODULE_INTERFACE,
                 SWIFT_ACTION_DUMP_AST,
-                SWIFT_ACTION_SYMBOL_GRAPH_EXTRACT,
             ],
             configurators = [_global_module_cache_configurator],
             features = [SWIFT_FEATURE_USE_GLOBAL_MODULE_CACHE],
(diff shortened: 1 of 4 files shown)


luispadron (approved):

keith:
thanks for submitting this! can you test with https://github.com/bazelbuild/rules_swift/pull/1870 instead? I think we can keep it mostly using explicit modules with that change

amorde:
@keith that seems to work with one caveat - `sqlite3` is missing for some targets that transitively depend on it

> <unknown>:0: error: missing required module 'SQLite3'

It seems the cause is this being an implicit dependency and the symbol graph extraction only includes the toolchain's `implicit_deps_providers` rather than collecting them with `get_swift_implicit_deps`. 

this patch appears to fix that issue:

```diff
diff --git swift/internal/symbol_graph_extracting.bzl swift/internal/symbol_graph_extracting.bzl
--- swift/internal/symbol_graph_extracting.bzl
+++ swift/internal/symbol_graph_extracting.bzl
@@ -20,7 +20,7 @@ load(":actions.bzl", "run_toolchain_action")
 load(":compiling.bzl", "transitive_swift_dependency_inputs")
 load(":features.bzl", "gather_toolchains")
 load(":toolchain_utils.bzl", "SWIFT_TOOLCHAIN_TYPE")
-load(":utils.bzl", "merge_compilation_contexts")
+load(":utils.bzl", "get_swift_implicit_deps", "merge_compilation_contexts")
 
 def extract_symbol_graph(
         *,
@@ -81,16 +81,24 @@ def extract_symbol_graph(
         toolchains = toolchains,
     )
 
+    # Use the same implicit dependencies as compilation, which also include
+    # the toolchain's system modules (e.g. the precompiled SDK modules under
+    # `swift.add_default_precompiled_modules`). Extraction cannot read the
+    # explicit Swift module map JSON file, so any system module missing here
+    # cannot be provided to the action at all.
+    implicit_swift_infos, implicit_cc_infos = get_swift_implicit_deps(
+        feature_configuration = feature_configuration,
+        swift_toolchain = toolchains.swift,
+    )
+
     merged_compilation_context = merge_compilation_contexts(
         transitive_compilation_contexts = compilation_contexts + [
             cc_info.compilation_context
-            for cc_info in toolchains.swift.implicit_deps_providers.cc_infos
+            for cc_info in implicit_cc_infos
         ],
     )
     merged_swift_info = SwiftInfo(
-        swift_infos = (
-            swift_infos + toolchains.swift.implicit_deps_providers.swift_infos
-        ),
+        swift_infos = swift_infos + implicit_swift_infos,
     )
```

I'll close this PR though as yours seems like the path forward

keith:
nice, i pushed a fix for that, can you test again with my PR?

amorde:
confirmed the current PR state works in our setup. thanks!

PR #1870, opened 2026-07-28
Fix symbolgraph extraction with explicit modules

keith:
This CLI doesn't accept all arguments, so remove it from some configs,
it also doesn't accept the explicit module json file, so pass all
modules directly

Closes https://github.com/bazelbuild/rules_swift/pull/1866


aaronsky (approved):

luispadron (approved):

amorde:
Confirmed this works in our project where I had encountered the original issue. Thank you!

keith:
thanks!

PR #1878, opened 2026-07-31
Allow discovering host Xcodes with vendored BUILD

keith:
Previously when you vendored an explicit modules BUILD file, this would
disable all scanning locally. This is still the default behavior but now
you can set an env var to also discover local Xcodes. This is useful for
2 cases:

1. If you run remote exec builds from Linux, you have to vendor a BUILD
   file, but you might not strictly enforce Xcode versions for local
   devs
2. You do enforce strict versions for local devs, but also want to test
   betas without having to manually generate these BUILD files before
   testing

### DIFF
--- a/doc/explicit_modules.md
+++ b/doc/explicit_modules.md
@@ -159,6 +159,17 @@ use_repo(system_sdk, "system_sdk")
 This is useful if you trigger remote macOS builds from Linux hosts. It
 is up to you to generate the file with whatever SDKs you need.
 
+If you want to have a vendored BUILD file, but still allow discovering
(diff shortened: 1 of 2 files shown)


keith:
i went back and forth on the api for something in the MODULE.bazel vs env vs both, I'm up for suggestions!

adincebic (approved):

adincebic:
> i went back and forth on the api for something in the MODULE.bazel vs env vs both, I'm up for suggestions!

I would avoid MODULE.bazel. Env seems more flexivle.

aaronsky (approved):

PR #1780, opened 2026-05-28
Implement swift layering check

keith:
Cherry picks:

- 545298b9952141cee46fe3ca5cea883202dfc66b
- 97edd06270589b2d3ba45713bcd02edadf56421f
- b6d5790ac4c0e7cd4001dea80d688f05944f3319
- 6288f6a4a25d712ad184adb5aa64013e211a9c91
- 702902999c75dc38ce7254d5f280d6430744de6e
- 3a9b60a44872603eb4957fffb74b26580433bd15
- 09571d2e99cff8269979ea4867ba605eaee03c9f
- d1736bf4954a354f74f39da2758c69fb1e1dd450
- 55f2e4792b248ff8628b8296d12121c8640ac10d
- fb61b07b0da69b4fd1a4f66ba81466c5caa3baa2

PR #1879, opened 2026-08-03
Gate the direct SwiftInfo pass-throughs on the layering features

mauriciogg:
b054a972 (#1780) re-exports dependency `SwiftInfo`s as *direct* in two places so the Swift layering check can see through targets that do not produce a module map:

1. `swift_clang_module_aspect`'s module-map-less path, as `SwiftInfo(direct_swift_infos = direct_swift_infos + swift_infos)`
2. `swift_library_group`, as `SwiftInfo(direct_swift_infos = get_providers(deps, SwiftInfo), ...)`

This is causing OOM errors in Blaze in our repo ~350K targets since the list is materialized for every single target. This change gates the flattening of the `swift_infos` on the layering check feature (which we have disabled) so we can rebase our fork to the latest version.

### DIFF
--- a/swift/swift_clang_module_aspect.bzl
+++ b/swift/swift_clang_module_aspect.bzl
@@ -22,6 +22,8 @@ load("//swift/internal:compiling.bzl", "compile", "precompile_clang_module")
 load(
     "//swift/internal:feature_names.bzl",
     "SWIFT_FEATURE_ENABLE_EMBEDDED",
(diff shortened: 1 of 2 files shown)


adincebic (approved):
Thanks.

keith:
I think we should come up with a different solution to this. it's good that this fixed the default case, but we would like people to use layering_check, but if it comes with this big downside they probably won't. im sure we can come up with something that better handles memory usage.

PR #1886, opened 2026-08-11
Pin xcode_swift_toolchain to a macOS execution platform

mauriciogg:
`xcode_swift_toolchain` requires Xcode and can only run on macOS, but the rule carries no execution-platform constraint. In builds that register non-mac execution platforms first (e.g. Linux remote executors used to cross-compile for Apple platforms), toolchain resolution hands the rule a Linux execution platform, and `use_cc_toolchain()` then pairs the Xcode Swift toolchain with that platform's C++ toolchain, breaking analysis.

Before the migration to `use_cc_toolchain()`, the constraint was expressed through the rule's default exec group; restore it on the rule itself.

### DIFF
--- a/swift/toolchains/xcode_swift_toolchain.bzl
+++ b/swift/toolchains/xcode_swift_toolchain.bzl
@@ -1165,6 +1165,12 @@ for incremental compilation using a persistent mode.
         },
     ),
     doc = "Represents a Swift compiler toolchain provided by Xcode.",
+    # This toolchain requires Xcode and can only run on macOS. Without an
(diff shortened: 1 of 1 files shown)


keith (commented):
I thought these were enough for this case? https://github.com/bazelbuild/rules_swift/blob/f765fda9083bc88c74c47f48b496d32664445e28/swift/toolchains/BUILD#L89-L105

seems harmless to me tho

keith:
im curious what you come up with on the rules_apple change, since similarly i would expect the toolchain definitions to be enough here

PR #1887, opened 2026-08-11
Support the protobuf worker protocol in order to support RBE

mauriciogg:
Support the protobuf worker protocol in order to support RBE

Our RBE environment only supports protobuf as worker protocol so we need this in orde to enable RBE for our iOS builds. `requires-worker-protocol` is still `json` so this is a noop unless explcitly overriden in command line


keith:
does this mean your remote exec provider doesn't support this even with https://github.com/bazelbuild/bazel/pull/30641 ? (or `--remote_default_exec_properties=persistentWorkerProtocol=json` for testing now)

tools/worker/BUILD
```diff
@@ -155,6 +170,7 @@ cc_library(
     }),
     deps = [
         "@nlohmann_json//:json",
+        "@protobuf//:protobuf",
```
keith, line 173: I always assumed that if we were building this building protoc was kinda fine, how much do we save by doing it this checked in way?

keith (commented):
IIRC moving to json was primarily to remove the protobuf compile, i think we should gate this behind a starlark bool_flag so there's no penalty to users who aren't using this

tools/worker/BUILD
aaronsky: does this still matter now that prebuilt_protoc is the default in bazel 9+ and protobuf 34.0+?

adincebic: Iirc it is opt in even in 9.

PR #1891, opened 2026-08-25
Add opt-in direct compiler plugin loading

### LINKED ISSUE #1884: Compiler plugins are loaded by all transitive dependents, making the plugin executable an input to every downstream Swift compile
## Summary

When a `swift_library` declares `plugins = [...]`, the plugin is passed via `-load-plugin-executable` not only to modules that directly depend on that library, but to every transitive consumer. The plugin executable then becomes an action input to the entire reverse-dependency closure - so any change to a macro's implementation recompiles every module downstream of the macro's interface library, even though none of them re-expand anything.

The code in `swift/internal/compiling.bzl` contradicts its own comment:

```python
    # As of the time of this writing (Xcode 15.0), macros are the only kind of
    # plugins that are available. Since macros do source-level transformations,
    # we only need to load plugins directly used by the module being compiled.
    # Plugins that are only used by transitive dependencies do *not* need to be
    # passed; the compiler does not attempt to load them when deserializing
    # modules.
    used_plugins = list(plugins)
    for module_context in transitive_modules:
        if module_context.swift and module_context.swift.plugins:
            used_plugins.extend(module_context.swift.plugins)
```

The comment says direct-only but the loop collects from `transitive_modules`. This has been the case since the original macro support (#1061).

## Impact

Measured in a large monorepo with a macro interface library consumed by a module at the root of the build graph (~2,200 transitive Swift dependents):

| | Stock rules_swift | Direct-only collection |
|---|---|---|
| Edit macro implementation (reachable code, plugin binary changes) | 2,224 actions / 565 ± 34 s | **7 actions / ~12 s** |

Even though each of the 2,200+ recompiles hits early cutoff, the large number of recompiles itself still has a cost.

## Why direct-only is correct

- Macros are source-level transformations at the expansion site. A module whose source contains macro invocations needs the plugin; a module that imports an already-compiled module consumes the expanded declarations as ordinary serialized decls. The compiler does not load plugins when deserializing dependency modules - as the comment above states.
- The interface *module* itself must remain visible transitively (hiding the module breaks dependents with `missing required module`). Plugin collection is separable from module propagation.
- ~SwiftPM behaves this way too: macro plugins are passed only to targets that declare the macro dependency, transitive consumers never receive `-load-plugin-executable`~.
- I wrongly assumed that SwiftPM behaves this way too. **It does not**, it goes through all transitive dependencies: https://github.com/swiftlang/swift-package-manager/blob/5d0f6086cd9af7d9097e31a785d39a72955dee9f/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift#L241-L247

## Behavior change to note

A module that uses a macro without directly depending on the interface library (relying on the plugin arriving transitively, e.g. via an `@_exported` re-export of the interface module) currently compiles by accident; with direct-only collection it fails with a "macro implementation not found" error.

markvasiv:
### What
- Adds an opt-in `swift.load_plugins_from_direct_dependencies` feature. When enabled, plugin collection in `compile()` walks only the direct modules of the compilation's `swift_infos` and `private_swift_infos`. 
- Transitive plugin loading remains the default behavior.
- Documents the feature on the `plugins` attribute.
- Adds `test/fixtures/macros` and a `macro_plugin` analysis-test suite verifying that:
  - Transitive dependents load the plugin by default.
  - Transitive dependents don't load the plugin if the feature is enabled

### Why
- Compiler plugin executables are currently inputs to every transitive dependent's Swift compilation action. As a result, changing a macro implementation invalidates the entire reverse-dependency closure which expensive for large dependency chains.
- Keeping this opt-in preserves compatibility with code that relies on transitive plugin loading.
- Resolves #1884.


swift/internal/feature_names.bzl
```diff
@@ -407,6 +407,15 @@ SWIFT_FEATURE__SUPPORTS_DEVELOPER_DIR = "swift._supports_developer_dir"
...
 SWIFT_FEATURE_DISABLE_SWIFT_SANDBOX = "swift.disable_swift_sandbox"
 
+# If enabled, compiler plugins are loaded only by the module that declares them
+# and modules that directly depend on it. This improves incremental build
+# performance when plugin implementations change, but it is incompatible with
+# code that relies on transitive plugin loading, such as macro use through a
+# transitive import or a macro-expanded default argument.
+SWIFT_FEATURE_LOAD_PLUGINS_FROM_DIRECT_DEPENDENCIES = (
+    "swift.load_plugins_from_direct_dependencies"
+)
```
keith:
```suggestion
SWIFT_FEATURE_LOAD_PLUGINS_FROM_DIRECT_DEPENDENCIES = "swift.load_plugins_from_direct_dependencies"
```

keith (approved):

PR #1894, opened 2026-09-06
Stop passing `-incremental` when SwiftDeriveFiles is enabled

adincebic:
Empty output file map does not contain the per-source dependency paths required for incremental compilation, which makes Swift driver to emit:

```
Incremental compilation has been disabled: Empty.swift has no swiftDeps file
Incremental compilation has been disabled: malformed dependencies file
```

This of course doesn't show up if the `WMO` is enabled which is true for a lot of consumers.

### DIFF
--- a/tools/worker/BUILD
+++ b/tools/worker/BUILD
@@ -28,6 +28,7 @@ cc_library(
         ":worker_protocol",
         "//tools/common:file_system",
         "//tools/common:temp_file",
+        "@abseil-cpp//absl/strings",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


keith:
should it be the other way instead that we fix the swiftdeps files for these? what does xcode do?

adincebic:
> should it be the other way instead that we fix the swiftdeps files for these? what does xcode do?

Yes but I wasn't sure about the future of incremental compilation in rules_swift. I'll support it then properly.

adincebic:
@keith https://github.com/bazelbuild/rules_swift/pull/1897

PR #1895, opened 2026-09-06
Stop emitting `.swiftsourceinfo` if not explicitly requested

adincebic:
Up until this point the feature `swift.emit_swiftsourceinfo` was only controlling whether Bazel should track the file as its output. This meant that `.swiftsourceinfo` was produced but not tracked by Bazel. Now we make sure that `swiftc` does not produce `.swiftsourceinfo` if not explicitly requested.

I also deleted the code which made sure that `.swiftsourceinfo` was part of the incremental compilation since that is not true today.

The `.swiftsourceinfo` handling was originally added in PR #698 (https://github.com/bazelbuild/rules_swift/pull/698) so Swift could skip the merge-module job. Source compilation was already being skipped, but module merging still had to run because the declared `.swiftsourceinfo` output was missing. At the time source info was always emitted, so keeping it around made sense.

Now that source info is optional, both cases are handled:

* When enabled, `.swiftsourceinfo` is still declared, emitted, and preserved.
* When disabled, we pass `-avoid-emit-module-source-info` and don't expect `.swiftsourceinfo` from the merge-module job, so that job can still be skipped.

This also lines up with what the Swift driver does:

* Disabling source info is marked `doesNotAffectIncrementalBuild` (https://github.com/swiftlang/swift-driver/blob/e1d6a2fefe27dd2d7cfae18154a250c846743deb/Sources/SwiftOptions/Options.swift#L49).
* Incremental state is loaded from the `.priors` dependency graph (https://github.com/swiftlang/swift-driver/blob/e1d6a2fefe27dd2d7cfae18154a250c846743deb/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift#L256-L277), using per-source `.swiftdeps` files (https://github.com/swiftlang/swift-driver/blob/e1d6a2fefe27dd2d7cfae18154a250c846743deb/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/DependencySource.swift#L103-L120).
* `.swiftsourceinfo` is computed separately as an optional auxiliary output (https://github.com/swiftlang/swift-driver/blob/e1d6a2fefe27dd2d7cfae18154a250c846743deb/Sources/SwiftDriver/Driver/Driver.swift#L4088-L4109).

Relevant Slack thread: https://bazelbuild.slack.com/archives/C096TL199U0/p1788956785964199


aaronsky (approved):

adincebic:
Relevant Slack thread: https://bazelbuild.slack.com/archives/C096TL199U0/p1788956785964199

PR #1898, opened 2026-09-12
Support precise debug module tracking

adincebic:
This adds the support for newly introduced [Swift precise debug module tracking](https://www.swift.org/blog/module-tracking-in-debug-info/).
The compiler stores the `.swiftmodule` path in each object file’s debug information, so LLDB knows which module file to load.
This replaces the previous linker-based handling (`-add_ast_path`) if toolchain supports it and Swift explicit modules are enabled.


doc/explicit_modules.md
```diff
@@ -40,6 +40,42 @@ would still be embedded in the swiftmodule files. This means debugging
...
+module features required above is disabled. In `opt` mode, this feature
+does not add debug module paths, linker inputs, or default outputs, even
+if debug info is enabled manually with `-g`.
+
+Expression evaluation still requires the `.swiftmodule` and `.pcm`
+files at their original or remapped locations. `swift_binary` and
+`swift_test` include these files in their default outputs, including
+private and implicit dependencies, so they are downloaded with
+`--remote_download_outputs=toplevel`. If you link through other binary
+rules, your build or debugger integration must make these files
```
aaronsky, line 60: does this mean rules_apple will require changes?

aaronsky (approved):

doc/explicit_modules.md
adincebic: Yes.

swift/toolchains/config/compile_config.bzl
```diff
@@ -1597,6 +1610,11 @@ def _emit_module_path_configurator(prerequisites, args):
     """Adds the `.swiftmodule` output path to the command line."""
     args.add("-emit-module-path", prerequisites.swiftmodule_file)
 
+def _debug_module_path_configurator(prerequisites, args):
+    """Adds the current target's `.swiftmodule` path to debug info."""
+
+    args.add("-debug-module-path", prerequisites.swiftmodule_file)
```
keith: might as well inline this

README.md
```diff
@@ -109,6 +109,9 @@ encountered first on your `PATH`.
 To make cacheable builds work correctly with debugging see
 [this doc](doc/debuggable_remote_swift.md).
 
+For Xcode builds using explicit modules, see
```
keith:
```suggestion
For explicit module builds, see
```

swift/internal/feature_names.bzl
```diff
@@ -224,6 +224,15 @@ SWIFT_FEATURE_ADD_DEFAULT_PRECOMPILED_MODULES = "swift.add_default_precompiled_m
 # using a JSON file instead of `-I` search paths.
 SWIFT_FEATURE_USE_EXPLICIT_SWIFT_MODULE_MAP = "swift.use_explicit_swift_module_map"
 
+# If enabled together with `swift.use_c_modules` and
+# `swift.use_explicit_swift_module_map` in dbg or fastbuild mode, object
```
keith:
```suggestion
# If enabled together with explicit modules and debug info,
```

doc/explicit_modules.md
```diff
@@ -40,6 +40,43 @@ would still be embedded in the swiftmodule files. This means debugging
...
+### Module tracking in debug info
+
+With Xcode 27 and newer, `swift.debug_module_path` is enabled by default.
+It requires Swift 6.4 or newer compiler and debugger support, and takes
+effect in `dbg` and `fastbuild` compilation modes when both
+`swift.use_c_modules` and `swift.use_explicit_swift_module_map` are enabled.
+Enable explicit modules as described in [Usage](#usage), including
+`swift.emit_c_module`.
+
+The compiler writes the `.swiftmodule` path into each object file's
```
keith:
```suggestion
In this mode the compiler writes the `.swiftmodule` path into each object file's
```

doc/explicit_modules.md
```diff
@@ -40,6 +40,43 @@ would still be embedded in the swiftmodule files. This means debugging
 in `lldb` with explicit modules would not work with cached swiftmodules
 and is therefore disabled.
 
+### Module tracking in debug info
+
+With Xcode 27 and newer, `swift.debug_module_path` is enabled by default.
```
keith:
```suggestion
With Swift 6.4 / Xcode 27 and newer, `swift.debug_module_path` is enabled by default.
```

doc/explicit_modules.md
```diff
@@ -40,6 +40,43 @@ would still be embedded in the swiftmodule files. This means debugging
...
+
+```
+build --features=-swift.debug_module_path --host_features=-swift.debug_module_path
+```
+
+The previous behavior is also retained when either of the explicit
+module features required above is disabled. In `opt` mode, this feature
+does not add debug module paths, linker inputs, or default outputs, even
+if debug info is enabled manually with `-g`.
+
```
keith:
```suggestion
```

swift/toolchains/config/compile_config.bzl
```diff
@@ -274,6 +275,18 @@ def compile_action_configs(
...
+        # Record this object's own module path in debugging configurations.
+        ActionConfigInfo(
+            actions = [SWIFT_ACTION_COMPILE],
+            configurators = [_debug_module_path_configurator],
+            features = [[
+                SWIFT_FEATURE_DEBUG_MODULE_PATH,
+                SWIFT_FEATURE_USE_C_MODULES,
+                SWIFT_FEATURE_USE_EXPLICIT_SWIFT_MODULE_MAP,
+                compilation_mode,
+            ] for compilation_mode in [SWIFT_FEATURE_DBG, SWIFT_FEATURE_FASTBUILD]],
```
keith:
```suggestion
            ] for compilation_mode in [SWIFT_FEATURE_DBG, SWIFT_FEATURE_FASTBUILD, SWIFT_FEATURE_FULL_DEBUG_INFO]],
```

swift/internal/linking.bzl
```diff
@@ -165,29 +164,32 @@ def _create_embedded_debugging_linking_context(
     Returns:
         A valid `CcLinkingContext`, or `None` if no linking context was created.
     """
-    if (
-        module_context and
-        module_context.swift and
-        should_embed_swiftmodule_for_debugging(
-            feature_configuration = feature_configuration,
-            module_context = module_context,
+    if not (module_context and module_context.swift and module_context.swift.swiftmodule):
```
keith:
```suggestion
    if not (module_context and module_context.swift):
```

this is a change from before, looks like it's implicit that this was required, but might as well keep the loud failure if it's missing then?

swift/internal/linking.bzl
```diff
@@ -165,29 +164,32 @@ def _create_embedded_debugging_linking_context(
...
+        return None
+
+    if uses_precise_debug_module_tracking(feature_configuration):
+        linker_input = cc_common.create_linker_input(
+            owner = label,
+            additional_inputs = depset([module_context.swift.swiftmodule]),
         )
```
keith: in this mode shouldn't we not need the swiftmodule in the link action instead? even the main module's shouldn't be used?

swift/swift_binary.bzl
```diff
@@ -254,6 +255,11 @@ def _swift_binary_impl(ctx):
             executable = output_file,
             files = depset(
                 [output_file] + additional_debug_outputs,
+                transitive = [debug_module_outputs(
+                    feature_configuration = feature_configuration,
+                    module_contexts = module_contexts,
+                    swift_infos = get_providers(ctx.attr.deps, SwiftInfo),
```
keith, line 261: isn't this missing the implicit deps aka swiftmodule files from the SDK for explicit modules? same with swift_test?

swift/swift_binary.bzl
```diff
@@ -254,6 +255,11 @@ def _swift_binary_impl(ctx):
             executable = output_file,
             files = depset(
                 [output_file] + additional_debug_outputs,
+                transitive = [debug_module_outputs(
```
keith, line 258: I wonder if we just shouldn't do this. today if you want to debug one of these binaries you have to manually specify that you want to download `.o` / `.a` files, this is similar behavior just with a new file extension.

README.md
adincebic: A yeah more technically correct.

swift/swift_binary.bzl
adincebic: Actually yes, everyone who needs it can use remote download regex.

keith: yea im not sure that's good ux or not, but im trying to think about why we would do it this way for this feature but not try to do the same for the .o files

swift/internal/linking.bzl
adincebic: Right I missed that.

swift/swift_binary.bzl
adincebic: Didn't use remote cache when testing so missed it. I can add those too but now I wonder whether to have a separate feature that requests all debug outputs like `swift.full_debug_info_outputs` or something like that? Or Drop it all together and simply expect people to use remote download regex.

swift/swift_binary.bzl
adincebic: Actually those are included through `compilation_context.debug_modules`. During compilation we include the implicit deps too:

```
all_swift_infos = (
     swift_infos_to_propagate + private_swift_infos + implicit_swift_infos
)
```

```
merged_swift_info = SwiftInfo(swift_infos = all_swift_infos)
transitive_modules = merged_swift_info.transitive_modules.to_list()
```

They are then stored in the compilation context:

```
debug_modules = collect_debug_module_files(transitive_modules),
```

And when collecting the debug outputs they get picked up too:

```
if module.compilation_context:
    transitive.append(module.compilation_context.debug_modules)
```

swift/swift_binary.bzl
adincebic: It occured to me that we should also think about output groups, they feel like a perfect fit here. There is already precedent with `dsyms`.

keith: i think if we want to do that we should have a more general "files required for debugging" output group and put other things there too. given that most of the complexity of this change is carrying these files around i think we should drop that and potentially add it later instead

adincebic: Sounds good.

doc/explicit_modules.md
adincebic: TBD.

keith (approved):

PR #1899, opened 2026-09-14
Embed `swiftsourceinfo` and `.swiftdoc` paths in explicit module map

adincebic:
Previously the explicit module map only included `.swiftmodule` paths. SourceKit could resolve imported symbols, but documentation and source locations were missing.

Now we include `docPath` and `sourceInfoPath` when available, and explicitly list them as action inputs. This follows Swift’s explicit module loader (https://github.com/swiftlang/swift/blob/d2a794410e16d9699afd72d24e0a26ed51bcd9c5/lib/Frontend/ModuleInterfaceLoader.cpp#L2417) and module map tests (https://github.com/swiftlang/swift/blob/d2a794410e16d9699afd72d24e0a26ed51bcd9c5/test/ScanDependencies/explicit-module-map.swift#L11).


swift/internal/compiling.bzl
```diff
@@ -155,9 +155,16 @@ def _explicit_swift_module_map_info(
         explicit_swift_module_map_file = explicit_swift_module_map_file,
         module_contexts = module_contexts,
     )
+    inputs = transitive_swift_dependency_inputs(module_contexts)
```
keith: lets add a false by default bool to this function to add these other ones to what is returned from here

keith (approved):

Issue #1126, opened 2023-11-02
Bug: Incremental compilation cache poisoning


brentleyjones:
I _assume_ it's from incremental compilation, because I don't know how a normal compile could do this.

I have a case where _somehow_ (I think a failed or cancelled compile) an old version of a `.swiftmodule` was uploaded to the remote cache. For example, I have an action that sets these env variables:

```
  "environmentVariables": [{
    "name": "APPLE_SDK_PLATFORM",
    "value": "MacOSX"
  }, {
    "name": "APPLE_SDK_VERSION_OVERRIDE",
    "value": "14.2"
  }, {
    "name": "SWIFT_AVOID_WARNING_USING_OLD_DRIVER",
    "value": "1"
  }, {
    "name": "XCODE_VERSION_OVERRIDE",
    "value": "15.1.0.15C5042i"
  }]
```

but I'm getting back a `.swiftmodule` from the remote cache that was compiled against macOS SDK 14.0:

```
/Users/brentley/Developer/rules_xcodeproj/tools/generators/lib/PBXProj/src/Dictionary+Extensions.swift:1:8: Cannot load module 'OrderedCollections' built with SDK 'macosx14.0' when using SDK 'macosx14.2': /Users/brentley/Developer/rules_xcodeproj/bazel-output-base/rules_xcodeproj.noindex/build_output_base/execroot/_main/bazel-out/macos-arm64-min13.0-applebin_macos-darwin_arm64-dbg-ST-95054d4cebee/bin/external/_main~non_module_deps~com_github_apple_swift_collections/OrderedCollections.swiftmodule
```

I feel there is an edge case in the incremental compilation support that could allow something like this to happen (because it has old versions laying around).

PR #1905, opened 2026-09-18
Fix cache poisoning with Swift incremental compilation

adincebic:
Previously we were copying module files  back and forth between incremental storage area and Bazel output location which led to cache poisoning in some instances.
Now we just copy from incremental storage area to Bazel output location after successful build which should prevent issues described in https://github.com/bazelbuild/rules_swift/issues/1126.


adincebic:
CI failing because of `xcode_config` missing from Bazel HEAD. https://github.com/bazelbuild/rules_swift/pull/1906 fixes it.

adincebic:
@keith thanks for rebasing.

PR #1903, opened 2026-09-16
Bump buildifier to 10.0.1

adincebic:

PR #1907, opened 2026-09-21
Restore feature_configuration argument to derive_swift_module_name

brentleyjones:
Restore the optional `feature_configuration` keyword argument accepted by `derive_swift_module_name` in 4.0.1 and 4.1.0. Its removal in #1903 breaks existing callers even though the argument is currently unused.

Keep the argument as a no-op with an explicit `unused-variable` suppression, and regenerate the API documentation. This preserves compatibility for the next patch release.


swift/module_name.bzl
```diff
@@ -18,7 +18,10 @@ load("@bazel_skylib//lib:types.bzl", "types")
 
 visibility("public")
 
-def derive_swift_module_name(*args):
+def derive_swift_module_name(
+        *args,
+        # buildifier: disable=unused-variable
+        feature_configuration = None):
```
adincebic, line 24: Is it possible to adjust the caller?

